PR #3, opened 2025-04-10
Add initial operation to json schema

Jephuff:
I'm not entirely sure how we get this into the toolbox yet, but this is the starting point of converting a basic operation to a json schema of it's variables.

### DIFF
--- a/crates/mcp-apollo-server/src/lib.rs
+++ b/crates/mcp-apollo-server/src/lib.rs
@@ -1,1 +1,1 @@
-pub mod operation_to_json_schema;
+pub mod operations;
--- a/crates/mcp-apollo-server/src/operations.rs
+++ b/crates/mcp-apollo-server/src/operations.rs
@@ -162,7 +162,7 @@ fn type_to_schema(
 mod tests {
     use std::collections::HashMap;
 
-    use crate::operation_to_json_schema::{ToolDefinition, operation_to_json_schema};
+    use crate::operations::{ToolDefinition, operation_to_json_schema};
     use apollo_compiler::parser::Parser;
     use rmcp::{
         schemars::schema::{InstanceType, SchemaObject, SingleOrVec},


crates/mcp-apollo-server/src/operation_to_json_schema.rs
```diff
@@ -0,0 +1,230 @@
...
+        schema: SchemaObject {
+            instance_type: Some(SingleOrVec::Single(Box::new(InstanceType::Object))),
+            object: Some(Box::new(obj)),
+            ..Default::default()
+        },
+        ..Default::default()
+    }
+}
+
+fn type_to_schema(variable_type: &Type) -> Schema {
```
pubmodmatt: Why not impl `From` or `TryFrom`?

crates/mcp-apollo-server/src/operation_to_json_schema.rs
```diff
@@ -0,0 +1,230 @@
+use apollo_compiler::{ast::{Definition, Type}, parser::Parser};
+use rmcp::schemars::schema::{ArrayValidation, InstanceType, ObjectValidation, RootSchema, Schema, SchemaObject, SingleOrVec, SubschemaValidation};
+
+/**
+ * TODO
+ * - support for input objects
+ * - support for custom scalars
+ * - error handling
+ */
+pub fn operation_to_json_schema(uri: &str, source_text: &str) -> RootSchema {
```
pubmodmatt: Limiting the variables to Int/Float/Boolean/String/ID means we only need the GraphQL document here, but we're eventually going to need the SDL as well so we can support input types. It maybe worth adding that on the signature now so anything building with this will be correct.

Jephuff: I'm going to work on porting the input type stuff now so I'll add it 👍

crates/mcp-apollo-server/src/operation_to_json_schema.rs
Jephuff: Never done it! I'll give it a go :D

Jephuff: This sounds like it's telling me I can't? 
```
only traits defined in the current crate can be implemented for types defined outside of the crate
impl doesn't have any local type before any uncovered type parameters
for more information see https://doc.rust-lang.org/reference/items/implementations.html#orphan-rules
define and implement a trait or new type insteadrustc[Click for full compiler diagnostic](rust-analyzer-diagnostics-view:/diagnostic%20message%20[1]?1#file:///Users/jeffrey/projects/mcp-apollo/crates/mcp-apollo-server/src/operation_to_json_schema.rs)
```

pubmodmatt: Welcome to Rust! You need to use a newtype in cases like this. Basically, that means you wrap the type in your own type, then you can impl whatever you want on it. But in this case, it's probably not what we want anyway since in your latest version, it now requires an additional parameter for the GraphQL schema. So disregard this.

crates/mcp-apollo-server/src/lib.rs
```diff
@@ -0,0 +1 @@
+pub mod operation_to_json_schema;
```
pubmodmatt: Just a nit here that this should be a simpler module name like `operations`

Jephuff: makes sense! done 👍

PR #7, opened 2025-04-15
refactor with fetch

Jephuff:
Adds the execute_operation and refactors to be a struct with `Tool` and `execute_operation` 
Does this refactor make sense? thoughts?

### DIFF
--- a/crates/mcp-apollo-server/src/operations.rs
+++ b/crates/mcp-apollo-server/src/operations.rs
@@ -243,9 +243,9 @@ mod tests {
                 "operation.graphql",
             )
             .expect("failed to parse operation");
-        let grpahql_schema = document.to_schema().unwrap();
+        let graphql_schema = document.to_schema().unwrap();
 
-        let operation = Operation::new(source_text, &grpahql_schema, custom_scalar_map);
+        let operation = Operation::new(source_text, &graphql_schema, custom_scalar_map);
         let tool = operation.as_tool();
         assert_eq!(json!(tool.input_schema), expected_json);
     }


crates/mcp-apollo-server/src/operations.rs
```diff
@@ -1,55 +1,100 @@
...
+            }
+            _ => panic!("no operations in document"),
+        }
+    }
 
-    let operation_name = operation
-        .name
-        .clone()
-        .expect("Operations require names")
+    pub async fn execute_operation(
```
pubmodmatt: Change the name to just `execute`, it's already an operation

crates/mcp-apollo-server/src/operations.rs
```diff
@@ -1,55 +1,100 @@
...
+        variables: serde_json::Value,
+    ) -> Result<String, reqwest::Error> {
+        let client = reqwest::Client::new();
+        let body = serde_json::json!({
+            "query": self.source_text,
+            "variables": variables,
+        })
         .to_string();
 
+        match client.post(endpoint).body(body).send().await {
```
pubmodmatt: You'll need to set `.header("Content-Type", "application/json")` here or the router will reject it

crates/mcp-apollo-server/src/operations.rs
```diff
@@ -1,55 +1,100 @@
...
-        Definition::FragmentDefinition(_) => None,
-        _ => {
-            eprintln!(
-                "Schema definitions were passed in, only operations and fragments are allowed"
-            );
-            None
-        }
-    });
+pub struct Operation {
+    pub tool: Tool,
```
pubmodmatt: Rather than make this pub, I'd add:

```rust
    pub fn as_tool(&self) -> &Tool {
        &self.tool
    }
```

PR #9, opened 2025-04-15
Add descriptions to input schema

Jephuff:
This adds the description from the schema into the input json schema
- if a custom scalar schema thats passed in doesn't have a description it will add the one from the schema
- enum value descriptions currently get lost, we could either replace json schemas enum with a `oneOf` with single enums, or maybe just combine their descriptions in the fields description

### DIFF
--- a/crates/mcp-apollo-server/src/operations.rs
+++ b/crates/mcp-apollo-server/src/operations.rs
@@ -121,7 +121,7 @@ fn get_json_schema(
         let type_name = variable.ty.inner_named_type();
         let schema = type_to_schema(
             // For the root description, for now we can use the type description.
-            get_description(type_name, graphql_schema),
+            description(type_name, graphql_schema),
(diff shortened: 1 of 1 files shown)


crates/mcp-apollo-server/src/operations.rs
```diff
@@ -129,51 +143,93 @@ fn get_json_schema(
...
         subschemas: subschema_validation.map(|validation| Box::new(validation)),
+        enum_values: enum_values,
+        metadata: Some(Box::new(Metadata {
+            description: description,
+            ..Default::default()
+        })),
         ..Default::default()
     })
 }
+fn get_description(name: &Name, graphql_schema: &GraphqlSchema) -> Option<String> {
```
pubmodmatt: Getters in Rust don't start with `get_`, just change to `description`

PR #10, opened 2025-04-16
Parse the comments before the operation as it’s description

Jephuff:
This adds a secondary description option. If the generated descriptions aren't working for the models, or they are too big users can add a comment before the operation that will be used for the description instead. If no comment exists, it will still do the generated description.

In this example the description would be "this tool will just return a hat for you."
```graphql
# this tool will just return a hat for you.
query AToolToGetHats {
  hat
}
```

### DIFF
--- a/crates/mcp-apollo-server/src/operations.rs
+++ b/crates/mcp-apollo-server/src/operations.rs
@@ -1,11 +1,10 @@
-use std::collections::HashMap;
-
 use apollo_compiler::ast::{FragmentDefinition, Selection};
 use apollo_compiler::{
     Name, Node, Schema as GraphqlSchema,
     ast::{Definition, OperationDefinition, Type},
     parser::Parser,
 };
+use regex::Regex;
 use rmcp::{
     model::Tool,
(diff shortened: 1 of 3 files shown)


crates/mcp-apollo-server/src/operations.rs
```diff
@@ -124,85 +146,104 @@ impl Operation {
...
+                None
+            } else {
+                Some(trimmed.to_string())
+            }
+        });
+
+        match comment_description {
+            Some(description) => description,
+            None => {
+                let mut tree_shaker = TreeShaker::new(graphql_schema, fragment_defs);
```
pubmodmatt, line 168: So if the description is overridden, we're also not providing any type information?

Jephuff: Right now yes, the comment is treated as a full override.

pubmodmatt: Ok for now. We may need to consider how to make this available to the agent. This may be where having the ability to introspect tools comes in.

PR #15, opened 2025-04-16
remove panics

nicholascioli:
This commit removes all uses of `expect`s and `unwrap`s in favor of `Result`s and proper error handling.

It also enables clippy's `#![deny(clippy::expect_used)]` and `#![deny(clippy::unwrap_used)]` so that future code errors on any uses of it.

### DIFF
--- a/crates/mcp-apollo-server/src/operations.rs
+++ b/crates/mcp-apollo-server/src/operations.rs
@@ -1,7 +1,6 @@
 use std::collections::HashMap;
 
-use apollo_compiler::ast::{Document, OperationType, Selection};
-use apollo_compiler::validation::WithErrors;
+use apollo_compiler::ast::{OperationType, Selection};
 use apollo_compiler::{
     Name, Node, Schema as GraphqlSchema,
     ast::{Definition, OperationDefinition, Type},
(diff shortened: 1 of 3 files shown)


crates/mcp-apollo-server/src/operations.rs
```diff
@@ -34,60 +59,62 @@ impl From<Operation> for Tool {
 }
 
 impl Operation {
-    pub fn new(
+    pub fn from_spec(
```
Jephuff: maybe from_document?
```suggestion
    pub fn from_document(
```

crates/mcp-apollo-server/src/operations.rs
```diff
@@ -34,60 +59,62 @@ impl From<Operation> for Tool {
...
+            .map_err(OperationError::GraphQLDocument)?;
 
-        let mut operation_defs = document.definitions.iter().filter_map(|def| match def {
+        let mut operation_defs = document.definitions.into_iter().filter_map(|def| match def {
             Definition::OperationDefinition(operation_def) => Some(operation_def),
             Definition::FragmentDefinition(_) => None,
             _ => {
-                eprintln!(
+                tracing::error!(
+                    spec=?def,
```
Jephuff, line 53: what is this 👀

crates/mcp-apollo-server/src/operations.rs
```diff
@@ -16,6 +17,30 @@ use rmcp::{
 };
 use serde_derive::Serialize;
 
+#[derive(Debug, thiserror::Error)]
```
pubmodmatt: It's common to have errors in a separate `errors` module - thoughts on that?

crates/mcp-apollo-server/src/operations.rs
pubmodmatt: or just `parse`?

crates/mcp-apollo-server/src/operations.rs
nicholascioli: It just prints it out with the key `spec` using the debug implementation for printing!

crates/mcp-apollo-server/src/operations.rs
nicholascioli: I feel like parse implies that it implements [`FromStr`](https://doc.rust-lang.org/std/str/trait.FromStr.html), so I'll go with `from_document`!

crates/mcp-apollo-server/src/operations.rs
nicholascioli: Moved!

crates/mcp-apollo-server/src/operations.rs
Jephuff: ooh neat!

PR #17, opened 2025-04-17
setup cli headers

Jephuff:
Allows headers to be passed in to the graphql fetch so users can control any that they need.


crates/mcp-apollo-server/src/main.rs
```diff
@@ -36,6 +36,10 @@ struct Args {
     /// The GraphQL endpoint the server will invoke
     #[clap(long, short = 'e', default_value = "http://127.0.0.1:4000")]
     endpoint: String,
+
+    /// The GraphQL endpoint the server will invoke
```
pubmodmatt:
```suggestion
    /// Headers to send to endpoint
```

PR #21, opened 2025-04-18
add persisted queries manifest support

nicholascioli:
This commit adds support for the persisted query manifest format when loading operations.


nicholascioli:
Closing in favor of #24

PR #27, opened 2025-04-25
add support for passing in a custom_scalar_map file

Jephuff:
This PR adds support for passing in a path to a JSON file that contains custom scalar schemas
```sh
mcp-apollo-server -c custom_scalar_map.json
```
The format of the file is keys are the Scalar names, and values are JsonSchemas to use for them
```
{
  "MyScalar": { "type": "string" }
}
```

### DIFF
--- a/crates/mcp-apollo-server/src/custom_scalar_map.rs
+++ b/crates/mcp-apollo-server/src/custom_scalar_map.rs
@@ -3,12 +3,12 @@ use rmcp::{
     schemars::schema::{Schema, SchemaObject, SingleOrVec},
     serde_json,
 };
-use std::{collections::HashMap, path::PathBuf};
+use std::{collections::HashMap, path::PathBuf, str::FromStr};
 
-impl TryFrom<&str> for CustomScalarMap {
-    type Error = ServerError;
+impl FromStr for CustomScalarMap {
+    type Err = ServerError;
 
-    fn try_from(string_custom_scalar_file: &str) -> Result<Self, Self::Error> {
+    fn from_str(string_custom_scalar_file: &str) -> Result<Self, Self::Err> {
         // Parse the string into an initial map of serde_json::Values
         let parsed_custom_scalar_file: serde_json::Map<String, serde_json::Value> =
             serde_json::from_str(string_custom_scalar_file)
@@ -44,7 +44,7 @@ impl TryFrom<&PathBuf> for CustomScalarMap {
         let custom_scalars_config_path = file_path_buf.as_path();
         tracing::info!(custom_scalars_config=?custom_scalars_config_path, "Loading custom_scalars_config");
         let string_custom_scalar_file = std::fs::read_to_string(custom_scalars_config_path)?;
-        CustomScalarMap::try_from(string_custom_scalar_file.as_str())
+        CustomScalarMap::from_str(string_custom_scalar_file.as_str())
     }
 }
 
@@ -79,7 +79,10 @@ fn has_invalid_schema(schema: &Schema) -> bool {
 
 #[cfg(test)]
 mod tests {
-    use std::collections::{BTreeMap, HashMap};
+    use std::{
+        collections::{BTreeMap, HashMap},
+        str::FromStr,
+    };
 
     use rmcp::schemars::schema::{
         InstanceType, ObjectValidation, Schema, SchemaObject, SingleOrVec,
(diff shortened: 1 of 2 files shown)


crates/mcp-apollo-server/src/json_schema_helpers.rs
```diff
@@ -0,0 +1,196 @@
+use crate::errors::ServerError;
+use rmcp::{
+    schemars::schema::{Schema, SchemaObject, SingleOrVec},
+    serde_json,
+};
+use std::collections::HashMap;
+
+pub fn str_to_custom_scalar_map(
```
pubmodmatt: Nit - rather than just making these "helper" functions, I'd consider making this a `CustomScalarMap` type that wraps the `HashMap` and implements `FromStr` to parse a stirng.

crates/mcp-apollo-server/src/main.rs
```diff
@@ -102,6 +107,14 @@ async fn main() -> anyhow::Result<()> {
...
         .headers(args.headers)
         .introspection(args.introspection)
+        .and_custom_scalar_map(
+            args.custom_scalars_config.map(|custom_scalars_config| {
+                let custom_scalars_config_path = custom_scalars_config.as_path();
+                info!(custom_scalars_config=?custom_scalars_config_path, "Loading custom_scalars_config");
+                let string_custom_scalar_file = std::fs::read_to_string(custom_scalars_config.as_path())?;
+                str_to_custom_scalar_map(&string_custom_scalar_file)
+            }).transpose()?
+        )
```
pubmodmatt: I'd also put the file loading code into the `CustomScalarMap` type impl, as a `from_file` method for example, just so all the logic to read custom scalars is in that one place.

Jephuff:
@pubmodmatt maybe take one more pass of this in the morning, I move it to a struct implementing a few try froms 👍

crates/mcp-apollo-server/src/custom_scalar_map.rs
```diff
@@ -0,0 +1,225 @@
...
+        tracing::info!(custom_scalars_config=?custom_scalars_config_path, "Loading custom_scalars_config");
+        let string_custom_scalar_file = std::fs::read_to_string(custom_scalars_config_path)?;
+        CustomScalarMap::try_from(string_custom_scalar_file.as_str())
+    }
+}
+
+#[derive(Debug)]
+pub struct CustomScalarMap {
+    custom_scalar_map: HashMap<String, SchemaObject>,
+}
```
pubmodmatt: Total nit here, but usually it's just a wrapper you'd do like:

```rust
pub struct CustomScalarMap(HashMap<String, SchemaObject>);
```
then reference the inner map with `.0`.

Alternatively, if you want to use a named field, in cases like this the convention is to just name it `inner` rather than repeat the name.

crates/mcp-apollo-server/src/custom_scalar_map.rs
```diff
@@ -0,0 +1,225 @@
...
+
+impl CustomScalarMap {
+    pub fn get(&self, key: &str) -> Option<&SchemaObject> {
+        self.custom_scalar_map.get(key)
+    }
+}
+
+// Unknown keys will be put into "extensions" in the schema object, check for those and consider those invalid
```
pubmodmatt, line 60: Another nit - if this message is for callers, use triple slashes for a method comment ///

crates/mcp-apollo-server/src/custom_scalar_map.rs
```diff
@@ -0,0 +1,225 @@
+use crate::errors::ServerError;
+use rmcp::{
+    schemars::schema::{Schema, SchemaObject, SingleOrVec},
+    serde_json,
+};
+use std::{collections::HashMap, path::PathBuf};
+
+impl TryFrom<&str> for CustomScalarMap {
```
pubmodmatt: Why not `FromStr`?

crates/mcp-apollo-server/src/custom_scalar_map.rs
Jephuff: This wasn't really intended for consumers, more for future devs editing the method to know what it considers invalid.

crates/mcp-apollo-server/src/custom_scalar_map.rs
Jephuff: Didn't know that existed :D 
A little confused why it exists and is slightly different from TryFrom<&str>, but I can switch to it!

pubmodmatt: It's used for the [`parse` method](https://doc.rust-lang.org/std/primitive.str.html#method.parse):

```rust
let csm: CustomScalarMap = "foo".parse()?;
```

Jephuff: ooohh neat!

PR #28, opened 2025-04-28
add license

Jephuff:
Router also had a [license.html](https://github.com/apollographql/router/blob/dev/licenses.html), [about.toml](https://github.com/apollographql/router/blob/dev/about.toml), and [about.hbs](https://github.com/apollographql/router/blob/dev/about.hbs) in the root. Do we need those as well?

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright 2021 Apollo Graph, Inc.
+Copyright 2025 Apollo Graph, Inc.
 
 Source code in this repository is covered by the Elastic License 2.0. The
 default throughout the repository is a license under the Elastic License 2.0,
--- a/crates/mcp-apollo-server/LICENSE
+++ b/crates/mcp-apollo-server/LICENSE
@@ -1,4 +1,4 @@
-Copyright 2021 Apollo Graph, Inc.
+Copyright 2025 Apollo Graph, Inc.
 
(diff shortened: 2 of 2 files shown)


LICENSE
```diff
@@ -1,21 +1,102 @@
...
-copies or substantial portions of the Software.
-
-THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
-IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
-FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
-AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
-LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
-OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
-SOFTWARE.
+Copyright 2021 Apollo Graph, Inc.
```
pubmodmatt: Should this date be updated?

Jephuff: Sure!

PR #29, opened 2025-04-29
create container

nicholascioli:
This PR sets up the infra needed to build containers using Nix and GitHub actions. The container sets the current working directory to `/data` and thus all configuration should be mounted there within the container.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -80,7 +80,7 @@
         inherit pkgs inputs;
         derivations = [cargoArtifacts toolchain] ++ mcp-server-tools;
       };
-    in {
+    in rec {
       devShells.default = pkgs.mkShell {
         nativeBuildInputs = with pkgs; [pkg-config];
         buildInputs =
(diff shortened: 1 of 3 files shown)


nicholascioli:
The digest calculation is definitely very brittle, but it should work for now until we want multi-arch images.

nicholascioli:
The arm64 build step for the release won't work until the repository goes public. Let me know if you'd like it removed until then so that it doesn't "fail".

flake.nix
```diff
@@ -132,5 +132,41 @@
...
+          created = commitDate;
+          mtime = commitDate;
+
+          contents = [
+            packages.mcp-apollo-server
+          ];
+
+          config = {
+            # Make the entrypoint the server
+            Entrypoint = ["mcp-apollo-server" "-d" "/data"];
```
pubmodmatt, line 166: How would a customer configure other options? If they're using the local file options, they'll also need a way to include those in the image. If they're using `--uplink`, they'd also need a way to set the `APOLLO_KEY` and `APOLLO_GRAPH_REF` environment variables.

Adding a README entry like [Gary has here](https://github.com/apollographql/router-all-in-one?tab=readme-ov-file#running-your-all-in-one-container) would help understand this (presumably environment variables would be set like he has shown there, for example).

nicholascioli: Added a section to the README

PR #30, opened 2025-04-29
log character counts and estimated tokens

Jephuff:
Logs the character size of each tool as it loads, logs the total character length of all tools and an estimate of the tokens (seems like ~4 characters is roughly a token, but will obviously vary)

Formatting I have right now, definitely open to suggestions to make it better
```
2025-04-29T19:33:04.100713Z  INFO mcp_apollo_server: Loading schema schema_path="supergraph.graphql"
2025-04-29T19:33:04.107599Z  INFO mcp_apollo_server::server: Loading operation operation_path="graphql/weather/operations/alerts.graphql"
2025-04-29T19:33:04.108761Z  INFO mcp_apollo_server::operations: Tool GetAlerts loaded with a character count of 754
2025-04-29T19:33:04.108793Z  INFO mcp_apollo_server::server: Loading operation operation_path="graphql/weather/operations/all.graphql"
2025-04-29T19:33:04.109415Z  INFO mcp_apollo_server::operations: Tool GetAllWeatherData loaded with a character count of 1489
2025-04-29T19:33:04.109429Z  INFO mcp_apollo_server::server: Loading operation operation_path="graphql/weather/operations/forecast.graphql"
2025-04-29T19:33:04.109708Z  INFO mcp_apollo_server::operations: Tool GetForecast loaded with a character count of 801
2025-04-29T19:33:04.109723Z  INFO mcp_apollo_server::server: Size of tools:
2025-04-29T19:33:04.109727Z  INFO mcp_apollo_server::server:     characters: 3044
2025-04-29T19:33:04.109730Z  INFO mcp_apollo_server::server:     estimated tokens: 761
2025-04-29T19:33:04.109892Z  INFO mcp_apollo_server: Starting MCP server in stdio mode
```

### DIFF
--- a/crates/mcp-apollo-server/src/operations.rs
+++ b/crates/mcp-apollo-server/src/operations.rs
@@ -194,7 +194,7 @@ impl Operation {
                 "Tool {} loaded with a character count of {}. Estimated tokens: {}",
                 operation_name,
                 length,
-                length / 4
+                length / 4 // We don't know the tokenization algorithm, so we just use 4 characters per token as a rough estimate. https://docs.anthropic.com/en/docs/resources/glossary#tokens
             ),
             Err(_) => tracing::info!(
                 "Tool {} loaded with an unknown character count",
--- a/crates/mcp-apollo-server/src/operations.rs
+++ b/crates/mcp-apollo-server/src/operations.rs
@@ -25,7 +25,6 @@ pub struct Operation {
     tool: Tool,
     source_text: String,
     persisted_query_id: Option<String>,
-    character_count: usize,
 }
 
 impl AsRef<Tool> for Operation {
@@ -119,13 +118,14 @@ impl Operation {
             ));
         };
 
-        let tool = Tool::new(operation_name.clone(), description, schema);
+        let tool: Tool = Tool::new(operation_name.clone(), description, schema);
         let character_count = tool_character_length(&tool);
         match character_count {
             Ok(length) => tracing::info!(
-                "Tool {} loaded with a character count of {}",
+                "Tool {} loaded with a character count of {}. Estimated tokens: {}",
                 operation_name,
-                length
+                length,
+                length / 4
             ),
             Err(_) => tracing::info!(
                 "Tool {} loaded with an unknown character count",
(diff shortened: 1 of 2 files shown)


crates/mcp-apollo-server/src/server.rs
```diff
@@ -66,11 +66,19 @@ impl Server {
             )?);
         }
 
-        info!(
+        tracing::debug!(
             "Loaded operations:\n{}",
             serde_json::to_string_pretty(&operations)?
         );
 
+        tracing::info!("Size of tools:");
```
pubmodmatt: You'll have a conflict with the latest code here. Only operations from static operations files are known here. We load PQs async, and they can be updated at any time if there are changes. I'm not sure if we need the total, or if just having the info per tool is enough. If we want the total, you'll have to re-log it as operation changes become known.

Jephuff: should we just skip the total for now and move the estimated tokens to per operation?

pubmodmatt: I think per operation, but don't feel strongly. Unfortunately, I'm currently rewriting a lot of this *again* to support schema changes from uplink (a schema change effective re-creates all the operations since they have schema-based descriptions).

Jephuff: Just updated removing the total, we can always add it in the future :) 

ooh yep, and the input schema could change

crates/mcp-apollo-server/src/operations.rs
```diff
@@ -118,8 +118,22 @@ impl Operation {
...
         };
 
+        let tool: Tool = Tool::new(operation_name.clone(), description, schema);
+        let character_count = tool_character_length(&tool);
+        match character_count {
+            Ok(length) => tracing::info!(
+                "Tool {} loaded with a character count of {}. Estimated tokens: {}",
+                operation_name,
+                length,
+                length / 4
```
pubmodmatt: Why is it dividing by 4 here?

pubmodmatt: I think I found it [here](https://docs.anthropic.com/en/docs/resources/glossary#tokens) - "For Claude, a token approximately represents 3.5 English characters". If that's the case, let's get a comment in here explaining the computation.

Jephuff: Oh yes, 4 was the magical number that everyone seems to say is a reasonable estimate across models. I'll add a comment 👍

PR #31, opened 2025-04-30
Initial uplink implementation

pubmodmatt:
Add the ability to load Persisted Queries from GraphOS through Uplink.

Changes:

* Replace `rover-copy` (based on Rover code) with `mcp-apollo-registry` (based on Apollo Router code). This doesn't support the Relay format, so only the Apollo PQ manifest format will be supported going forward. This aligns the MCP Server with Apollo Router's PQ support, and adds the ability to use Uplink and hot reloading (both for changes from Uplink or changes to a local manifest file).
  * The router code is currently duplicated in `mcp-apollo-registry` to move quickly, but in the future will need to be refactored into a common crate shared with router. The code is mostly the same, but has some minor improvements, such as the use of `secrecy` for API keys, and support for external change notifications.
* Adds [MCP tool change notifications](https://modelcontextprotocol.io/specification/2025-03-26/server/tools#list-changed-notification) when PQ changes are detected. Support for this varies widely depending on the MCP client, and the Rust SDK support is not yet perfect, but these events work with MCP Inspector.
* Persisted Queries are now actually executed as Persisted Queries, using the sha256Hash. Ensure that you have persisted queries enabled in the router behind the endpoint. If you use client names, ensure that you configure the MCP server to send that client name in the appropriate header using the `--header` option.
* Adds a log header for the MCP Server at the INFO level, which includes the version and license.
* Adds ANSI color to log output.

Example of PQ changes being picked up by the MCP Server, and sending tool change notifications in response:

```console
2025-04-30T18:18:04.995069Z  INFO mcp_apollo_registry::uplink::persisted_queries::manifest_poller: Loaded 3 persisted queries.
2025-04-30T18:18:04.995359Z  INFO mcp_apollo_registry::uplink::persisted_queries::manifest_poller: persisted query manifest successfully updated (3 operations total)
2025-04-30T18:18:04.995502Z  INFO mcp_apollo_server::server: Persisted query manifest changed, notifying 1 peers of tool change
```

And in MCP Inspector:

```json
{
  "method": "notifications/tools/list_changed"
}
```

### DIFF
--- a/crates/mcp-apollo-server/src/server.rs
+++ b/crates/mcp-apollo-server/src/server.rs
@@ -124,7 +124,7 @@ impl Server {
             }
         }
 
-        let execute_tool = introspection.then(|| Execute::new());
+        let execute_tool = introspection.then(Execute::new);
         let get_schema_tool = introspection.then(|| GetSchema::new(schema.clone()));
 
         let peers = Arc::new(RwLock::new(vec![]));
@@ -162,32 +162,20 @@ impl Server {
                 for peer in peers.iter() {
                     match peer.notify_tool_list_changed().await {
                         Ok(_) => retained_peers.push(peer.clone()),
-                        Err(e) => {
-                            match e {
-                                ServiceError::Transport(e) => {
-                                    if let Some(error_data) = e.get_ref() {
-                                        if error_data.to_string() == *"disconnected" {
-                                            // TODO: this always gets a "disconnected" error due to a bug in the SDK, but it actually works
-                                            retained_peers.push(peer.clone());
-                                        } else {
-                                            error!(
-                                                "Failed to notify peer of tool list change: {:?} - dropping peer",
-                                                error_data
-                                            );
-                                        }
-                                    } else {
-                                        error!(
-                                            "Failed to notify peer of tool list change: {:?}",
-                                            e
-                                        );
-                                        retained_peers.push(peer.clone());
-                                    }
-                                }
-                                _ => {
-                                    error!("Failed to notify peer of tool list change {:?}", e);
-                                    retained_peers.push(peer.clone());
-                                }
+                        Err(ServiceError::Transport(e)) if e.get_ref().is_some() => {
+                            if e.to_string() == *"disconnected" {
+                                // TODO: this always gets a "disconnected" error due to a bug in the SDK, but it actually works
+                                retained_peers.push(peer.clone());
+                            } else {
+                                error!(
+                                    "Failed to notify peer of tool list change: {:?} - dropping peer",
+                                    e
+                                );
                             }
+                        }
+                        Err(e) => {
+                            error!("Failed to notify peer of tool list change {:?}", e);
+                            retained_peers.push(peer.clone());
                         }
                     }
                 }


crates/mcp-apollo-registry/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+reqwest = { version = "0.12.15", features = ["json", "gzip"] }
+secrecy = "0.10.3"
+serde = { version = "1.0", features = ["derive"] }
+serde_json = "1.0"
+thiserror = "2.0.12"
+tokio = { version = "1.0", features = ["fs", "sync", "time", "rt-multi-thread", "macros"] }
+tokio-stream = "0.1"
+tower = "0.5.2"
+tracing = "0.1"
+url = "2.4"
```
nicholascioli: Some of these are in the root workspace.

crates/mcp-apollo-server/src/main.rs
```diff
@@ -65,33 +67,27 @@ struct Args {
...
 struct ManifestArgs {
-    /// The path to the manifest containing operations to publish.
+    /// The path to the persisted query manifest containing operations.
     #[arg(long, required = false)]
     manifest: PathBuf,
-
-    /// The format of the manifest file.
-    #[arg(long, value_enum, default_value_t = PersistedQueriesManifestFormat::Apollo)]
-    manifest_format: PersistedQueriesManifestFormat,
 }
```
nicholascioli: You could probably get rid of this and just have the top level args struct have this field now that we don't need the format to be paired with the manifest.

crates/mcp-apollo-server/src/server.rs
```diff
@@ -85,22 +124,101 @@ impl Server {
             }
         }
 
+        let execute_tool = if introspection {
+            Some(Execute::new())
+        } else {
+            None
+        };
```
nicholascioli:
```suggestion
        let execute_tool = introspection.then(|| Execute::new());
```

crates/mcp-apollo-server/src/server.rs
```diff
@@ -85,22 +124,101 @@ impl Server {
...
+        let get_schema_tool = if introspection {
+            Some(GetSchema::new(schema.clone()))
+        } else {
+            None
+        };
```
nicholascioli:
```suggestion
        let get_schema_tool = introspection.then(|| GetSchema::new(schema.clone()));
```

crates/mcp-apollo-server/src/server.rs
```diff
@@ -85,22 +124,101 @@ impl Server {
...
+                                        );
+                                        retained_peers.push(peer.clone());
+                                    }
+                                }
+                                _ => {
+                                    error!("Failed to notify peer of tool list change {:?}", e);
+                                    retained_peers.push(peer.clone());
+                                }
+                            }
+                        }
```
nicholascioli: Not sure if this makes it more legible, but you could probably do the following:

```suggestion
                    match peer.notify_tool_list_changed().await {
                        Ok(_) => retained_peers.push(peer.clone()),
                        Err(ServiceError::Transport(e)) if e.get_ref().is_some() => {
                            if e.to_string() == *"disconnected" {
                                // TODO: this always gets a "disconnected" error due to a bug in the SDK, but it actually works
                                retained_peers.push(peer.clone());
                            } else {
                                error!(
                                    "Failed to notify peer of tool list change: {:?} - dropping peer",
                                    e
                                );
                            }
                        }
                        Err(other) => {
                            error!("Failed to notify peer of tool list change {:?}", other);
                            retained_peers.push(peer.clone());
                        }
                    }
```

PR #33, opened 2025-05-01
Subscription and mutation handling

Jephuff:
This is a starting point for how we can handle mutation and subscriptions. We may want to re think some things for the introspection tool, but hopefully this gives us a starting point. 

- adds configuration for allowing mutations with 3 modes
    - All -> allows mutations everywhere
    - Explicit -> Will error if the LLM tries to run a mutation
    - None -> will error both when the LLM tries to run a mutation, and if the operation list pasted in includes a mutation
- Introspection tool uses a subset of the schema only including the schema that it is allowed to run
    - Subscription type and types only referenced by subscriptions always removed
    - Mutation type and types only referenced by mutations removed if mutation mode isn't ALL


crates/mcp-apollo-server/src/errors.rs
```diff
@@ -25,6 +29,12 @@ pub enum OperationError {
 
     #[error(transparent)]
     File(#[from] std::io::Error),
+
+    #[error("Mutations are not allowed: {0}")]
+    MutationNotAllowed(Node<OperationDefinition>),
+
+    #[error("Subscriptions are not allowed: {0}")]
```
pubmodmatt, line 40: I wonder if this should say "...not allowed by configuration" so people don't think the server isn't capable of this at all

crates/mcp-apollo-server/src/schema_tree_shake.rs
```diff
@@ -0,0 +1,465 @@
+//! Tree shaking for GraphQL schema types
```
pubmodmatt: Having both a `tree_shake` and `schema_tree_shake` (with the same comment even) is confusing. The other one is also for schema types...

crates/mcp-apollo-server/src/errors.rs
Jephuff: Mutations definitely , I'll update it. 

I think the message does make sense for subscriptions though because we don't support those ever?

crates/mcp-apollo-server/src/schema_tree_shake.rs
Jephuff: yeah, names here definitely leave something to be desired. I think I should be able to move the other use of tree shaking to this document based version. Maybe I'll try that this morning and see if we can just have 1 version.

crates/mcp-apollo-server/src/errors.rs
pubmodmatt: Right 👍

crates/mcp-apollo-server/src/schema_tree_shake.rs
```diff
@@ -0,0 +1,849 @@
...
+                        })
+                        .and_then(|retained| retained.then_some(def)),
+                })
+                .collect();
+
+        let mut document = Document::new();
+        document.definitions = filtered_definitions;
+        document
+            .to_schema_validate()
+            .map_err(|e| ServerError::GraphQLSchema(Box::new(e)))
```
pubmodmatt: If a schema type was not found, previously it would just warn. With these changes, the application fails to start here:

```
Error: Could not parse GraphQL schema: Error: cannot find type `InputCoordinate` in this document
Unable to fetch source '': source file not found: 3
```

Jephuff: I think anything like that would be a bug in the tree shaking, I've got some spots I missed so I'm fixing those now. 
But I'll also return the validation result from the tree shaker, and instead of the error we can grab the schema with error and warn 👍 

what was the operation/schema pair you had to get this? 👀

pubmodmatt: I'm starting the server with the mcp-example@mcp server graph ref, which currently has operations in the PQ list that actually aren't in the contrat variant. That's incorrect, because the input type is not in the contract variant, but it used to handle it more gracefully.

crates/mcp-apollo-server/src/server.rs
```diff
@@ -134,7 +137,11 @@ struct Starting {
 
 impl Starting {
     /// Run the MCP server once the schema is ready
-    async fn run(self, schema: Valid<Schema>) -> Result<Running, ServerError> {
+    async fn run(
+        self,
+        schema: Valid<Schema>,
+        mutation_mode: MutationMode,
```
pubmodmatt: This should be part of the Starting struct, not passed in

crates/mcp-apollo-server/src/server.rs
```diff
@@ -227,18 +235,23 @@ struct Running {
...
+    ) -> Result<Running, ServerError> {
         info!("Schema updated:\n{}", schema);
 
         // Update the operations based on the new schema. This is necessary because the MCP tool
         // input schemas and description are derived from the schema.
         let operations = self
             .operation_poller
-            .operations(&schema, self.custom_scalar_map.as_ref())
+            .operations(&schema, self.custom_scalar_map.as_ref(), mutation_mode)
```
pubmodmatt: Here too, since we know mutation mode at startup, it can be put into the OperationPoller directly rather than passed in each time.

crates/mcp-apollo-server/src/main.rs
```diff
@@ -74,6 +76,10 @@ struct Args {
     /// The path to the persisted query manifest containing operations
     #[arg(long)]
     manifest: Option<PathBuf>,
+
+    // Configure when to allow mutations
+    #[clap(long, short = 'm', default_value_t, value_enum)]
+    allow_mutations: MutationMode,
```
pubmodmatt, line 82: Nit - this is a little verbose, maybe just `mutations`?

crates/mcp-apollo-server/src/schema_tree_shake.rs
```diff
@@ -0,0 +1,927 @@
...
+    directive_nodes: HashMap<String, TreeNode>,
+    operation_types: Vec<OperationType>,
+    operation_type_names: RootOperationNames,
+    named_fragments: HashMap<String, Node<FragmentDefinition>>,
+}
+
+#[derive(Clone, Default)]
+struct TreeNode {
+    referenced_type_names: Vec<String>,
+    referected_directive_names: Vec<String>,
```
pubmodmatt: referected?

crates/mcp-apollo-server/src/schema_tree_shake.rs
```diff
@@ -0,0 +1,927 @@
...
+}
+
+fn retain_directive(
+    directive_name: &str,
+    named_type_nodes: &mut HashMap<String, TreeNode>,
+    directive_nodes: &mut HashMap<String, TreeNode>,
+    named_fragments: &HashMap<String, Node<FragmentDefinition>>,
+    schema: &Schema,
+) {
+    // let type_name = type_def.name().as_str();
```
pubmodmatt: why leave this commented code?

crates/mcp-apollo-server/src/server.rs
Jephuff: Makes sense! OperationPoller is an enum though, how do I add a value to it 👀

crates/mcp-apollo-server/src/main.rs
Jephuff: My 1 worry is that mutations sounds like it might be passing in a list of mutations rather then configuring if they are allowed?  Definitely can rename though!

crates/mcp-apollo-server/src/schema_tree_shake.rs
Jephuff: This has been removed in the latest commit. But also, thank you for adding the spell checker in that other PR :D

crates/mcp-apollo-server/src/schema_tree_shake.rs
Jephuff: I'll 🧹  that up!

crates/mcp-apollo-server/src/errors.rs
```diff
@@ -27,6 +33,12 @@ pub enum OperationError {
 
     #[error(transparent)]
     File(#[from] std::io::Error),
+
+    #[error("Mutation {0} not allowew while with --allow-mutations {1:?}")]
```
pubmodmatt:
```suggestion
    #[error("Mutation {0} not allowed while with --allow-mutations {1:?}")]
```

crates/mcp-apollo-server/src/main.rs
pubmodmatt: Make sense.

PR #36, opened 2025-05-05
Fix tool to open Apollo Explorer

pubmodmatt:
Fixes:
* There were intermittent encoding problems with the URL, due to use of the wrong compression method
* If the LLM sent invalid input, it would try to use it and fail. The input is now validated. Missing fields are defaulted.

### DIFF
--- a/crates/mcp-apollo-server/src/explorer.rs
+++ b/crates/mcp-apollo-server/src/explorer.rs
@@ -4,8 +4,10 @@ use rmcp::model::{CallToolResult, Content, ErrorCode, Tool};
 use rmcp::schemars::JsonSchema;
 use rmcp::serde_json::Value;
 use rmcp::{schemars, serde_json};
-use serde::Deserialize;
+use serde::{Deserialize, Serialize};
 use tracing::info;
+use tracing::log::Level::Info;
+use tracing::log::log_enabled;
 
 pub(crate) const EXPLORER_TOOL_NAME: &str = "explorer";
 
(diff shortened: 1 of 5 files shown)


crates/mcp-apollo-server/src/explorer.rs
```diff
@@ -43,17 +47,42 @@ impl Explorer {
     }
 
     fn create_explorer_url(&self, input: &Value) -> String {
-        let compressed = lz_str::compress_to_uint8_array(input.to_string().as_str());
-        let encoded = base64::engine::general_purpose::STANDARD.encode(compressed);
+        let mut input = input.clone();
```
nicholascioli: Could this take an owned `Value` to avoid the clone here?

crates/mcp-apollo-server/src/explorer.rs
```diff
@@ -43,17 +47,42 @@ impl Explorer {
...
     }
 
     pub async fn execute(&self, input: Value) -> Result<CallToolResult, McpError> {
-        webbrowser::open(self.create_explorer_url(&input).as_str())
+        let url = self.create_explorer_url(&input);
+        info!(
+            "Opening Apollo Explorer URL: {} for input operation: {}",
+            url,
+            serde_json::to_string_pretty(&input).unwrap_or("<unable to serialize>".into())
+        );
```
nicholascioli: `serde_json::to_string_pretty`already alocates, so if you shuffle these around you can avoid a clone.

```suggestion
        let pretty = serde_json::to_string_pretty(&input).unwrap_or_else(|| "<unable to serialize>".into());
        let url = self.create_explorer_url(input);
        info!(
            "Opening Apollo Explorer URL: {url} for input operation: {pretty}",
        );
```

crates/mcp-apollo-server/src/explorer.rs
```diff
@@ -43,17 +47,42 @@ impl Explorer {
...
+            if let Some(obj) = input.as_object_mut() {
+                obj.insert("variables".to_string(), Value::String("{}".to_string()));
+            }
+        }
+        let headers = input.get("headers").and_then(|v| v.as_str());
+        if headers.is_none() || headers == Some("") {
+            if let Some(obj) = input.as_object_mut() {
+                obj.insert("headers".to_string(), Value::String("{}".to_string()));
+            }
+        }
```
nicholascioli: Would it make sense to just use serialize / deserialize the `Input` struct with default values if empty? The struct is already defined above and using it would allow us to remove the `#[allow(dead)]`.

nicholascioli (approved):
LGTM! I left a few nits, but feel free to ignore.

PR #37, opened 2025-05-05
simplify CI pipeline

nicholascioli:
This commit simplifies the nix CI pipeline by depending on upstream nixpkgs version of rust and cargo. This no longer requires us to manually compile rust and then cache that build.


nicholascioli:
This is no longer needed since we now use [oxalica/rust-overlay](https://github.com/oxalica/rust-overlay) for rust tooling.

PR #39, opened 2025-05-07
Add launch configurations for VS Code and Cursor

pubmodmatt:


.vscode/extensions.json
```diff
@@ -1,5 +1,8 @@
 {
     "recommendations": [
-        "mkhl.direnv"
+        "mkhl.direnv",
+        "vadimcn.vscode-lldb",
+        "streetsidesoftware.code-spell-checker",
+        "apollographql.vscode-apollo"
```
Jephuff, line 6: Our extension doesn't seem to identify operation files and gives a whole lot of errors if you open any graphql file. are you seeing the same thing?

Jephuff (approved):

.vscode/extensions.json
pubmodmatt: Yes

PR #40, opened 2025-05-08
add option to disable type descriptions or schema descriptions

Jephuff:
https://apollographql.atlassian.net/browse/GT-116
Add an option to disable schema or type descriptions. Not sure these are the best names for the options so very open to suggestions! :)

### DIFF
--- a/crates/apollo-mcp-server/src/main.rs
+++ b/crates/apollo-mcp-server/src/main.rs
@@ -81,13 +81,13 @@ struct Args {
     #[clap(long, short = 'm', default_value_t, value_enum)]
     allow_mutations: MutationMode,
 
-    /// Use type descriptions for tool descriptions
+    /// Disable operation root field types in tool description
     #[arg(long)]
-    type_description: bool,
+    disable_type_description: bool,
 
-    /// Use schema types in descriptions for tool descriptions
+    /// Disable schema type definitions referenced by all fields returned by the operation in the tool description
     #[arg(long)]
-    schema_description: bool,
+    disable_schema_description: bool,
 }
 
 #[tokio::main]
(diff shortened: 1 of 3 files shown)


crates/mcp-apollo-server/src/main.rs
```diff
@@ -80,6 +80,14 @@ struct Args {
     // Configure when to allow mutations
     #[clap(long, short = 'm', default_value_t, value_enum)]
     allow_mutations: MutationMode,
+
+    /// Use type descriptions for tool descriptions
```
pubmodmatt: I'm not sure it's clear enough what this does. Maybe something like "Incude operation root field type in tool description".

crates/mcp-apollo-server/src/main.rs
```diff
@@ -80,6 +80,14 @@ struct Args {
...
+    #[arg(long)]
+    type_description: bool,
+
+    /// Use schema types in descriptions for tool descriptions
```
pubmodmatt: Here maybe: "Include schema type definitions referenced by all fields returned by the operation in the tool description"

crates/mcp-apollo-server/src/main.rs
```diff
@@ -141,6 +149,8 @@ async fn main() -> anyhow::Result<()> {
         .headers(default_headers)
         .introspection(args.introspection)
         .mutation_mode(args.allow_mutations)
+        .type_description(args.type_description)
+        .schema_description(args.schema_description)
```
pubmodmatt: We may want to default these to `true` until we have a better way to introspect operations. If the user doesn't provide a very detailed description of what the operation does, the LLM will have difficulty without the type information being present.

Jephuff: Should these maybe be inverted and be `--disable-schema-description` and `--disable-type-description` instead?

pubmodmatt: That sounds fine.

pubmodmatt (approved):

PR #44, opened 2025-05-08
Create an introspection tool to get schema type information

pubmodmatt:
The `schema` tool is now replaced with `introspect`. Rather than returning the *entire* schema at once (which would not work well with large schemas), this new tool allows the LLM to introspect types progressively. It can start by asking about `Query` or `Mutation` (if mutations are enabled in the MCP server).

There is a depth parameter, which allows the LLM to decide how far to recurse types in a single call. An agent can be configured to prompt it to be more or less aggressive on the depth, which will reduce or increase the number of tool calls needed to obtain the necessary type information. For example, I have the following configured in my Claude Desktop preferences: "When using the `introspect` tool from Apollo, you should minimize the number of calls by preferring to use a larger depth argument."

Note that although GraphQL natively supports introspection, it is often disabled, especially in production environments. This tool allows exposing the type information just to the LLM through the MCP Server, rather than through the router endpoint. It additionally allows the depth to be specified to allow bulk acquisition of type information.


crates/mcp-apollo-server/src/schema_tree_shake.rs
```diff
@@ -497,14 +528,22 @@ fn selection_set_to_fields(
...
+#[allow(clippy::too_many_arguments)]
 fn retain_type(
     extended_type: &ExtendedType,
     selection_set: Option<&Vec<Selection>>,
     named_type_nodes: &mut HashMap<String, TreeTypeNode>,
     directive_nodes: &mut HashMap<String, TreeDirectiveNode>,
     named_fragments: &HashMap<String, Node<FragmentDefinition>>,
     schema: &Schema,
+    max_depth: u32,
+    current_depth: u32,
```
Jephuff: I don't think we need to args here, we should be able to track this as 1 depth argument, and just decrement it when passing it instead of incrementing it. so the check if it should early return would be `depth == 0` and then when passing it we would pass ` depth - 1`. 

It looks like a max_depth of `0` here would also return the whole schema, which  could still be returned with `-1`.

crates/mcp-apollo-server/src/schema_tree_shake.rs
```diff
@@ -542,6 +581,8 @@ fn retain_type(
             directive_nodes,
             named_fragments,
             schema,
+            max_depth,
+            current_depth,
```
Jephuff: Are we skipping incrementing the depth because theres currently no way to introspect directives? Maybe just toss a comment on the calls to retain directive?

crates/mcp-apollo-server/src/schema_tree_shake.rs
```diff
@@ -625,6 +670,8 @@ fn retain_type(
                                     directive_nodes,
                                     named_fragments,
                                     schema,
+                                    max_depth,
+                                    current_depth + 1,
```
Jephuff: Should we increment depth here, or should we skip it like the other retain directive?  (this would apply to all of the retain_directives below this as well)

crates/mcp-apollo-server/src/introspection.rs
```diff
@@ -4,39 +4,149 @@ use crate::errors::McpError;
...
         Self {
+            mutation_mode,
             schema,
             tool: Tool::new(
-                GET_SCHEMA_TOOL_NAME,
-                "Get the GraphQL schema. Operations on this schema can be executed using the `execute` tool.",
-                schema_from_type!(GetSchemaInput),
+                INTROSPECT_TOOL_NAME,
+                format!(
+                    "Get detailed information about specific types from the GraphQL schema. Use `{QUERY_TYPE_NAME}`{} to get available root fields - the `$` character is important.",
```
Jephuff: Should we just send the actual query name so we don't have to add the bit about the $ being important? 
```suggestion
                    "Get detailed information about specific types from the GraphQL schema. Use `{schema}`{} to get available root fields - the `$` character is important.",
```

crates/mcp-apollo-server/src/schema_tree_shake.rs
```diff
@@ -129,13 +130,39 @@ impl<'schema> SchemaTreeShaker<'schema> {
...
             tracing::error!("root operation type {} not found in schema", operation_type);
         }
     }
 
-    pub fn retain_operation(&mut self, operation: &OperationDefinition, document: &Document) {
+    /// Retain a specific type, and recursively every type it references, up to a given depth.
+    pub fn retain_type(
+        &mut self,
+        retain: &ExtendedType,
+        selection_set: Option<&Vec<Selection>>,
```
Jephuff: `selection_set` might have some complications in this mode, maybe we should just omit the argument for now and pass None to the inner retain type for now?

crates/mcp-apollo-server/src/introspection.rs
```diff
@@ -4,39 +4,149 @@ use crate::errors::McpError;
...
+        let mut tree_shaker = SchemaTreeShaker::new(&schema);
+        match schema.types.get(type_name) {
+            Some(extended_type) => tree_shaker.retain_type(extended_type, None, input.depth),
+            None => {
+                return Ok(CallToolResult {
+                    content: vec![],
+                    is_error: None,
+                });
+            }
+        }
```
Jephuff: Should this happen inside `tree_shaker.retain_type`? then retain_type would just take a type name, and return a result? 
No strong feelings here if you want to keep it as is though 👍

crates/mcp-apollo-server/src/introspection.rs
pubmodmatt: We need to disambiguate. It's an edge case, but the schema could define the query root item to be Foo, and have a different type named Query.

crates/mcp-apollo-server/src/schema_tree_shake.rs
pubmodmatt: We'll need this for operation introspection later. What difficulties are you concerned about?

crates/mcp-apollo-server/src/introspection.rs
pubmodmatt: None of the other retain methods are fallible, so passing in the actual type seems more consistent to me.

crates/mcp-apollo-server/src/introspection.rs
Jephuff: right, but we could encode that in the text here?  (just realized I put a very incomplete suggestion up there haha)

So it could say something like `Use Foo to get available root fields` and it would remove the need to special case  in execute. 
```suggestion
                    "Get detailed information about specific types from the GraphQL schema. Use `{SCHEMA.root_operation(OperationType::Query).expect("handle this case of course, just an example").as_str()}`{} to get available root fields",
```

crates/mcp-apollo-server/src/schema_tree_shake.rs
Jephuff: In any operation there would be multiple different selection sets at different depths in the schema. It might not be an issue, but we would need to test to make sure the LLM doesn't think "i've got the type, no need to ask for more". It might be smart enough to figure out it's missing parts of the selection and rerequest, I just wonder if we might want to test. 

We might also not want it to request the same type over and over returning different subsets, so we might want to preshake the full operation, so we can return `type User { id name age city }` instead of `type User { id }`, `type User { name }`, `type User { age }`, and `type User { city }` all separate.

crates/mcp-apollo-server/src/introspection.rs
Jephuff: makes sense to me! 👍

crates/mcp-apollo-server/src/schema_tree_shake.rs
pubmodmatt: Got it. Yeah, at some point along the way when it was shaking the full tree there was code to take the union of the selection sets, but here since its taking each type one at a time, it'd have to pre-compute. Ok, I'll remove this for now!

Jephuff: Yeah, the `retain_operation` will do a union of the selection sets right now. So what we might want is to add a new version of `shaken` that takes a type and returns it's depth.

crates/mcp-apollo-server/src/introspection.rs
pubmodmatt: Gotcha, I can change this.

crates/mcp-apollo-server/src/schema_tree_shake.rs
```diff
@@ -994,7 +1064,7 @@ mod test {
         let mut shaker = SchemaTreeShaker::new(&schema);
         let (operation_document, operation_def, _comments) =
             operation_defs("query TestQuery { id }", false, MutationMode::None).unwrap();
-        shaker.retain_operation(&operation_def, &operation_document);
+        shaker.retain_operation(&operation_def, &operation_document, DepthLimit::Unlimited);
```
Jephuff, line 1067: Could you add a test for the happy path of depth limit?

Jephuff (approved):
Looks good! only request is adding a test that hits a depth limit

crates/mcp-apollo-server/src/schema_tree_shake.rs
pubmodmatt: I'll create a new PR for that, so we can get this in and unblock work

PR #45, opened 2025-05-09
split execute and instrospection

Jephuff:
Going to wait and rebase on https://github.com/apollographql/apollo-mcp-server/pull/44

### DIFF
--- a/crates/mcp-apollo-server/src/introspection.rs
+++ b/crates/mcp-apollo-server/src/introspection.rs
@@ -27,12 +27,24 @@ pub struct GetSchema {
 pub struct GetSchemaInput {}
 
 impl GetSchema {
-    pub fn new(schema: Arc<Mutex<Valid<Schema>>>) -> Self {
+    pub fn new(
+        schema: Arc<Mutex<Valid<Schema>>>,
+        execute_introspection: bool,
+        enable_explorer: bool,
+    ) -> Self {
         Self {
             schema,
             tool: Tool::new(
                 GET_SCHEMA_TOOL_NAME,
-                "Get the GraphQL schema. Operations on this schema can be executed using the `execute` tool.",
+                if execute_introspection && enable_explorer {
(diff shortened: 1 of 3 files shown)


pubmodmatt (approved):

pubmodmatt (changes requested):
Actually, we need to add this to the docs

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### 🚀 Features
 - Add an optional `--sse-address` argument to set the bind address of the MCP server. Defaults to 127.0.0.1. (#63)
+- split out the execute tool into separate --execute-introspection flag, `--introspection` is now `--introspection --execute-introspection`
```
pubmodmatt: Technically, this is a breaking change - there's a separate changelog header for that

Jephuff:
Is there a docs preview generated somewhere 👀

CHANGELOG.md
```diff
@@ -13,6 +13,11 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### 📚 Documentation
 -->
 
+## [UNRELEASED]
+
+### ❗ BREAKING ❗
+- split out the execute tool into separate --execute-introspection flag, `--introspection` is now `--introspection --execute-introspection`
+
```
pubmodmatt, line 20: I'm starting to have second thoughts about this. It may confuse people if this stops working, and it's yet more options they have to enable. I wonder if we could leave `---introspection` as is, but then add `--introspect` and `---execute` to override and enable/disable individual tools? The clients also generally have a way to disable specific tools.

pubmodmatt: Also, we have the same option in rover, and would need an update there.

Jephuff: Chatted about this in standup but, but for the history I agree! we are going to try to do a backwards compatible enum here and as a fallback we could add a `--disable-execute` instead

PR #47, opened 2025-05-09
Add TheSpaceDevs example folder into the repository

michael-watson:


graphql/TheSpaceDevs/.vscode/tasks.json
```diff
@@ -0,0 +1,29 @@
+{
```
pubmodmatt, line 1: I don't think these `.vscode` files do anything at this level of the workspace?

graphql/TheSpaceDevs/apollo.config.json
```diff
@@ -0,0 +1,9 @@
+{
```
pubmodmatt, line 1: Same with this - there's already one of these at the top level

michael-watson: As it stands today, the `apollo.config.json` can't include configuration for both `rover` and the client settings. The Space Devs graph doesn't show a developer workflow of creating a Connectors graph, rather just a GraphQL API endpoint. The `apollo.config.json` contains the necessary config for our VS Code extension that enables type completion based on the schema you have when writing operations in VS Code

Screenshot 2025-05-11 at 10 40 29 PM

I think this is part of the developer tooling that we want to showcase. I think a future state should enable the `apollo.config.json` to have both configurations in it and then the folder vs code specific things could be moved to the root.

graphql/TheSpaceDevs/.vscode/tasks.json
michael-watson: They don't do anything unless you open the `/graphql/TheSpaceDevs` folder in VSCode. But I personally think we should keep that experience because we have some current conflicts that don't show the full power of VS Code. 

Specifically, the `apollo.config.json` and because of that, I think we should keep this. See that comment

graphql/TheSpaceDevs/apollo.config.json
michael-watson: It could also be that we create a different content piece to highlight this and omit this from this repository. Definitely should be something that is docs when we have it. I can go either way.

kbychu (approved):
LGTM

PR #51, opened 2025-05-09
Add tests for schema shaker depth limit

pubmodmatt:


crates/apollo-mcp-server/src/schema_tree_shake.rs
```diff
@@ -200,30 +200,20 @@ impl<'schema> SchemaTreeShaker<'schema> {
...
-        }
-        if self.operation_types.contains(&OperationType::Subscription) {
-            if let Some(subscription_name) = self.schema.schema_definition.subscription.clone() {
-                filtered_root_operations.push(Node::new((
-                    OperationType::Subscription,
-                    subscription_name.name,
-                )));
-            }
-        }
+        let root_operations = self
```
Jephuff, line 203: Hmm, this change I believe is going to make invalid schemas, a query type is required. 
Now that we don't return a Valid<Schema> maybe that ok though 🤔

Jephuff (approved):

crates/apollo-mcp-server/src/schema_tree_shake.rs
pubmodmatt: It inserts a stub below if the Query is not included.

PR #62, opened 2025-05-16
add multiarch container

nicholascioli:
This PR adds support for multiarch container images.


pubmodmatt (approved):
Maybe add a step to the RELEASE_CHECKLIST.md to run a docker command that verifies the latest version of the image? Also, should we have a workflow_dispatch on release-bins.yml as well? Otherwise, looks good!

PR #63, opened 2025-05-19
Add an argument to set the bind address

pubmodmatt:


crates/apollo-mcp-server/src/main.rs
```diff
@@ -53,7 +54,11 @@ struct Args {
     #[arg(long = "header", action = clap::ArgAction::Append)]
     headers: Vec<String>,
 
-    /// Start the server using the SSE transport on the given port
+    /// The IP address to bind the SSE server to (default: 127.0.0.1)
+    #[arg(long)]
+    sse_address: Option<IpAddr>,
```
nicholascioli, line 59: Could we use clap defaults here instead of wrapping this in an `Option` that needs to be `unwrap_or` later?

nicholascioli (approved):
Looks great! I left a small nit, but feel free to ignore.

crates/apollo-mcp-server/src/main.rs
pubmodmatt: We use the presence of the port and address options as indicators of whether to use SSE mode, so we can't default them at the clap level as we'd lose that signal. We could alternatively add a boolean `--sse` option and then let the address/ports default and be unused if `false`. That's be a breaking change at this point, though (but ok since we're still "experimental").

pubmodmatt: I'll merge this as is, but we can revisit.

PR #69, opened 2025-05-21
Hot reload operation files and directories

pubmodmatt:
* The `--operations` option now support specifying a directory. All `.graphql` files in the directory will be loaded as operations.
* The `--operations` option now supports hot reloading. If a directory is specified, the MCP Server will handle files being added or removed, as well as changes to individual files. If a file is specified, it will hot reload any changes to that file.
* The event streaming has been unified - instead of having multiple spawned tasks to pick up PQ events separately from schema events, there is now one unified event stream. This required refactoring the server states to add configuration states. Only once both schema and operations events have been received does the server enter the `Starting` state.

### DIFF
--- a/crates/apollo-mcp-registry/src/uplink/persisted_queries/event.rs
+++ b/crates/apollo-mcp-registry/src/uplink/persisted_queries/event.rs
@@ -1,0 +1,18 @@
+use std::fmt::Debug;
+use std::fmt::Formatter;
+
+/// Persisted Query events
+pub enum Event {
+    /// The persisted query manifest was updated
+    UpdateManifest(Vec<(String, String)>),
+}
+
+impl Debug for Event {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Event::UpdateManifest(_) => {
+                write!(f, "UpdateManifest(<redacted>)")
+            }
+        }
+    }
+}
--- a/crates/apollo-mcp-server/src/operations.rs
+++ b/crates/apollo-mcp-server/src/operations.rs
@@ -1,5 +1,6 @@
 use crate::custom_scalar_map::CustomScalarMap;
 use crate::errors::{McpError, OperationError};
+use crate::event::Event;
 use crate::graphql;
 use crate::schema_tree_shake::{DepthLimit, SchemaTreeShaker};
 use apollo_compiler::ast::{Document, OperationType, Selection};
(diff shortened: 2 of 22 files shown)


Meschreiber (approved):

crates/apollo-mcp-registry/src/uplink/persisted_queries/event.rs
nicholascioli, line 1: Might be worth doing something similar to Router here and wrapping any sensitive string in a newtype called `SecretString` or something. Then we could implement Debug and Display on that to redact the information and not have to worry about leaking the value accidentally anywhere else.

crates/apollo-mcp-server/src/operations.rs
```diff
@@ -23,89 +25,98 @@ use rmcp::{
...
+                            (!content.trim().is_empty()).then(|| RawOperation::from(content))
+                        })
+                    })
+                    .collect::<Vec<_>>();
+                Event::OperationsUpdated(operations.clone())
+            })
+        }))
+        .boxed()
+    }
+}
```
nicholascioli, line 109: This seems like it might be worth refactoring `files::watch` so that it emits the paths that have changed. That way we can remove many of these clones.

crates/apollo-mcp-server/src/server.rs
```diff
@@ -311,6 +496,40 @@ impl Running {
...
+    async fn update_operations(
+        self,
+        operations: Vec<RawOperation>,
+    ) -> Result<Running, ServerError> {
+        // Update the operations based on the current schema
+        {
+            let schema = &*self.schema.lock().await;
+            let updated_operations: Vec<Operation> = operations
+                .iter()
+                .cloned()
```
nicholascioli:
```suggestion
            let updated_operations: Vec<Operation> = operations
                .into_iter()
```

nicholascioli (approved):
Left a few nits, but feel free to ignore!

crates/apollo-mcp-registry/src/uplink/persisted_queries/event.rs
pubmodmatt: I can't see where this comment was anymore, but there's an entire secrecy dependency we use for that in the MCP server - router didn't do any of that.

crates/apollo-mcp-server/src/operations.rs
pubmodmatt: I'm trying to minimize changes from router until we can refactor this code out of router.

PR #74, opened 2025-05-21
fix recursive schema handling

Jephuff:
This fixes the schema shaker overflowing the stack if the schema is recursive. 
https://apollographql.atlassian.net/browse/GT-135
```
type Query { id: TypeA }
type TypeA { id: TypeB }
type TypeB { id: TypeA }
```

### DIFF
--- a/crates/apollo-mcp-server/src/schema_tree_shake.rs
+++ b/crates/apollo-mcp-server/src/schema_tree_shake.rs
@@ -68,6 +68,13 @@ impl DepthLimit {
             DepthLimit::Limited(depth) => DepthLimit::Limited(depth - 1),
         }
     }
+
+    pub fn is_unlimited(&self) -> bool {
+        match self {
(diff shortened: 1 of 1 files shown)


crates/apollo-mcp-server/src/schema_tree_shake.rs
```diff
@@ -68,6 +68,13 @@ impl DepthLimit {
             DepthLimit::Limited(depth) => DepthLimit::Limited(depth - 1),
         }
     }
+
+    pub fn is_unlimited(&self) -> bool {
```
pubmodmatt: Rust already has the [`matches!` macro](https://doc.rust-lang.org/std/macro.matches.html), you don't need a method for this.

Jephuff: oohh nice!

pubmodmatt (approved):

PR #82, opened 2025-05-23
Add log level option and reduce logging at INFO

pubmodmatt:


CHANGELOG.md
```diff
@@ -13,6 +13,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### 📚 Documentation
 -->
 
+## [Unreleased]
+### 🚀 Features
+- Add a `--log` option to specify the log level used by the MCP Server (default is INFO). Reduce the level of many messages emitted by the server so INFO is less verbose. (#XX)
```
DaleSeo: I guess this is not intentional?

```suggestion
- Add a `--log` option to specify the log level used by the MCP Server (default is INFO). Reduce the level of many messages emitted by the server so INFO is less verbose. (#82)
```

crates/apollo-mcp-server/src/server.rs
```diff
@@ -198,7 +198,7 @@ struct Configuring {
 
 impl Configuring {
     async fn set_schema(self, schema: Valid<Schema>) -> Result<SchemaConfigured, ServerError> {
-        info!("Received schema:\n{}", schema);
+        debug!("Received schema:\n{}", schema);
```
DaleSeo, line 201: I had to comment this out today because the schema was too large and filled up my terminal buffer. 😂

DaleSeo (approved):
I love this new CLI flag! 😍

crates/apollo-mcp-server/src/main.rs
```diff
@@ -92,6 +92,10 @@ struct Args {
     /// Disable schema type definitions referenced by all fields returned by the operation in the tool description
     #[arg(long)]
     disable_schema_description: bool,
+
+    /// The log level for the MCP Server (default: INFO)
+    #[arg(long = "log", short = 'l', global = true)]
+    log_level: Option<Level>,
```
Jephuff: I think if we use default_value_t it will handle the default value in `--help` for us as well so we don't need to worry about keeping them in sync in the future if we change it :) 
```suggestion
    /// The log level for the MCP Server
    #[arg(long = "log", short = 'l', global = true, default_value_t = Level::INFO)]
    log_level: Level,
```

crates/apollo-mcp-server/src/server.rs
pubmodmatt: Thanks, I just got pulled into a meeting before I had a chance to update it!

PR #84, opened 2025-05-23
Improving the MCP quick start

kbychu:
- Remove the unnecessary option of building the MCP server
- Add expectation setting up front
- Add troubleshooting guide (thanks AI!)
- Add some additional meaningful context


docs/source/quickstart.mdx
```diff
@@ -23,106 +38,173 @@ If you learn best with videos and exercises, this [interactive course](https://w
...
-    ```
-1. Open a browser and go to the Inspector URL at [`http://127.0.0.1:6274`](http://127.0.0.1:6274).
-1. Fill in the details in Inspector:
-    - **Transport Type**: Select `STDIO`
-    - **Command**: Should be prepopulated with the command to run the MCP Server
+### Common Issues
+
+#### MCP Server Won't Start
+- **Error**: "Port 5000 is already in use"
+  - Solution: Kill any existing processes using port 5000 or specify a different port with `--port` flag
```
lizhennessy: I noticed earlier that I tried `--port` and got an error message.  `--mcp-sse-port` worked.

docs/source/quickstart.mdx
```diff
@@ -23,106 +38,173 @@ If you learn best with videos and exercises, this [interactive course](https://w
...
+- **Error**: "Failed to load supergraph configuration"
+  - Solution: Verify you're running the command from the repo root directory
+  - Solution: Check that the path to `supergraph.yaml` is correct
 
-1. Click **Connect**, then click **List Tool**. Inspector should show the tools from your server.
+#### MCP Inspector Connection Issues
+- **Error**: "Failed to connect to server"
+  - Solution: Ensure the MCP server is running (check terminal output)
+  - Solution: Verify you're using the correct URL (`http://localhost:5000/sse`)
```
lizhennessy: Not sure if it was just me, but only `http://127.0.0.1:5000/sse` worked. Localhost as an alias failed with a Connection Error - tried two machines and it was the same result

kbychu: Hmmm localhost works for me. @pubmodmatt Should we say localhost or the ip address?

pubmodmatt: The IP address is probably more reliable, since localhost will depend on the contents of /etc/hosts. Also, let's get rid of any reference to SSE and replace it with Streamable HTTP. The path is now `/mcp` instead of `/sse`.

pubmodmatt: To be clear, use this URL:
```
http://127.0.0.1:5000/mcp
```

kbychu: Maybe let's dos a scrub of sse separate from this PR. It's probably everywhere. I know we also say things like streamable support is coming soon.

pubmodmatt: I already updated all the docs to remove SSE references. That already merged, so this would add the SSE reference back in. This PR has merge conflicts with those changes that you'll need to resolve.

kbychu: Ah ok! Will do

docs/source/command-reference.mdx
```diff
@@ -85,7 +85,7 @@ Specifying either the HTTP port or address (or both) will enable the Streamable
 
 ### Mapping rover dev options
 
-You can use the [`rover dev`](/rover/commands/dev) command of Rover CLI v0.31 or later to run an Apollo MCP Server instance for local development. 
+You can use the [`rover dev`](/rover/commands/dev) command of Rover CLI v0.32 or later to run an Apollo MCP Server instance for local development.
```
kbychu, line 123: 0.32 is not technically required but it is recommended since we recommend Streamable HTTP. Are you ok with the change @pubmodmatt

pubmodmatt: Yes, I think that's a good idea!

lizhennessy (approved):

PR #91, opened 2025-05-27
Ignore mutation and subscription operations rather than erroring

pubmodmatt:
The MCP Server would error out if mutation or subscription operations were specifed but not allowed. It should be possible to have such operations in a PQ list or operation files, but just not use them in the MCP Server. For example, the user may want to define mutations, but turn them on or off using the `--allow-mutations` option.

* Subscription operations are now simply ignored as they are unsupported by the MCP server
* Mutation operations are ignored with a warning if the specified mutation mode does not allow them
* The `introspect` tool will not return mutations if the mutation mode does not allow them

### DIFF
--- a/crates/apollo-mcp-server/src/operations.rs
+++ b/crates/apollo-mcp-server/src/operations.rs
@@ -29,7 +29,7 @@ use std::collections::HashMap;
 use std::fs;
 use std::path::PathBuf;
 use std::sync::{Arc, Mutex};
-use tracing::{info, warn};
+use tracing::{debug, info, warn};
 
 const OPERATION_DOCUMENT_EXTENSION: &str = "graphql";
 
@@ -192,7 +192,7 @@ impl RawOperation {
         mutation_mode: MutationMode,
         disable_type_description: bool,
         disable_schema_description: bool,
-    ) -> Result<Operation, OperationError> {
+    ) -> Result<Option<Operation>, OperationError> {
         Operation::from_document(
             &self.source_text,
             schema,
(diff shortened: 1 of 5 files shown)


crates/apollo-mcp-server/src/operations.rs
```diff
@@ -229,11 +229,11 @@ impl Operation {
     }
 }
 
+#[allow(clippy::type_complexity)]
 pub fn operation_defs(
     source_text: &str,
     allow_mutations: bool,
-    mutation_mode: MutationMode,
-) -> Result<(Document, Node<OperationDefinition>, Option<String>), OperationError> {
+) -> Result<Option<(Document, Node<OperationDefinition>, Option<String>)>, OperationError> {
```
pubmodmatt, line 236: For now, I've made this `Option`. There is a separate task open to allow multiple operations per document - that will require this to be a `Vec`. Once that is implemented, this may just return an empty `Vec` if the document only contained a mutation and they are not allowed. But using a `Vec` here required too many fan-out changes for this PR.

crates/apollo-mcp-server/src/operations.rs
```diff
@@ -276,17 +276,25 @@ pub fn operation_defs(
 
     match operation.operation_type {
         OperationType::Subscription => {
-            return Err(OperationError::SubscriptionNotAllowed(operation));
+            debug!(
+                "Skipping subscription operation {}",
```
Jephuff, line 280: I wonder if this should be an `info`? Or maybe tally up the skipped subscriptions on print  the total. 
But maybe thats also unnecessary noise. so feel free to leave it on this PR!

pubmodmatt: I'm trying to reduce logging noise generally, and since the server doesn't support subscriptions at all, info seemed to noisy. But this is an easy 2-way door if we want to change it.

PR #96, opened 2025-05-28
Return JSON content and set is_error

pubmodmatt:
Since GrqphQL returns JSON, use `Content::json`. This makes the output look much better in some MCP clients, such as MCP Inspector.

Also set `is_error` based on the presence of errors in the response, following the [guildlines here](https://modelcontextprotocol.io/docs/concepts/tools#error-handling-2). 

# Before

before

# After

after

### DIFF
--- a/crates/apollo-mcp-server/src/graphql.rs
+++ b/crates/apollo-mcp-server/src/graphql.rs
@@ -55,7 +55,7 @@ pub trait Executable {
                     None,
                 )
             })?
-            .text()
+            .json::<Value>()
             .await
             .map_err(|reqwest_error| {
                 McpError::new(
@@ -64,10 +64,13 @@ pub trait Executable {
                     None,
                 )
             })
-            .and_then(Content::json)
-            .map(|result| CallToolResult {
-                content: vec![result],
-                is_error: None,
+            .map(|json| CallToolResult {
+                content: vec![Content::json(&json).unwrap_or(Content::text(json.to_string()))],
+                is_error: Some(
+                    json.get("errors")
+                        .filter(|value| !matches!(value, Value::Null))
+                        .is_some(),
+                ),
             })
     }
 }
--- a/crates/apollo-mcp-server/src/graphql.rs
+++ b/crates/apollo-mcp-server/src/graphql.rs
@@ -69,7 +69,11 @@ pub trait Executable {
                 is_error: Some(
                     json.get("errors")
                         .filter(|value| !matches!(value, Value::Null))
-                        .is_some(),
+                        .is_some()
+                        && json
+                            .get("data")
+                            .filter(|value| !matches!(value, Value::Null))
+                            .is_none(),
                 ),
             })
     }


crates/apollo-mcp-server/src/graphql.rs
```diff
@@ -64,10 +64,13 @@ pub trait Executable {
                     None,
                 )
             })
-            .and_then(Content::json)
-            .map(|result| CallToolResult {
-                content: vec![result],
-                is_error: None,
+            .map(|json| CallToolResult {
+                content: vec![Content::json(&json).unwrap_or(Content::text(json.to_string()))],
+                is_error: Some(
```
Jephuff, line 69: hmm, I'm not 100% we want to set this based on the errors array. With graphql you could have a successful operation with errors in it. 
We should probably test how the LLMs react to is_error: true when it did successfully do something. i.e. if it were a `CreateUser` tool, and the operation succeeded in creating the user, but includes a different error in the array, does the LLM react better or worse with `is_error` true

pubmodmatt: This will differ depending on the model, so I think its best to adhere to the guidance - "This allows the LLM to see and potentially handle the error". This is in line with GraphQL error responses - we're returning the data and the errors, and notifying the MCP client that there were errors.

> you could have a successful operation with errors in it

Not according to [the spec](https://spec.graphql.org/October2021/#sec-Errors) - "If no errors were raised during the request, the errors entry should not be present in the result."

Jephuff: The problem here is that this isn't a 1:1. "errors" in graphql are not also an indication that the Operation errored, studio is a great example 90% of our graphql requests contain errors in the errors array when they are successful because of we don't fetch based on permissions, we fetch what we want and allow the permission errors to guard the fields.

Definitely understand different models will treat it differently, but if we want to do this need to test this to make sure at least the major models still use partial responses and don't throw it out if is_error is true. I'm assuming they are smart enough, but we gotta test that. 

An expanded contrived example could be adding a user to an org where an extra field is requested that the user does not have permission to read. The user IS added to the org, and the error only relates to reading "homeAddress"
```
mutation AddUserToOrg($id: ID! $orgId: ID!) {
  user(id: $id) {
    addToOrg(id: $orgId) {
      id
      homeAddress
  }
}
``` 
Response
```
{
  data: {
    user: {
      addToOrg: {
        id: "user-id",
        homeAddress: null
      }
    },
    errors: [
     // permission error for homeAddress
    ],
}
```

Jephuff (approved):
Carrying over a bit of the discussion from https://apollograph.slack.com/archives/C08LW3Z8PJ4/p1748454261841939 

the tests showed that a few models don't give much notice to the is_error value so changing it is pretty low risk.

crates/apollo-mcp-server/src/graphql.rs
pubmodmatt: Your platform API permissions example convinced me. I changed it to only set `is_error` to `true` if there are `errors` *and* if `data` is null or missing.

PR #98, opened 2025-05-28
Implement Streamable HTTP transport

### LINKED ISSUE #78: Support for Streamable HTTP
Hello :) hyped by the project !

I recently saw the new specification and SSE is replaced

https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#streamable-http

pubmodmatt:
* Use the latest MCP Rust SDK from GitHub, since there have been no recent releases to crates.io
* Update the MCP Server to work with the latest SDK changes
* Add the Streamable HTTP transport via new `--http-port` and `--http-address` options
* Update docs to use Streamable HTTP instead of the deprecated SSE transport

Closes #78


docs/source/guides/index.mdx
```diff
@@ -230,16 +230,16 @@ Starting MCP inspector...
 
 1. In a browser, go to the URL returned by Inspector, then click **Connect** and **List Tools**. You should see the tools for the operations you provided.
 
-### Debug over SSE transport
```
DaleSeo, line 239: Debugging over SSE transport will no longer work? I thought we would provide another debugging option.

pubmodmatt: It's deprecated since November. This PR removes it from the docs. Eventually we'll remove it from the code.

CHANGELOG.md
```diff
@@ -13,6 +13,11 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### 📚 Documentation
 -->
 
+## [Unreleased]
```
Jephuff: Should we call out deprecating `--sse-port` and `sse_address` here?

crates/apollo-mcp-server/src/main.rs
```diff
@@ -96,13 +96,26 @@ struct Args {
     /// The log level for the MCP Server
     #[arg(long = "log", short = 'l', global = true, default_value_t = Level::INFO)]
     log_level: Level,
+
+    /// The IP address to bind the Streamable HTTP server to (default: 127.0.0.1)
```
Jephuff: I see we can't use the clap defaults here, but maybe we should still match claps default formatting
```suggestion
    /// The IP address to bind the Streamable HTTP server to
    ///
    /// [default: 127.0.0.1]
```

crates/apollo-mcp-server/src/main.rs
```diff
@@ -96,13 +96,26 @@ struct Args {
...
+    #[arg(long)]
+    http_address: Option<IpAddr>,
+
+    /// Start the server using the Streamable HTTP transport on the given port (default: 5000)
+    #[arg(long)]
```
Jephuff: I wonder if we should also try to keep defaults in the args next to their defaults so it's less likely we forget to update one or the other when making changes
```suggestion
    #[arg(long, default_value_if("http_address", ArgPredicate::IsPresent, Some("5000")))]
```

crates/apollo-mcp-server/src/main.rs
```diff
@@ -96,13 +96,26 @@ struct Args {
     /// The log level for the MCP Server
     #[arg(long = "log", short = 'l', global = true, default_value_t = Level::INFO)]
     log_level: Level,
+
+    /// The IP address to bind the Streamable HTTP server to (default: 127.0.0.1)
+    #[arg(long)]
```
Jephuff: I think for these we can also use conflicts with to prevent http+sse at the same time.
```suggestion
    #[arg(long, conflicts_with_all(["sse_port", "sse_address"]))]
```

DaleSeo: Oh, this seems neat!

crates/apollo-mcp-server/src/main.rs
```diff
@@ -188,6 +201,7 @@ async fn main() -> anyhow::Result<()> {
         .await?)
 }
 
+#[allow(clippy::result_large_err)]
```
DaleSeo, line 212: Just curious why this attribute is needed here.

crates/apollo-mcp-server/src/main.rs
pubmodmatt: Sure, I'll update the existing ones as well.

crates/apollo-mcp-server/src/main.rs
pubmodmatt: The new `ServerInitializeError` added in the SDK pushed this over clippy's warning limit.

CHANGELOG.md
pubmodmatt: The transport itself is deprecated, but I don't think we should deprecate the use of it yet.

The [guidance here](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#backwards-compatibility) indicates that servers may want to support older clients through SSE. Following that guidance shouldn't require using something deprecated.

crates/apollo-mcp-server/src/main.rs
pubmodmatt: You'd still have to redundantly use the default to unwrap it, which seems worse because now the default is in two places.

CHANGELOG.md
Jephuff: sounds good!

crates/apollo-mcp-server/src/main.rs
Jephuff: yeah, that is unfortunate :( The second default wouldn't be used, but it would make the code confusing. Totally fair to not use this one 👍

DaleSeo (approved):
I'm so excited about this support!

PR #110, opened 2025-06-02
Extend RawOperation to include variables and headers

Jephuff:
Extending RawOperation to include headers/variables to support operation collection data. 

Headers will act as overrides for default headers passed in by CLI

For variables I think they should act as overrides and remove them from the inputs, if they are defaults I don't think we are added any value with them. 

Ways to define variables
### Inline
no changing the value, and multiple uses need to repeat the value
```
query QueryName { field1(v: "v") field2(v: "v") }
```

### passed in
shared value across multiple fields, must be passed in
```
query QueryName($v: ID!) { field1(v: $v) field2(v: $v) }
```

### passed in with default
shared value across multiple fields, can be omited and default value will be used
```
query QueryName($v: ID! = "v") { field1(v: $v) field2(v: $v) }
```

### passed in with RawOperationVariable
shared value across multiple fields, but cannot be passed in
```
query QueryName($v: ID!) { field1(v: $v) field2(v: $v) }
```
```
{ v: "value" }
```

### DIFF
--- a/crates/apollo-mcp-server/src/graphql.rs
+++ b/crates/apollo-mcp-server/src/graphql.rs
@@ -5,6 +5,7 @@ use apollo_compiler::response::serde_json_bytes::serde_json;
 use apollo_compiler::response::serde_json_bytes::serde_json::Value;
 use reqwest::header::{HeaderMap, HeaderValue};
 use rmcp::model::{CallToolResult, Content, ErrorCode};
+use std::borrow::Cow;
 
 pub struct Request<'a> {
     pub input: Value,
@@ -24,13 +25,13 @@ pub trait Executable {
     fn variables(&self, input: Value) -> Result<Value, McpError>;
 
     /// Get the headers to execute the operation with
-    fn headers(&self, default_headers: &HeaderMap<HeaderValue>) -> HeaderMap<HeaderValue>;
+    fn headers(&self, default_headers: Cow<HeaderMap<HeaderValue>>) -> HeaderMap<HeaderValue>;
 
     /// Execute as a GraphQL operation using the endpoint and headers
     async fn execute(&self, request: Request<'_>) -> Result<CallToolResult, McpError> {
         reqwest::Client::new()
             .post(request.endpoint)
-            .headers(self.headers(&request.headers))
+            .headers(self.headers(Cow::Borrowed(&request.headers)))
             .body(if let Some(id) = self.persisted_query_id() {
                 serde_json::json!({
                     "extensions": {
--- a/crates/apollo-mcp-server/src/graphql.rs
+++ b/crates/apollo-mcp-server/src/graphql.rs
@@ -5,7 +5,6 @@ use apollo_compiler::response::serde_json_bytes::serde_json;
 use apollo_compiler::response::serde_json_bytes::serde_json::Value;
 use reqwest::header::{HeaderMap, HeaderValue};
 use rmcp::model::{CallToolResult, Content, ErrorCode};
-use std::borrow::Cow;
 
 pub struct Request<'a> {
     pub input: Value,
@@ -25,13 +24,13 @@ pub trait Executable {
     fn variables(&self, input: Value) -> Result<Value, McpError>;
 
     /// Get the headers to execute the operation with
-    fn headers(&self, default_headers: Cow<HeaderMap<HeaderValue>>) -> HeaderMap<HeaderValue>;
+    fn headers(&self, default_headers: &HeaderMap<HeaderValue>) -> HeaderMap<HeaderValue>;
 
     /// Execute as a GraphQL operation using the endpoint and headers
     async fn execute(&self, request: Request<'_>) -> Result<CallToolResult, McpError> {
         reqwest::Client::new()
             .post(request.endpoint)
-            .headers(self.headers(Cow::Borrowed(&request.headers)))
+            .headers(self.headers(&request.headers))
             .body(if let Some(id) = self.persisted_query_id() {
                 serde_json::json!({
                     "extensions": {
--- a/crates/apollo-mcp-server/src/graphql.rs
+++ b/crates/apollo-mcp-server/src/graphql.rs
@@ -24,16 +24,13 @@ pub trait Executable {
     fn variables(&self, input: Value) -> Result<Value, McpError>;
 
     /// Get the headers to execute the operation with
-    fn headers(
-        &self,
-        default_headers: &HeaderMap<HeaderValue>,
-    ) -> Result<HeaderMap<HeaderValue>, McpError>;
+    fn headers(&self, default_headers: &HeaderMap<HeaderValue>) -> HeaderMap<HeaderValue>;
 
     /// Execute as a GraphQL operation using the endpoint and headers
     async fn execute(&self, request: Request<'_>) -> Result<CallToolResult, McpError> {
         reqwest::Client::new()
             .post(request.endpoint)
-            .headers(self.headers(&request.headers)?)
+            .headers(self.headers(&request.headers))
             .body(if let Some(id) = self.persisted_query_id() {
                 serde_json::json!({
                     "extensions": {
--- a/crates/apollo-mcp-server/src/graphql.rs
+++ b/crates/apollo-mcp-server/src/graphql.rs
@@ -23,7 +23,7 @@ pub trait Executable {
     /// Get the variables to execute the operation with
     fn variables(&self, input: Value) -> Result<Value, McpError>;
 
-    /// Get the headers to execute the operation with
+    /// Get the variables to execute the operation with
     fn headers(
         &self,
         default_headers: &HeaderMap<HeaderValue>,
--- a/crates/apollo-mcp-server/src/graphql.rs
+++ b/crates/apollo-mcp-server/src/graphql.rs
@@ -3,7 +3,7 @@
 use crate::errors::McpError;
 use apollo_compiler::response::serde_json_bytes::serde_json;
 use apollo_compiler::response::serde_json_bytes::serde_json::Value;
-use reqwest::header::HeaderMap;
+use reqwest::header::{HeaderMap, HeaderValue};
 use rmcp::model::{CallToolResult, Content, ErrorCode};
 
 pub struct Request<'a> {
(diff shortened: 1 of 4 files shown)


crates/apollo-mcp-server/src/graphql.rs
```diff
@@ -23,11 +23,17 @@ pub trait Executable {
     /// Get the variables to execute the operation with
     fn variables(&self, input: Value) -> Result<Value, McpError>;
 
+    /// Get the variables to execute the operation with
```
pubmodmatt:
```suggestion
    /// Get the headers to execute the operation with
```

crates/apollo-mcp-server/src/operations.rs
```diff
@@ -166,13 +168,19 @@ pub struct RawOperation {
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    variables: Option<HashMap<String, Value>>,
 }
 
 impl From<String> for RawOperation {
     fn from(source_text: String) -> Self {
         Self {
             source_text,
             persisted_query_id: None,
+            headers: None,
```
pubmodmatt, line 211: I'm not seeing how someone would create a `RawOperation` with headers and variables set?

crates/apollo-mcp-server/src/operations.rs
```diff
@@ -810,28 +823,78 @@ fn type_to_schema(
...
+                        None,
+                    )
+                })?;
+                let header_value = HeaderValue::from_str(value.as_str()).map_err(|e| {
+                    McpError::new(
+                        ErrorCode::INVALID_PARAMS,
+                        format!("Failed to parse header value: {e}"),
+                        None,
+                    )
+                })?;
```
pubmodmatt: We have the full `RawOperation` at the time we construct it, when operations are loaded. We should error out at that point, rather than waiting until the AI agent tries to execute the tool.

crates/apollo-mcp-server/src/operations.rs
```diff
@@ -810,28 +823,78 @@ fn type_to_schema(
 
 impl graphql::Executable for Operation {
     fn persisted_query_id(&self) -> Option<String> {
-        self.inner.persisted_query_id.clone()
+        // TODO: id was being overridden, should we be returning? Should this be behind a flag? self.inner.persisted_query_id.clone()
+        None
```
pubmodmatt, line 856: I'm not following this comment - why is this being removed? This will break the operation if the router only allows pesisted queries.

crates/apollo-mcp-server/src/operations.rs
```diff
@@ -810,28 +823,78 @@ fn type_to_schema(
...
+                })?;
+                let header_value = HeaderValue::from_str(value.as_str()).map_err(|e| {
+                    McpError::new(
+                        ErrorCode::INVALID_PARAMS,
+                        format!("Failed to parse header value: {e}"),
+                        None,
+                    )
+                })?;
+
+                headers.insert(header_name, header_value);
```
pubmodmatt: May want to add more debug logging around headers and when they are overridden (and variables).

crates/apollo-mcp-server/src/operations.rs
```diff
@@ -810,28 +823,78 @@ fn type_to_schema(
...
+                Ok(())
+            })?;
+            Ok(headers)
+        } else {
+            Ok(default_headers.clone())
```
pubmodmatt: You could likely avoid some of these clones by using `Cow`

crates/apollo-mcp-server/src/operations.rs
```diff
@@ -810,28 +823,78 @@ fn type_to_schema(
...
+                Value::Object(obj) => Ok(obj.clone()),
+                _ => Err(McpError::new(
+                    ErrorCode::INVALID_PARAMS,
+                    "Invalid input".to_string(),
+                    None,
+                )),
+            }?;
+
+            raw_variables.iter().for_each(|(key, value)| {
+                variables.insert(key.clone(), value.clone());
```
pubmodmatt: It seems very strange to me that we'd ignore an input from the LLM. This will likely lead to the LLM becoming confused about why its input didn't have the intended effect. Rather, the variable should be removed from the input schema altogether, so it doesn't know this exists.

pubmodmatt (changes requested):

crates/apollo-mcp-server/src/operations.rs
pubmodmatt: Actually, it looks like it is behaving that way in `get_json_schema`. In that case, if there is a conflict here, it may be better to return an `INVALID_PARAMS`, since this isn't a valid input for the tool.

crates/apollo-mcp-server/src/operations.rs
Jephuff: You can't yet, the first way will be when I get operation collections setup. This PR is just updating to make the common target.

crates/apollo-mcp-server/src/operations.rs
Jephuff: great callout, I'll update!

crates/apollo-mcp-server/src/operations.rs
Jephuff: The code I'm replacing always passed persistedQueries in as `None` so this maintains the existing behaviour. I wasn't sure if that was done for a reason and I also wanted to keep this PR a noop. 

So if that wasn't done for a reason, I can do a follow up to change the behaviour and pass the persisted query id to the server 👍

crates/apollo-mcp-server/src/operations.rs
Jephuff: yep, they are removed from the schema 👍 
I can make this an INVALID_PARAMS if we hit a case where the LLM passes extra data 👍

crates/apollo-mcp-server/src/operations.rs
pubmodmatt: Still not getting it, I may be missing something... The intention is that when operations are loaded from a PQ manifest, it sets the [persisted_query_id here](https://github.com/apollographql/apollo-mcp-server/blob/a8b898b8519a195b81e02a27109bf2d8c031abfc/crates/apollo-mcp-server/src/operations.rs#L191). It then passes that ID in when [calling the operation here](https://github.com/apollographql/apollo-mcp-server/blob/7d629993cebd5e0c089737e7cbf421b5916cb069/crates/apollo-mcp-server/src/graphql.rs#L37). This method needs to return that persisted_query_id.

Jephuff: The raw operation had the persisted query id, but the call to  from_document never passed it. https://github.com/apollographql/apollo-mcp-server/blob/main/crates/apollo-mcp-server/src/operations.rs#L201 it was hard coded to None. That hardcoded None is then used to make the inner RawOperation https://github.com/apollographql/apollo-mcp-server/blob/main/crates/apollo-mcp-server/src/operations.rs#L358 which was separate then the source RawOperation.

crates/apollo-mcp-server/src/operations.rs
Jephuff: Finally read the docs for Cow and now I understand why I see if everywhere 😂 
Done!

Jephuff: Ok, I had it all broken so it hid that I did Cow wrong 😂 Undid it but will revisit :P

crates/apollo-mcp-server/src/operations.rs
pubmodmatt: That appears to have been just a bug. It should have been `self.persisted_query_id` not `None`

Jephuff: Yep, sounds like it. I'd still rather change that behaviour in a separate PR, so I can do a follow up PR that changes that behaviour and adds the change log. 

Fixing this will break some workflows with Persisted queries if people were to have a contract variant with a persisted query list and not a separate router instance, that won't work anymore.

pubmodmatt: I'm fine with a follow-up PR, since this won't change the existing (broken) behavior.

crates/apollo-mcp-server/src/operations.rs
```diff
@@ -161,18 +163,53 @@ pub enum MutationMode {
...
 pub struct RawOperation {
     source_text: String,
-    #[serde(skip_serializing_if = "Option::is_none")]
     persisted_query_id: Option<String>,
+    headers: Option<HeaderMap<HeaderValue>>,
+    variables: Option<HashMap<String, Value>>,
+}
+
+// Custom Serialize implementation for RawOperation
+impl serde::Serialize for RawOperation {
```
pubmodmatt, line 175: Why is the custom implementation needed? Can you add that to the comment?

crates/apollo-mcp-server/src/operations.rs
```diff
@@ -810,28 +852,76 @@ fn type_to_schema(
...
+                    "Invalid input".to_string(),
+                    None,
+                )),
+            }?;
+
+            raw_variables.iter().try_for_each(|(key, value)| {
+                if variables.contains_key(key) {
+                    Err(McpError::new(
+                        ErrorCode::INVALID_PARAMS,
+                        "Variable {key} conflict, only variables in input schema should be passed",
```
pubmodmatt: Nit - from the MCP client perspective, variables aren't a thing. The MCP documentation calls the tool input "parameters".
```suggestion
                        "No such parameter: {key}",
```

crates/apollo-mcp-server/src/operations.rs
pubmodmatt: Actually, it looks like `Cow` didn't help at all - it's still cloning in every case. May not be worth the extra complexity then.

crates/apollo-mcp-server/src/operations.rs
Jephuff: HeaderMap/HeaderValue/HeaderName don't derive Serialize so it yells at me. 
I'll add a comment 👍

crates/apollo-mcp-server/src/operations.rs
```diff
@@ -161,18 +163,53 @@ pub enum MutationMode {
...
+#[derive(Debug, Clone)]
 pub struct RawOperation {
     source_text: String,
-    #[serde(skip_serializing_if = "Option::is_none")]
     persisted_query_id: Option<String>,
+    headers: Option<HeaderMap<HeaderValue>>,
+    variables: Option<HashMap<String, Value>>,
+}
+
+// Custom Serialize implementation for RawOperation
```
Jephuff, line 173:
```suggestion
// Custom Serialize implementation for RawOperation
// This is needed because reqwests HeaderMap/HeaderValue/HeaderName don't derive Serialize
```

crates/apollo-mcp-server/src/operations.rs
```diff
@@ -161,18 +163,54 @@ pub enum MutationMode {
...
+        state.serialize_field("source_text", &self.source_text)?;
+        if let Some(ref id) = self.persisted_query_id {
+            state.serialize_field("persisted_query_id", id)?;
+        }
+        if let Some(ref variables) = self.variables {
+            state.serialize_field("variables", variables)?;
+        }
+        if let Some(ref headers) = self.headers {
+            state.serialize_field(
+                "variables",
```
pubmodmatt:
```suggestion
                "headers",
```

crates/apollo-mcp-server/src/operations.rs
Jephuff: alright, I can remove it!

PR #115, opened 2025-06-03
Update to latest MCP SDK and fix nix develop

pubmodmatt:
Switching to depend on the MCP Rust SDK through git had broken `nix develop`:

* Add git package to nix
* Specify the specific `main` branch in the dependency to work around a bug where crane assumes the branch is `master`
* Ignore the result symlink created by nix

These changes caused the lock file to be updated, which pulled in the latest changes to the MCP Rust SDK. Specifically, this alters the way Streamable HTTP servers are started. Uptake those changes.


flake.nix
```diff
@@ -63,6 +63,7 @@
         nativeBuildInputs = apollo-mcp-builder.nativeDependencies;
         buildInputs =
           [
+            pkgs.git
```
nicholascioli: Would you mind moving this to the block below (starting with `with pkgs` and add a comment why it is needed?

PR #117, opened 2025-06-04
refactor: separate out states

nicholascioli:
This PR separates out all of the available state machine states into their own files to make it easier to navigate and manage.


crates/apollo-mcp-server/src/states/configuring.rs
```diff
@@ -0,0 +1,70 @@
...
+    custom_scalar_map::CustomScalarMap,
+    errors::ServerError,
+    operations::{MutationMode, RawOperation},
+    server::Transport,
+    states::operations_configured::OperationsConfigured,
+};
+
+use super::schema_configured::SchemaConfigured;
+
+pub(crate) struct Configuring {
```
pubmodmatt: Can you change all the `pub(crate)` in all submodules of `states` to be `pub(super)` instead? These are implementation details of the `server` module, and shouldn't be expoed to the whole crate.

crates/apollo-mcp-server/src/states.rs
```diff
@@ -0,0 +1,236 @@
...
+use starting::Starting;
+
+mod configuring;
+mod operations_configured;
+mod running;
+mod schema_configured;
+mod starting;
+
+#[allow(clippy::large_enum_variant)]
+enum State {
```
pubmodmatt: I'm not sure I like separating this out from the `server` module - a `State` doesn't stand on its own, it's only usable in the context of a server. This move causes it to run afowl of the convention to avoid [repeating the module name in the type name](https://rust-lang.github.io/rust-clippy/master/index.html#module_name_repetitions), since this becomes `states::State`. `server::State` is much better, since that's exactly what it is. Some type-state implementations go so far as to make the state a generic parameter on the type, like `Server<State>`, and transition the `Server` itself between states. This implementation isn't as tightly coupled, but a `State` is still a server state.

This also ties in with my other comment about `pub(super)`. All of this is implementation detail of the `Server`, so these types should all only be accessible from the `server` module.

crates/apollo-mcp-server/src/states/configuring.rs
nicholascioli: Done!

crates/apollo-mcp-server/src/states.rs
nicholascioli: I ended up hiding this from the server entirely since it is only used by the state machine.

pubmodmatt (approved):
Sorry this took so long - this looks fine

nicholascioli:
I've rebased this on the latest changes.

PR #118, opened 2025-06-04
Add support for Operation Collections

Jephuff:
Adds operation collection as an OperationSource. It's currently setup so that it will poll for changes every 30s and will only fetch the fully data for operations that change.

A few TODOs which I can either do in this PR or as fast follows
- [x] The first request should be the full operation collection data to avoid running the poll request and then fetching the data
- [ ] Explorer supports jsonc, should we consider supporting that here as well?

### DIFF
--- a/crates/apollo-mcp-registry/src/platform_api/operation_collections/collection_poller.rs
+++ b/crates/apollo-mcp-registry/src/platform_api/operation_collections/collection_poller.rs
@@ -274,7 +274,7 @@ impl CollectionSource {
                         }
                     }
                     Ok(None) => {
-                        tracing::info!("Operation collection unchanged");
+                        tracing::debug!("Operation collection unchanged");
                     }
                     Err(err) => {
                         if let Err(e) = sender.send(CollectionEvent::CollectionError(err)).await {


crates/apollo-mcp-server/src/operation_collection.rs
```diff
@@ -0,0 +1,296 @@
...
+use reqwest::header::{HeaderMap, HeaderName, HeaderValue};
+use rmcp::serde_json::{self, Value};
+use std::collections::HashMap;
+use std::pin::Pin;
+use std::str::FromStr;
+use std::time::Duration;
+use tokio::sync::mpsc::channel;
+use tokio_stream::wrappers::ReceiverStream;
+
+const STUDIO_API: &str = "https://graphql.api.apollographql.com/api/graphql";
```
DaleSeo: Since we call it platform api in the rest of the file?

```suggestion
const PLATFORM_API: &str = "https://graphql.api.apollographql.com/api/graphql";
```

crates/apollo-mcp-server/src/operation_collection.rs
```diff
@@ -0,0 +1,296 @@
...
+}
+
+#[derive(Clone)]
+pub struct CollectionSource {
+    pub collection_id: String,
+    pub platform_api_config: PlatformApiConfig,
+}
+
+impl CollectionSource {
+    pub fn into_stream(self) -> Pin<Box<dyn Stream<Item = CollectionEvent> + Send>> {
```
DaleSeo: This return type looks Greek to me 😂

crates/apollo-mcp-server/src/operation_collection.rs
```diff
@@ -0,0 +1,296 @@
...
+        };
+
+        tokio::task::spawn(task);
+
+        Box::pin(ReceiverStream::new(receiver))
+    }
+}
+
+pub enum CollectionEvent {
+    OperationCollectionUpdate(Vec<RawOperation>),
```
DaleSeo: I suppose this sounds slightly more common for an event name?

```suggestion
    OperationCollectionUpdated(Vec<RawOperation>),
```

crates/apollo-mcp-server/src/operation_collection.rs
```diff
@@ -0,0 +1,296 @@
...
+pub struct CollectionSource {
+    pub collection_id: String,
+    pub platform_api_config: PlatformApiConfig,
+}
+
+impl CollectionSource {
+    pub fn into_stream(self) -> Pin<Box<dyn Stream<Item = CollectionEvent> + Send>> {
+        let (sender, receiver) = channel(2);
+        let collection_id = self.collection_id.clone();
+        let platform_api_config = self.platform_api_config.clone();
```
DaleSeo: Why should we clone these? To be thread-safe?

crates/apollo-mcp-server/src/operation_collection.rs
```diff
@@ -0,0 +1,296 @@
...
+use tokio::sync::mpsc::channel;
+use tokio_stream::wrappers::ReceiverStream;
+
+const STUDIO_API: &str = "https://graphql.api.apollographql.com/api/graphql";
+
+/// Configuration for polling Apollo Platform API.
+#[derive(Clone, Debug, Default)]
+pub struct PlatformApiConfig {
+    /// The Apollo key: `<YOUR_GRAPH_API_KEY>`
+    pub apollo_key: String,
```
DaleSeo: Are operation collections graph variant-agnostic?

crates/apollo-mcp-server/src/operation_collection.rs
```diff
@@ -0,0 +1,296 @@
...
+type Timestamp = String;
+
+#[derive(GraphQLQuery)]
+#[graphql(
+    query_path = "src/operation_collection_query.graphql",
+    schema_path = "src/platform-api.graphql",
+    request_derives = "Debug",
+    response_derives = "PartialEq, Debug, Deserialize, Clone"
+)]
+struct OperationCollectionQuery;
```
DaleSeo: I was a bit confused about this query with the other one. Maybe this makes it clearer?

```suggestion
struct OperationCollectionEntryQuery;
```

crates/apollo-mcp-server/src/operation_collection.rs
```diff
@@ -0,0 +1,296 @@
+use crate::errors::CollectionError;
+use crate::operations::RawOperation;
+use futures::Stream;
+use graphql_client::{GraphQLQuery, Response};
+use operation_collection_polling_query::{
+    OperationCollectionPollingQueryOperationCollection as OPPollingQueryOP, // this is to long and break formatting.
+    OperationCollectionPollingQueryOperationCollectionOnOperationCollection,
```
DaleSeo: Where do these guys come from? Are they generated by `graphql_client` somehow?

crates/apollo-mcp-server/src/operation_collection.rs
```diff
@@ -0,0 +1,296 @@
...
+                        },
+                    );
+                    Some(operation.id.clone())
+                }
+            } else {
+                previous_updated_at.insert(
+                    operation.id.clone(),
+                    CollectionData {
+                        last_updated_at: updated_at,
+                        raw_operation: None,
```
DaleSeo: Why are we setting this to `None`? Shouldn't we set it to the updated operation?

crates/apollo-mcp-server/src/operation_collection.rs
```diff
@@ -0,0 +1,296 @@
...
+#[derive(GraphQLQuery)]
+#[graphql(
+    query_path = "src/operation_collection_polling_query.graphql",
+    schema_path = "src/platform-api.graphql",
+    request_derives = "Debug",
+    response_derives = "PartialEq, Debug, Deserialize"
+)]
+struct OperationCollectionPollingQuery;
+
+fn changed_ids(
```
DaleSeo: I think this function name can be a bit misleading. From what I see, it does two things: it returns changed IDs and it mutates `previous_updated_at`.

crates/apollo-mcp-server/src/operation_collection.rs
```diff
@@ -0,0 +1,296 @@
...
+                Some(operation.id.clone())
+            }
+        })
+        .collect()
+}
+
+#[derive(Clone)]
+pub struct CollectionData {
+    last_updated_at: String,
+    raw_operation: Option<RawOperation>,
```
DaleSeo: Would it be okay to use `RawOperation` here? It seems like it was designed for persisted queries?

crates/apollo-mcp-server/src/operation_collection.rs
pubmodmatt, line 1: Since this has to do with operations, I'd make all of these new `operation_*` files sub-modules of the existing `operations` module, so maybe `operations::collection` (the `operations` module maybe should be singular `operation` instead). 

Arguably, I'd consider this new code to really be part of the "registry". The `apollo-mcp-registry` crate is meant to contain anything where we're getting artifacts from a registry like Uplink. Getting the operation collection from the Platform API qualifies, in my view - we're using GraphOS like a registry of operations (and eventually, this may evolve into an actual operation registry). So I'd suggest moving all of this into that crate.

crates/apollo-mcp-server/src/operation_collection.rs
```diff
@@ -0,0 +1,296 @@
...
+use std::str::FromStr;
+use std::time::Duration;
+use tokio::sync::mpsc::channel;
+use tokio_stream::wrappers::ReceiverStream;
+
+const STUDIO_API: &str = "https://graphql.api.apollographql.com/api/graphql";
+
+/// Configuration for polling Apollo Platform API.
+#[derive(Clone, Debug, Default)]
+pub struct PlatformApiConfig {
```
pubmodmatt: Similar to my earlier comment - this belongs alongside `UplinkConfig` in `apollo-mcp-registry`.

pubmodmatt (commented):
I'll have more commets, but I'm suggesting some things that have a wide impact so getting these out first

crates/apollo-mcp-server/src/platform-api.graphql
pubmodmatt, line 1: Definitely don't think we should have this giant platform API schema as a top level source in the MCP server. This is another thing that should move to `apollo-mcp-registry`.

crates/apollo-mcp-server/src/main.rs
```diff
@@ -164,6 +169,11 @@ async fn main() -> anyhow::Result<()> {
         OperationSource::from(ManifestSource::LocalHotReload(vec![manifest]))
     } else if !args.operations.is_empty() {
         OperationSource::from(args.operations)
+    } else if let Some(collection_id) = args.collection {
+        OperationSource::from(CollectionSource {
+            collection_id,
+            platform_api_config: platform_api_config()?,
+        })
```
DaleSeo: Can't we just this without implementing the `From` trait? This should be a dumb question... 😂

```suggestion
        OperationSource::Collection(CollectionSource {
            collection_id,
            platform_api_config: platform_api_config()?,
        })
```

crates/apollo-mcp-server/src/event.rs
```diff
@@ -15,6 +16,9 @@ pub enum Event {
     /// An error occurred when loading operations
     OperationError(io::Error),
 
+    /// An error occurred when loading operations
```
DaleSeo: This comment is the same as the one above.

```suggestion
    /// An error occurred when loading operations from collection
```

crates/apollo-mcp-server/src/platform-api.graphql
DaleSeo, line 1: Do we need the entire schema to execute just two queries? 😮

DaleSeo (approved):
Sorry @Jephuff for showering your PR with newbie comments! I was just using it as my Rust crash course. 🙈 Learned heaps, though, so thanks for putting up with my noise!

Now that I’ve wrapped my head around the code, I'll spend some time tomorrow testing it with a graph on Studio.

crates/apollo-mcp-server/src/platform-api.graphql
pubmodmatt: Also, [Rover already has this](https://github.com/apollographql/rover/blob/4550a8526ffa42452e5c17fa251886ee99d140f6/crates/rover-client/.schema/schema.graphql). Ideally, that `rover-client` should move out to a shared repo - I'd like to make this into a "Rust SDK for Apollo Platform API". The rover code has a way to pick up changes to the schema using a hash.

crates/apollo-mcp-server/src/operation_collection.rs
Jephuff: moving to the other crate feels like it makes a lot of sense to me! Let the refactor begin 🐎

crates/apollo-mcp-server/src/platform-api.graphql
Jephuff: Yeah, splitting off our own sdk crate would be awesome, I wasn't going to bother with adding pulling updated schemas yet because our use is so small. Sounds like we can just leave it off and when we split out the rover code we will get it?

Jephuff:
> Sorry @Jephuff for showering your PR with newbie comments! I was just using it as my Rust crash course. 🙈 Learned heaps, though, so thanks for putting up with my noise!
> 
> Now that I’ve wrapped my head around the code, I'll spend some time tomorrow testing it with a graph on Studio.

Love the comments! keep the noise coming, we will learn together :D hahaha

crates/apollo-mcp-server/src/operation_collection.rs
pubmodmatt: `RawOperation` is just the raw operation from the operation source, no matter what that source is (PQ, operation collection, graphql document file, etc.), before processing it as an MCP Tool. It's not specific to PQ. The persisted query ID on that type is optional, in case the source is a PQ.

crates/apollo-mcp-server/src/operation_collection.rs
Jephuff: They aren't, but the graph ref isn't required to fetch them. We could add a warning if a graph ref is passed in that doesn't match the graph ref of the collection, but might not be worth the noise.

crates/apollo-mcp-server/src/operation_collection.rs
Jephuff: yep, they are generated by graphql_client.
This stuff generates it 
```
#[derive(GraphQLQuery)]
#[graphql(
    query_path = "src/operation_collection_polling_query.graphql",
    schema_path = "src/platform-api.graphql",
    request_derives = "Debug",
    response_derives = "PartialEq, Debug, Deserialize"
)]
struct OperationCollectionPollingQuery;
```

crates/apollo-mcp-server/src/operation_collection.rs
Jephuff: We don't have the operation data yet at this point because we've only run the poll query. 

But thinking about it, writing to this cache would only really help if we have overlapping requests and we probably should avoid that. So maybe we could make raw operations required and write to the cache once we get the full body back 🤔

crates/apollo-mcp-server/src/platform-api.graphql
pubmodmatt: I think that's the right approach 👍

crates/apollo-mcp-server/src/operation_collection.rs
DaleSeo: Thanks for the clarification, @pubmodmatt. Makes sense!

DaleSeo:
@Jephuff I had no luck getting this work locally 😞 The server crashes with `Error in response: Not Authorized` when I use `---collection` option though `APOLLO_KEY` is set. Am I missing anything?

```sh
$ APOLLO_GRAPH_REF=🔑 APOLLO_KEY=🔑 cargo run -p apollo-mcp-server -- -i -u --http-address 127.0.0.1 --http-port 5000 --collection 4efe3d61-a3cf-4906-bd6d-6cdf6c18239b -l debug 
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.16s
     Running `target/debug/apollo-mcp-server -i -u --http-address 127.0.0.1 --http-port 5000 -l debug --collection 4efe3d61-a3cf-4906-bd6d-6cdf6c18239b`
2025-06-06T16:24:08.383024Z  INFO Apollo MCP Server v0.3.0 // (c) Apollo Graph, Inc. // Licensed as ELv2 (https://go.apollo.dev/elv2)
2025-06-06T16:24:08.386057Z DEBUG starting new connection: https://uplink.api.apollographql.com/    
2025-06-06T16:24:08.386799Z DEBUG starting new connection: https://graphql.api.apollographql.com/    
2025-06-06T16:24:08.485455Z DEBUG connecting to 34.30.93.119:443
2025-06-06T16:24:08.485628Z DEBUG connecting to 34.117.186.194:443
2025-06-06T16:24:08.534777Z DEBUG connected to 34.117.186.194:443
2025-06-06T16:24:08.558533Z DEBUG connected to 34.30.93.119:443
2025-06-06T16:24:08.809370Z DEBUG uplink response Response { url: "https://uplink.api.apollographql.com/", status: 200, headers: {"content-type": "application/json;charset=utf-8", "x-cloud-trace-context": "27bd18f8775451310924e30b1eb5b6c4", "date": "Fri, 06 Jun 2025 16:24:08 GMT", "server": "Google Frontend", "content-length": "4338", "via": "1.1 google", "alt-svc": "h3=\":443\"; ma=2592000,h3-29=\":443\"; ma=2592000"} }
2025-06-06T16:24:08.814416Z DEBUG pooling idle connection for ("https", uplink.api.apollographql.com)
2025-06-06T16:24:08.833726Z DEBUG Received schema: ✂️(The schema is too long to paste here)✂️

2025-06-06T16:24:08.868704Z DEBUG pooling idle connection for ("https", graphql.api.apollographql.com)
Error: Failed to create operation: Error loading collection: Error in response: Not Authorized

Caused by:
    Error loading collection: Error in response: Not Authorized
```

Jephuff:
@DaleSeo  How did you get the APOLLO_KEY? Is it a graph key?

DaleSeo:
@Jephuff Yes, it's a graph key from a graph on my free account. The log indicates that it fetched the schema without any issues using the key.

Shot 2025-06-06 at 19 08 46

Jephuff:
It looks like you grabbed a personal collection id instead of a shared one. Personal ones would require a user token, but the shared ones would be able to be fetched by the graph token! :)

DaleSeo:
> It looks like you grabbed a personal collection id instead of a shared one. Personal ones would require a user token, but the shared ones would be able to be fetched by the graph token! :)

Thanks @Jephuff for clarification. I never thought using a personal collection versus a shared collection could make a difference.

I tried using a user key instead, but it fails to fetch the schema from Uplink. 🙃 
It seems like I can't use a personal collection unless we introduce a new environment variable. 😞

```sh
2025-06-09T13:02:20.456238Z ERROR uplink error, the request will not be retried: code=AUTHENTICATION_FAILED message=Only graph tokens can be used with uplink. User tokens are not supported at this time.
Error: No valid schema was supplied
```

DaleSeo:
I got it working using a shared collection and a graph key! 🎉

Shot 2025-06-09 at 09 21 37

crates/apollo-mcp-server/src/operation_collection.rs
Jephuff: Changed to UpdateOperationCollection to match the manifest event

crates/apollo-mcp-server/src/operation_collection.rs
Jephuff: Doesn't look like I need to clone them! 🎉

crates/apollo-mcp-server/src/operation_collection.rs
Jephuff: done!

crates/apollo-mcp-server/src/platform-api.graphql
Jephuff: the schema isn't needed for executing the schema, it's used for the codegen step, so we need the full schema because we don't know what queries we will right

crates/apollo-mcp-server/Cargo.toml
```diff
@@ -28,6 +28,8 @@ tokio.workspace = true
 tracing.workspace = true
 tracing-subscriber = { version = "0.3.19", features = ["env-filter"] }
 tokio-util = "0.7.15"
+graphql_client.workspace = true
+tokio-stream.workspace = true
```
pubmodmatt: Not needed anymore with the refactor to apollo-mcp-registry?

crates/apollo-mcp-server/src/operations.rs
```diff
@@ -164,10 +195,10 @@ pub enum MutationMode {
...
 #[derive(Debug, Clone)]
 pub struct RawOperation {
-    source_text: String,
-    persisted_query_id: Option<String>,
-    headers: Option<HeaderMap<HeaderValue>>,
-    variables: Option<HashMap<String, Value>>,
+    pub source_text: String,
+    pub persisted_query_id: Option<String>,
+    pub headers: Option<HeaderMap<HeaderValue>>,
+    pub variables: Option<HashMap<String, Value>>,
```
pubmodmatt: Why make these `pub`? It looks like there's a `From` implementation below that will create the new case.

crates/apollo-mcp-registry/src/platform_api.rs
```diff
@@ -0,0 +1,30 @@
...
+    pub poll_interval: Duration,
+
+    /// The HTTP client timeout for each poll
+    pub timeout: Duration,
+}
+
+impl PlatformApiConfig {
+    /// Mock uplink configuration options for use in tests
+    /// A nice pattern is to use wiremock to start an uplink mocker and pass the URL here.
+    pub fn for_tests() -> Self {
```
pubmodmatt: Can this be behind `#[cfg(test)]`?

crates/apollo-mcp-registry/src/platform_api/operation_collections/operation_collection_polling_query.graphql
```diff
@@ -0,0 +1,20 @@
+query OperationCollectionPollingQuery($operationCollectionId: ID!) {
+    operationCollection(id: $operationCollectionId) {
+        __typename
+        ... on OperationCollection {
+            operations {
+                lastUpdatedAt
+                id
+            }
```
pubmodmatt: My understanding here is that this will always return an entry for *every* operation in the collection, even if it hasn't changed? So if there are 1 million operations in the collection, every poll would return these values for all of them?

crates/apollo-mcp-registry/src/platform_api/operation_collections/collection_poller.rs
```diff
@@ -0,0 +1,272 @@
...
+pub struct CollectionCache {
+    last_updated_at: String,
+    operation_data: Option<OperationData>,
+}
+pub async fn fetch_operation_collection(
+    collection_entry_ids: Vec<String>,
+    platform_api_config: &PlatformApiConfig,
+) -> Result<Response<operation_collection_entries_query::ResponseData>, CollectionError> {
+    let key_header_value = HeaderValue::from_str(platform_api_config.apollo_key.expose_secret())
+        .map_err(CollectionError::HeaderValue)?;
```
pubmodmatt: Nitpick - I generally prefer not using single use local variable bindings. As in, just put this directlty in the code below. This can be more than just stylistic in cases like this - here you're exposing a secret, and storing it in this variable can lead to it being exposed unnecessarily (for example if there was an error between here and the point where it was passed in) or longer than necessary (until this variable goes out of scope). Though in this case, it doesn't look like either of those is an issue, so it's more just style and practice.

crates/apollo-mcp-registry/src/platform_api/operation_collections/collection_poller.rs
```diff
@@ -0,0 +1,272 @@
...
+
+    let response = reqwest::Client::new()
+        .post(PLATFORM_API)
+        .headers(HeaderMap::from_iter(vec![
+            (
+                HeaderName::from_static("apollographql-client-name"),
+                HeaderValue::from_static("apollo-mcp-server"),
+            ),
+            // TODO: add apollographql-client-version header
+            (HeaderName::from_static("x-api-key"), key_header_value),
```
pubmodmatt: There's a lot of duplication of this client code between here and `fetch_operation_collection`, can you DRY this up?

crates/apollo-mcp-registry/src/platform_api/operation_collections/collection_poller.rs
```diff
@@ -0,0 +1,272 @@
...
+        .map_err(CollectionError::Request)?
+        .data
+        .ok_or(CollectionError::Response("missing data".to_string()))?;
+
+    match response.operation_collection {
+        OperationCollectionPollingQueryOperationCollection::OperationCollection(collection) => {
+            let changed_ids = changed_ids(previous_updated_at, collection);
+
+            if changed_ids.is_empty() {
+                tracing::info!("no operation changed");
```
pubmodmatt: Info seems excessive here.

crates/apollo-mcp-registry/src/platform_api/operation_collections/collection_poller.rs
```diff
@@ -0,0 +1,272 @@
...
+        ]))
+        .timeout(platform_api_config.timeout)
+        .json(&OperationCollectionPollingQuery::build_query(
+            operation_collection_polling_query::Variables {
+                operation_collection_id: collection_id.clone(),
+            },
+        ))
+        .send()
+        .await
+        .map_err(CollectionError::Request)?
```
pubmodmatt: Can we do any special handling for rate limiting errors here, like back off the duration of the polling?

crates/apollo-mcp-registry/src/platform_api/operation_collections/collection_poller.rs
```diff
@@ -0,0 +1,272 @@
...
+        .post(PLATFORM_API)
+        .headers(HeaderMap::from_iter(vec![
+            (
+                HeaderName::from_static("apollographql-client-name"),
+                HeaderValue::from_static("apollo-mcp-server"),
+            ),
+            // TODO: add apollographql-client-version header
+            (HeaderName::from_static("x-api-key"), key_header_value),
+        ]))
+        .timeout(platform_api_config.timeout)
```
pubmodmatt, line 315: IIRC, by default reqwest doesn't do any retries - can you confirm? I wouldn't want to in this case, since we're polling on a durtaion - if a request fails the retry will be the next poll.

crates/apollo-mcp-registry/src/platform_api/operation_collections/collection_poller.rs
```diff
@@ -0,0 +1,272 @@
...
+            }
+        };
+
+        tokio::task::spawn(task);
+
+        Box::pin(ReceiverStream::new(receiver))
+    }
+}
+
+async fn poll_operation_collection(
```
pubmodmatt: This poll operation seems fairly heavy-weight. It looks like it will require:
1. TLS termination
2. Validation of the API key
3. Presumably a data store lookup of the operation collection ID

Do we have any average latency numbers on this query? I'm a bit concerned that every deployed MCP server will be hitting this every 30 seconds. Initially the number of requests might be small, but over time this could put a fairly high burden on the platform API.

crates/apollo-mcp-server/src/operations.rs
Jephuff: Oohh I think this is left over from some code I deleted, I should be able to undo!

crates/apollo-mcp-registry/src/platform_api.rs
Jephuff: Honestly it's not used right now, I'm just going to delete it for now.

crates/apollo-mcp-registry/src/platform_api/operation_collections/operation_collection_polling_query.graphql
Jephuff: yep, thats exactly right 👍  
I think in hardening we would want to do something like adding a new field to track changes in ANY operation, or add another field that gives you chances since a time or something. But I think this is unfortunately the best we can do right now.

crates/apollo-mcp-registry/src/platform_api/operation_collections/collection_poller.rs
Jephuff: You definitely don't need to tell me to inline twice! 
Updated! :)

crates/apollo-mcp-registry/src/platform_api/operation_collections/collection_poller.rs
Jephuff: oops, yes I did that for testing purposes 👍 I'll make it debug

crates/apollo-mcp-registry/src/platform_api/operation_collections/collection_poller.rs
Jephuff: we probably should! Would you be ok with that being a follow up PR? Or would you rather it in this one?

pubmodmatt: Follow up is fine.

pubmodmatt:
Also this definitely needs a changeset entry and doc updates.

crates/apollo-mcp-registry/src/platform_api/operation_collections/collection_poller.rs
Jephuff: ticket created 👍

crates/apollo-mcp-registry/src/platform_api/operation_collections/operation_collection_polling_query.graphql
pubmodmatt: For the short-term, it may be worth capping the collection size at some reasonable number?

crates/apollo-mcp-registry/src/platform_api/operation_collections/collection_poller.rs
Jephuff: Ok! Dried it up!

crates/apollo-mcp-registry/src/platform_api/operation_collections/operation_collection_polling_query.graphql
Jephuff: We can't limit it in the query, but I can disable polling if we are over the threshold? 
What do we think is a reasonable number? I imagine people wouldn't be adding a huge number of tools, so maybe 100?

crates/apollo-mcp-registry/src/platform_api/operation_collections/collection_poller.rs
Jephuff: yep looks like any retries need to be manually configured 👍

crates/apollo-mcp-registry/src/platform_api/operation_collections/operation_collection_polling_query.graphql
pubmodmatt: This can be added separately. 100 or 1000 might be good places to start. I'd probably rather it just error out on startup, than have a different behavior mode.

Jephuff: hmm, if we only error on startup then someone could add any number of operations to the collection after startup and it wouldn't prevent the polling. Are we ok with that?

docs/source/command-reference.mdx
```diff
@@ -109,6 +109,7 @@ apollo-mcp-server [OPTIONS] --directory <DIRECTORY>
 | `-x, --explorer`                                      | Expose a tool that returns the URL to open a GraphQL operation in Apollo Explorer (requires `APOLLO_KEY` and `APOLLO_GRAPH_REF`).                                                                                                    |
 | `-o, --operations [<OPERATIONS>...]`                  | Operation files to expose as MCP tools. [Learn more](/apollo-mcp-server/guides/#from-operation-files).                                                                                                                               |
 | `--manifest <MANIFEST>`                               | The path to the persisted query manifest containing operations.                                                                                                                                                                      |
+| `--collection <COLLECTION_ID>`                        | The id of an operation collection to use as the source for operations                                                                                                                                                                |
```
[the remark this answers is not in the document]
BlenderDude: Ignore this for now 😄 working to resolve this issue with the prompt

.changesets/feat_jeffrey_poll_operation_collections.md
```diff
@@ -0,0 +1,3 @@
+### Add `--collection <COLLECTION_ID>` as another option for operation source
+
+Use operation collections as the source of operations for your MCP server. The server will watch for changes and automatically update when you change your operation collection.
```
pubmodmatt, line 3:
```suggestion
Use shared operation collections as the source of operations for your MCP server. The server will watch for changes and automatically update when you change your operation collection.
```

docs/source/command-reference.mdx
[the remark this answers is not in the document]
Jephuff: Any chance we can turn these off if they aren't quite ready? It's quite a lot and contradicting itself 😂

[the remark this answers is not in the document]
pubmodmatt: I actually agree with the ID capitalization.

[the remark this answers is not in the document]
Jephuff: yeah the ID I definitely agree with :)

.changesets/feat_jeffrey_poll_operation_collections.md
Jephuff: I'm not sure how to word this, you 100% can use a sandbox or personal collection as long as your APOLLO_KEY has access to it. If you do, it is incompatible with uplink but you can still pass in a schema.

pubmodmatt: I think we need a larger doc section, beyond just the options in the command reference, that explains this. This will be an immediate area of confusion. But that can be in a follow-on doc PR.

crates/apollo-mcp-registry/src/platform_api/operation_collections/operation_collection_polling_query.graphql
pubmodmatt: We can work this out after this PR, but I think it's probably both. It may depend on how we integrate this with GraphOS generally - if it's a special MCP collection, maybe that collection simply can't be updated past the limit, for example.

crates/apollo-mcp-registry/src/platform_api/operation_collections/collection_poller.rs
pubmodmatt: This was discussed separately - it sounds like we have some ideas here, but I'd like to see some follow up after this merges.

PR #119, opened 2025-06-05
fix: Indicate filename for operations files that error while loading

swcollard:
When we hit an error while loading operation files we are not currently including which file contained the error in the messaging. This change propagates the path alongside the operation so that if we hit one of the several errors in the operation parsing we can display which file it is associated with.

Updated unit tests and validated with a few examples

Multiple operations in single file error
```sh
➜  cargo build && ./target/debug/apollo-mcp-server \
     --directory ./graphql/TheSpaceDevs \
     --schema api.graphql \
     --operations operations/ExploreCelestialBodies.graphql operations/ExploreCelestialBodies2.graphql \
     --endpoint https://thespacedevs-production.up.railway.app/ \
     --http-port 5000 \
     --http-address 127.0.0.1

Error: Failed to create operation: operations/ExploreCelestialBodies2.graphql: Too many operations. Expected 1 but got 2

Caused by:
    operations/ExploreCelestialBodies2.graphql: Too many operations. Expected 1 but got 2
```

No operation name error
```sh
➜  cargo build && ./target/debug/apollo-mcp-server \
     --directory ./graphql/TheSpaceDevs \
     --schema api.graphql \
     --operations operations/ExploreCelestialBodies.graphql operations/ExploreCelestialBodies3.graphql \
     --endpoint https://thespacedevs-production.up.railway.app/ \
     --http-port 5000 \
     --http-address 127.0.0.1

2025-06-05T14:26:24.697796Z  INFO Apollo MCP Server v0.3.0 // (c) Apollo Graph, Inc. // Licensed as ELv2 (https://go.apollo.dev/elv2)
2025-06-05T14:26:24.721298Z  INFO Tool ExploreCelestialBodies loaded with a character count of 759. Estimated tokens: 189
Error: Failed to create operation: operations/ExploreCelestialBodies3.graphql: Operation is missing its required name: query($search: String, $limit: Int = 10, $offset: Int = 0) { bodies: celestialBodies(search: $search, limit: $limit, offset: $offset) { pageInfo { count next previous } results { id name diameter mass gravity lengthOfDay atmosphere type { id name } image { url thumbnail credit } description wikiUrl } } }

Caused by:
    operations/ExploreCelestialBodies3.graphql: Operation is missing its required name: query($search: String, $limit: Int = 10, $offset: Int = 0) { bodies: celestialBodies(search: $search, limit: $limit, offset: $offset) { pageInfo { count next previous } results { id name diameter mass gravity lengthOfDay atmosphere type { id name } image { url thumbnail credit } description wikiUrl } } }
```

### DIFF
--- a/crates/apollo-mcp-server/src/errors.rs
+++ b/crates/apollo-mcp-server/src/errors.rs
@@ -1,3 +1,5 @@
+use std::path::PathBuf;
+
 use apollo_compiler::{Schema, ast::Document, validation::WithErrors};
 use apollo_federation::error::FederationError;
 use reqwest::header::{InvalidHeaderName, InvalidHeaderValue};
@@ -13,17 +15,17 @@ pub enum OperationError {
     #[error("Internal error: {0}")]
     Internal(String),
 
-    #[error("Operation is missing its required name: {0}")]
-    MissingName(String),
+    #[error("{0}: Operation is missing its required name: {1}")]
+    MissingName(PathBuf, String),
 
-    #[error("No operations defined")]
-    NoOperations,
+    #[error("{0}: No operations defined")]
+    NoOperations(PathBuf),
 
     #[error("Invalid JSON: {0}")]
     Json(#[from] serde_json::Error),
 
-    #[error("Too many operations. Expected 1 but got {0}")]
-    TooManyOperations(usize),
+    #[error("{0}: Too many operations. Expected 1 but got {1}")]
+    TooManyOperations(PathBuf, usize),
 
     #[error(transparent)]
     File(#[from] std::io::Error),
--- a/crates/apollo-mcp-server/src/operations.rs
+++ b/crates/apollo-mcp-server/src/operations.rs
@@ -88,12 +88,22 @@ impl OperationSource {
                                     if entry_path.extension().and_then(|e| e.to_str())
                                         == Some(OPERATION_DOCUMENT_EXTENSION)
                                     {
-                                        if let Ok(content) = fs::read_to_string(&entry_path) {
-                                            // Be forgiving of empty files in the directory case.
-                                            // It likely means a new file was created in an editor,
-                                            // but the operation hasn't been written yet.
-                                            if !content.trim().is_empty() {
-                                                operations.push(RawOperation::from(content));
+                                        match fs::read_to_string(&entry_path) {
+                                            Ok(content) => {
+                                                // Be forgiving of empty files in the directory case.
+                                                // It likely means a new file was created in an editor,
+                                                // but the operation hasn't been written yet.
+                                                if !content.trim().is_empty() {
+                                                    operations.push(RawOperation::from((
+                                                        content, entry_path,
+                                                    )));
+                                                }
+                                            }
+                                            Err(e) => {
+                                                return Some(Event::OperationError(
+                                                    e,
+                                                    entry_path.clone(),
+                                                ));
                                             }
                                         }
                                     }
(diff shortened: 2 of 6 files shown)


crates/apollo-mcp-server/src/errors.rs
```diff
@@ -13,17 +15,17 @@ pub enum OperationError {
     #[error("Internal error: {0}")]
     Internal(String),
 
-    #[error("Operation is missing its required name: {0}")]
-    MissingName(String),
+    #[error("{0}: Operation is missing its required name: {1}")]
+    MissingName(PathBuf, String),
```
Jephuff: sources could be files/manifest file/operation collection/uplink, I think this only makes sense for files and the other sources will end up printing `: Operation is missing its required name:` which is a bit weird. Maybe instead of having this be PathBuf we could make it a more generic `String` that can be set to anything like the file name, the operation collection operation name (different then the operation name), and then I guess the manifest file name or uplink?

crates/apollo-mcp-server/src/operations.rs
```diff
@@ -104,12 +114,13 @@ impl OperationSource {
                             match fs::read_to_string(&path) {
                                 Ok(content) => {
                                     if !content.trim().is_empty() {
-                                        operations.push(RawOperation::from(content));
+                                        operations
+                                            .push(RawOperation::from((content, path.clone())));
```
DaleSeo: Rust noop question: Why do we pass `entry_path.clone()` here to `RawOperation::from` while we pass just `entry_path` to it on line 98?

crates/apollo-mcp-server/src/errors.rs
swcollard: Agree Jeffrey. That makes sense, I'll make the path non-optional and move it to a String so it gets populated by each method.

crates/apollo-mcp-server/src/operations.rs
pubmodmatt: The `RawOperation::from` method takes ownership of the argument passed in. On line 98, it is passing in `entry_path`, which is never used again, so transferring that ownership is fine (`entry_path` is also used on lin 105, but that is in a different match arm so both code paths cannot both execute. The clone on line 105 is actually unnecessary and should be removed).

In this case here on line 118, it's using `path`, which *is* used later in the function, for example on line 128 and 143. So ownership of `path` cannot be transferred into `RawOperation::from` here, because we need it later on. The last clone on line 143 could also be removed, because `path` is not used again after that.

crates/apollo-mcp-registry/src/uplink/persisted_queries/manifest_poller.rs
```diff
@@ -38,7 +38,13 @@ impl ManifestSource {
                     Ok(manifest) => Event::UpdateManifest(
                         manifest
                             .iter()
-                            .map(|(k, v)| (k.operation_id.clone(), v.clone()))
+                            .map(|(k, v)| {
+                                (
+                                    k.operation_id.clone(),
+                                    v.clone(),
+                                    "PersistedQuery".to_string(),
```
swcollard: I couldn't find anything in the actual manifest that was more human readable to call out as the source of the operation. Repeating the ID didn't feel quite right either so I just put "PersistedQuery" as a placeholder.

There is also a separate conversation of whether it makes sense to update this RawOperation interface to include the path, but I didn't see another way to pass the Path down to where the operations are parsed and errors are thrown.

pubmodmatt: It adds no value to the `UpdateManifest` event to include the hard-coded "PersistedQuery" string. I'd move this to the point where you convert `UpdateManifest` to a `RawOperation`, and insert the hard-coded string there, if needed.

crates/apollo-mcp-server/src/operations.rs
```diff
@@ -168,6 +191,7 @@ pub struct RawOperation {
     persisted_query_id: Option<String>,
     headers: Option<HeaderMap<HeaderValue>>,
     variables: Option<HashMap<String, Value>>,
+    source_path: String,
```
pubmodmatt: A `source_path` doesn't make sense for cases other than local operation files or manifest files. It should be optional. In general, when designing Rust types, you should try to make it impossible to create invalid data. Having an empty source path is invalid - either it should have a path, or specify that there is no path for this operation, which is the case `Option` is made for.

PR #121, opened 2025-06-05
docs: update diagram colors

Meschreiber:


DaleSeo:
The digram looks better with the colors!

Meschreiber:
Was running into issues with dark mode. (See https://apollograph.slack.com/archives/C01BMLR2YBG/p1749137512133999)

PR #122, opened 2025-06-05
Update license to MIT

kbychu:
Based on the decision in https://apollographql.atlassian.net/wiki/spaces/GT/pages/1581482098/Apollo+MCP+Server+The+Case+for+MIT+Licensing

We are updating the license to MIT


nix/apollo-mcp.nix
kbychu, line 1: Cursor thinks that in addition to updating line 40, it should update the configuration as well. It says 

"This configuration aligns with how the package is used in flake.nix and provides all the necessary components for building the Apollo MCP server package."

How to test/verify this is correct is beyond me.

kbychu: @nicholascioli do you mind taking a look?

xtask/LICENSE
pubmodmatt, line 1: Just remove this file.

xtask/Cargo.toml
```diff
@@ -5,7 +5,7 @@ name = "xtask"
 version = "1.5.0"
 authors = ["Apollo Graph, Inc. <packages@apollographql.com>"]
 edition = "2021"
-license = "LicenseRef-ELv2"
+license = "MIT"
```
pubmodmatt: Just remove this line

nix/apollo-mcp.nix
pubmodmatt: Whoa. This may be a case of Cursor going off the rails. This should just change the license, I think, and not all these other changes.

crates/apollo-mcp-server/Cargo.toml
```diff
@@ -2,7 +2,7 @@
 name = "apollo-mcp-server"
 version.workspace = true
 edition = "2024"
-license = "Elastic-2.0"
+license = "MIT"
```
pubmodmatt: I think it'd be better to remove the license entries from the individual crates and move this up to the workspace level Cargo.toml. Then in the workspace level file, specify:

```toml
license-file = "./LICENSE"
```

pubmodmatt (changes requested):

pubmodmatt:
This will also need a changelog update. Run:

```sh
cargo xtask changeset create
```

Then commit the file and push it to the PR.

LICENSE
```diff
@@ -1,102 +1,23 @@
...
-
-**your licenses** are all the licenses granted to you for the software under
-these terms.
-
-**use** means anything you do with the software requiring one of your licenses.
-
-**trademark** means trademarks, service marks, and similar rights.
+MIT License
+
+Copyright (c) 2025 Apollo Graph, Inc.
```
DaleSeo, line 3: Just out of curiosity, if we mention 2025 here, should we update it on January 1st of next year?

crates/apollo-mcp-server/Cargo.toml
DaleSeo: +1 for Single Source of Truth!

LICENSE
DaleSeo: I took a quick look at other repositories:

### apollo-server

https://github.com/apollographql/apollo-server/blob/98b1976d86633e0370896fa4d4513ed0689a0d8c/LICENSE#L3

Shot 2025-06-06 at 09 31 35

### rover

https://github.com/apollographql/rover/blob/6c2d0c05f9da4dab6b5f969363ed1da0f06d9044/LICENSE#L9

Shot 2025-06-06 at 09 36 28

kbychu: The copyright year on a software license typically includes the initial year of creation and the current year of modification

DaleSeo: @kbychu Makes sense!

crates/apollo-mcp-server/Cargo.toml
DaleSeo: @pubmodmatt I noticed that there are three `LICENSE` files in the repository. As suggested, I removed the one in the `xtasks` workspace. I also removed the ELv2 license file from the `apollo-mcp-server` workspace and updated `Cargo.toml` in both `apollo-mcp-server` and `apollo-mcp-registry` to reference the `LICENSE` files at the project root.

.changesets/breaking_kchu_mit_license.md
pubmodmatt, line 1: Don't make this breaking.  Maybe feature.

pubmodmatt (approved):

.changesets/feature_kchu_mit_license.md
```diff
@@ -0,0 +1,3 @@
+### Update license to MIT - @kbychu PR #122
+
+Updating the license to MIT
```
pubmodmatt: Nit - the change set should phrase this in terms of what it changes for the user, like "The Apollo MCP Server is now licensed under MIT instead of ELv2"

PR #124, opened 2025-06-06
docs: use 127.0.0.1 instead of localhost

DaleSeo:
I would like to update the user guide documentation to use `127.0.0.1` instead of `localhost` when connecting to the MCP server from the MCP inspector. I have encountered this issue several times and noticed that other folks are experiencing the same problem. This change should help prevent similar frustration for first-time users.

### DIFF
--- a/docs/source/guides/index.mdx
+++ b/docs/source/guides/index.mdx
@@ -264,7 +264,6 @@ You can also deploy the server as a container using the instructions in [Deployi
 ```sh
 target/debug/apollo-mcp-server \
   --directory <absolute path to this git repo> \
-  --http-address 127.0.0.1 \
   --http-port 5000 \
   --schema graphql/weather/api.graphql \
   --operations graphql/weather/operations/forecast.graphql graphql/weather/operations/alerts.graphql graphql/weather/operations/all.graphql


docs/source/guides/index.mdx
```diff
@@ -264,6 +264,7 @@ You can also deploy the server as a container using the instructions in [Deployi
 ```sh
 target/debug/apollo-mcp-server \
   --directory <absolute path to this git repo> \
+  --http-address 127.0.0.1 \
```
pubmodmatt: this is unnecessary - it's the default

docs/source/guides/index.mdx
```diff
@@ -277,6 +278,6 @@ npx @modelcontextprotocol/inspector
 
 1. In a browser, go to the URL returned by Inspector, then fill in the details:
     - **Transport Type**: Select `Streamable HTTP`
-    - **URL**: Enter `http://localhost:5000/mcp`, where the port must match the `--http-port` option
+    - **URL**: Enter `http://127.0.0.1:5000/mcp`, where the port must match the `--http-port` option
```
pubmodmatt, line 280: The issue is that we've had problems both ways. For some users, only localhost works. There is already mention of this in the troubleshooting section. But I'm ok with updating this default - I think it's what MCP inspector uses out of the box so its consistent.

pubmodmatt: And all of this is wrong in an IPv6-only environment, where it should be `::`

docs/source/guides/index.mdx
DaleSeo: I thought this could be extra clearer for first-time users, but it might be unnecessary since it defaults to `127.0.0.1` if not specified. Please let me know if anyone opposes this.

pubmodmatt: We suffer from too many options at the moment. I think we should specify just the bare minimum to get it running.

docs/source/guides/index.mdx
pubmodmatt: Also, we need to remove MCP inspector entirely from this page, IMO, so it's all kind of moot.

docs/source/guides/index.mdx
DaleSeo: I agree. That's why I was hesitant. I've removed it.

pubmodmatt (approved):

docs/source/guides/index.mdx
DaleSeo: > There is already mention of this in the troubleshooting section.

@pubmodmatt Where can I find that section?

pubmodmatt: https://www.apollographql.com/docs/apollo-mcp-server/quickstart#troubleshooting

PR #125, opened 2025-06-06
Reorganize introspection module

pubmodmatt:
No new functionality - this just reoganizes the code in preparation for doing more work in this area.


crates/apollo-mcp-server/src/introspection.rs
```diff
@@ -1,311 +1,3 @@
...
-            Executable::variables(&execute, input),
-            Err(McpError::new(
-                ErrorCode::INVALID_PARAMS,
-                "Invalid input".to_string(),
-                None
-            ))
-        );
-    }
-}
+pub(crate) mod tools;
```
DaleSeo, line 3: What do you think about replacing the `introspection` folder with the `tools` folder to simplify the structure? I feel that the folder name doesn't represent well what is in it but I might be off. 😆

crates/apollo-mcp-server/src/introspection/tools.rs
```diff
@@ -0,0 +1,4 @@
+//! MCP tools to allow an AI agent to introspect a GraphQL schema and execute operations.
+
+pub(crate) mod execute;
+pub(crate) mod introspect;
```
DaleSeo, line 4: Oh, how neat! ✨

DaleSeo (approved):
Thanks for the refactor! I appreciate the separation by tool!

crates/apollo-mcp-server/src/introspection.rs
pubmodmatt: These tools are turned on and off by the `--introspection` option, so that's where the name comes from. This differentiates it from other tools like the `explorer` tool which is turned on by a different option. I'm also future-proofing a bit here as there may be other tools going forward.

PR #134, opened 2025-06-12
Allow passing in url overrides

Jephuff:
add APOLLO_UPLINK_ENDPOINTS and APOLLO_REGISTRY_URL as args to override registry endpoints the same way router and rover do

### DIFF
--- a/.changesets/feat_jeffrey_allow_passing_in_url_overrides.md
+++ b/.changesets/feat_jeffrey_allow_passing_in_url_overrides.md
@@ -1,0 +1,3 @@
+### Allow overriding registry endpoints 
+
+read APOLLO_UPLINK_ENDPOINTS and APOLLO_REGISTRY_URL to override endpoints for fetching schemas and operations
--- a/crates/apollo-mcp-registry/src/platform_api.rs
+++ b/crates/apollo-mcp-registry/src/platform_api.rs
@@ -1,11 +1,14 @@
 use secrecy::SecretString;
 use std::fmt::Debug;
 use std::time::Duration;
+use url::Url;
 
 pub mod operation_collections;
 
+const DEFAULT_PLATFORM_API: &str = "https://registry.apollographql.com/api/graphql";
+
 /// Configuration for polling Apollo Uplink.
-#[derive(Clone, Debug, Default)]
+#[derive(Clone, Debug)]
 pub struct PlatformApiConfig {
     /// The Apollo key: `<YOUR_GRAPH_API_KEY>`
     pub apollo_key: SecretString,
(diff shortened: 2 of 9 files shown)


.changesets/feat_jeffrey_allow_passing_in_url_overrides.md
```diff
@@ -0,0 +1,3 @@
+### Allow overriding registry endpoints 
+
+read APOLLO_UPLINK_ENDPOINTS and APOLLO_REGISTRY_URL to override endpoints for fetching schemas and operations
```
pubmodmatt: Nit: frame changes from the user perspective, not your perspective.
```suggestion
Set APOLLO_UPLINK_ENDPOINTS and APOLLO_REGISTRY_URL to override the endpoints for fetching schemas and operations
```

pubmodmatt: Also, are there any other endpoints for end users to set? Or is this just for our internal use? If the former, let's just remove the changeset.

crates/apollo-mcp-registry/src/platform_api.rs
```diff
@@ -15,4 +18,26 @@ pub struct PlatformApiConfig {
...
+        timeout: Duration,
+        registry_url: Option<Url>,
+    ) -> Self {
+        Self {
+            apollo_key,
+            poll_interval,
+            timeout,
+            #[allow(clippy::expect_used)]
+            registry_url: registry_url
+                .unwrap_or(Url::parse(DEFAULT_PLATFORM_API).expect("default URL should be valid")),
```
pubmodmatt, line 40: If someone were to break the DEFAULT_PLATFORM_API URL, it wouldn't be caught until someone ran without the env variable set. Presumably that would happen quickly, but it'd better to add a simple unit test.

crates/apollo-mcp-server/src/main.rs
```diff
@@ -118,6 +119,69 @@ struct Args {
...
+
+    #[clap(env)]
+    apollo_registry_url: Option<String>,
+
+    /// Your Apollo key.
+    #[clap(skip = std::env::var("APOLLO_KEY").ok())]
+    apollo_key: Option<String>,
+
+    /// Your Apollo graph reference.
+    #[clap(skip = std::env::var("APOLLO_GRAPH_REF").ok())]
```
pubmodmatt, line 146: There are other place where these variables are used, such as when creating tools - can you update those to use this as well?

crates/apollo-mcp-server/src/main.rs
```diff
@@ -118,6 +119,69 @@ struct Args {
     /// collection id to expose as MCP tools (requires APOLLO_KEY)
     #[arg(long, conflicts_with_all(["operations", "manifest"]))]
     collection: Option<String>,
+
+    /// The endpoints (comma separated) polled to fetch the latest supergraph schema.
+    #[clap(long, env, action = ArgAction::Append)]
+    // Should be a Vec<Url> when https://github.com/clap-rs/clap/discussions/3796 is solved
+    apollo_uplink_endpoints: Option<String>,
```
pubmodmatt, line 136: Same question as above - do we expect end users to use this? If not, these options shouldn't be in the help output.

pubmodmatt:
If these variables are something we want to expose to customers, they should also be added to the doc - unfortunately there is no way to update the doc at the time the next release goes out - it deploys immediately. So this will require a follow up doc PR after the next release. We need a better process here, but for now it'd be good to queue up a separate doc PR.

.changesets/feat_jeffrey_allow_passing_in_url_overrides.md
Jephuff: I think it's more for internal use, so I can remove it!

crates/apollo-mcp-registry/src/platform_api.rs
Jephuff: ohh unit test, great idea! 
I wanted to make it const, but that doesn't work with the Url::parse

.changesets/feat_jeffrey_allow_passing_in_url_overrides.md
Jephuff: maybe not exclusively for internal use, it is documented for router so I'll keep this for now.

PR #137, opened 2025-06-13
Add client metadata to GraphQL requests

pubmodmatt:
* Add `ApolloClientMetadata` for GraphQL requests. This will allow us to collect usage data for traffic from the MCP server that goes through Apollo Router
* Add a `x-apollo-mcp-server-version` header to Uplink requests. This is similar to the header router uses, `x-router-version`, and should allow collecting usage data for calls to Uplink from the MCP server
* Change the crates to pick up `workspace.package.version` and `workspace.package.author`. This ensures that `CARGO_PKG_VERSION` will be the top level Apollo MCP Server version
* Fix broken arguments that prevent the MCP server from starting up
* Fix clippy warnings

Example of router logs with the new metadata:

`2025-06-13T13:33:18.616606Z INFO router{http.route=/,http.request.method=POST,http.request.method=POST,http.request.method=POST,http.route=/,url.path=/,client.name=my-web-app,client.version=,}  http.request.version=HTTP/1.1 http.request.method=POST http.request.uri=/ http.request.headers={"apollographql-client-name": "my-web-app", "content-type": "application/json", "accept": "*/*", "accept-encoding": "gzip", "host": "127.0.0.1:4000", "content-length": "205"} http.request.body="{\"query\":\"query GetAlerts($state: String!) { alerts(state: $state) { severity description instruction } }\",\"variables\":{\"state\":\"AK\"},\"extensions\":{\"ApolloClientMetadata\":{\"type\":\"mcp\",\"version\":\"0.3.0\"}}}"  kind=router.request
`

### DIFF
--- a/crates/apollo-mcp-server/src/graphql.rs
+++ b/crates/apollo-mcp-server/src/graphql.rs
@@ -28,6 +28,10 @@ pub trait Executable {
 
     /// Execute as a GraphQL operation using the endpoint and headers
     async fn execute(&self, request: Request<'_>) -> Result<CallToolResult, McpError> {
+        let client_metadata = serde_json::json!({
+            "type": "mcp",
+            "version": std::env!("CARGO_PKG_VERSION")
+        });
         reqwest::Client::new()
             .post(request.endpoint)
             .headers(self.headers(&request.headers))
@@ -38,6 +42,7 @@ pub trait Executable {
                             "version": 1,
                             "sha256Hash": id,
                         },
+                        "ApolloClientMetadata": client_metadata,
                     },
                     "variables": self.variables(request.input)?,
                 })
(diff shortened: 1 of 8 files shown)


crates/apollo-mcp-server/src/graphql.rs
```diff
@@ -28,6 +28,10 @@ pub trait Executable {
 
     /// Execute as a GraphQL operation using the endpoint and headers
     async fn execute(&self, request: Request<'_>) -> Result<CallToolResult, McpError> {
+        let client_metadata = serde_json::json!({
+            "type": "mcp",
+            "version": std::env!("CARGO_PKG_VERSION")
+        });
```
swcollard, line 34: Is this something the router already knows how to parse? Or is it extra data collection we'll need to add there?

pubmodmatt: It'll work out of the box! This is using existing support for identifying clients like mobile vs web.

PR #129, opened 2025-06-09
validate required environment variables for explorer option

DaleSeo:
The `--uplink` and `explorer` options require `APOLLO_KEY` and `APOLLO_GRAPH_REF`. This PR adds validation logic for the `--explorer` option, similar to that of the `--uplink` option.

Shot 2025-06-09 at 16 15 03

PR #140, opened 2025-06-13
fix env args with requires

Jephuff:
The args I copied from rover and router were using skip, but clap supports envs directly. Using it this way fixes the requires problem

### DIFF
--- a/crates/apollo-mcp-server/src/main.rs
+++ b/crates/apollo-mcp-server/src/main.rs
@@ -81,13 +81,8 @@ struct Args {
     )]
     uplink: bool,
 
-    /// Expose a tool to open queries in Apollo Explorer (requires APOLLO_KEY and APOLLO_GRAPH_REF)
-    #[arg(
-        long,
-        short = 'x',
-        requires = "apollo_key",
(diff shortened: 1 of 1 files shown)


crates/apollo-mcp-server/src/main.rs
```diff
@@ -73,11 +73,21 @@ struct Args {
...
+        requires = "apollo_graph_ref"
+    )]
     uplink: bool,
 
     /// Expose a tool to open queries in Apollo Explorer (requires APOLLO_KEY and APOLLO_GRAPH_REF)
-    #[arg(long, short = 'x')]
+    #[arg(
+        long,
+        short = 'x',
+        requires = "apollo_key",
```
DaleSeo: This option doesn't require `APOLLO_KEY` as discussed in https://github.com/apollographql/apollo-mcp-server/pull/129#discussion_r2136516203

Jephuff: Oops, you are right!  removing 👍

Jephuff: Fix the description and the requires 👍

PR #146, opened 2025-06-18
Operation Collections - fix update checking

Jephuff:
The update checking was off so it would update all operations on every poll. This will fix it so that it only deletes removed operations and will only update the tool list if something has changed.

### DIFF
--- a/.changesets/fix_jeffrey_fix_collection_update_checking.md
+++ b/.changesets/fix_jeffrey_fix_collection_update_checking.md
@@ -1,0 +1,3 @@
+### Fix tool update on every poll - @Jephuff PR #146
+
+Only update the tool list if an operation was removed, changed, or added.
--- a/crates/apollo-mcp-registry/src/platform_api/operation_collections/collection_poller.rs
+++ b/crates/apollo-mcp-registry/src/platform_api/operation_collections/collection_poller.rs
@@ -78,36 +78,36 @@ async fn handle_poll_result(
 
     if changed_ids.is_empty() && removed_ids.is_empty() {
         tracing::debug!("no operation changed");
-        Ok(None)
-    } else {
-        if !removed_ids.is_empty() {
-            tracing::info!("removed operation ids: {:?}", removed_ids);
-            for id in removed_ids {
-                previous_updated_at.remove(id);
-            }
-        }
-
-        if !changed_ids.is_empty() {
-            tracing::info!("changed operation ids: {:?}", changed_ids);
-            let full_response = graphql_request::<OperationCollectionEntriesQuery>(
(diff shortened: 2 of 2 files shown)


.changesets/fix_jeffrey_fix_collection_update_checking.md
```diff
@@ -0,0 +1,3 @@
+### Fix tool update on every poll - @Jephuff PR #146
```
DaleSeo, line 1: 💯

crates/apollo-mcp-registry/src/platform_api/operation_collections/collection_poller.rs
```diff
@@ -60,12 +60,11 @@ async fn handle_poll_result(
...
-    for id in previous_updated_at.clone().keys() {
-        if keep_ids.all(|keep_id| keep_id != id) {
-            previous_updated_at.remove(id);
-        }
-    }
+    let removed_ids = previous_updated_at.clone();
+    let removed_ids = removed_ids
+        .keys()
+        .filter(|id| poll.iter().all(|(keep_id, _)| keep_id != *id))
+        .collect::<Vec<_>>();
```
DaleSeo, line 67: It seems like an unnecessary clone.

```suggestion
    let removed_ids: Vec<&String> = previous_updated_at
        .keys()
        .filter(|id| poll.iter().all(|(keep_id, _)| keep_id != *id))
        .collect();
```

crates/apollo-mcp-registry/src/platform_api/operation_collections/collection_poller.rs
```diff
@@ -77,26 +76,34 @@ async fn handle_poll_result(
         })
         .collect();
 
-    if changed_ids.is_empty() {
+    if changed_ids.is_empty() && removed_ids.is_empty() {
         tracing::debug!("no operation changed");
         Ok(None)
```
DaleSeo: Can we return early here, like with `return Ok(None);` to flatten the nested if statements below? Or is this not a good practice in Rust?

DaleSeo (approved):
Approving with a couple minor comments 👍

crates/apollo-mcp-registry/src/platform_api/operation_collections/collection_poller.rs
Jephuff: If I do that I get errors later on telling me previous_updated_at is borrowed :(

crates/apollo-mcp-registry/src/platform_api/operation_collections/collection_poller.rs
Jephuff: I think thats fine! I'mm update

PR #147, opened 2025-06-19
feat: retain argument descriptions for variables

alocay:
Pulling descriptions from the schema for input variables of an operation. Looking for descriptions in both fields and directives and in the case where multiple descriptions are found, the result will be concatenation of all the descriptions separated by a `#` delimiter.

### DIFF
--- a/crates/apollo-mcp-server/src/schema_tree_shake.rs
+++ b/crates/apollo-mcp-server/src/schema_tree_shake.rs
@@ -543,14 +543,14 @@ fn retain_argument_descriptions(
     arg: &Node<InputValueDefinition>,
     operation_arguments: &HashMap<&str, &Name>,
 ) {
-    let variable_name = operation_arguments.get(arg.name.as_str());
+    let operation_argument_name = operation_arguments.get(arg.name.as_str());
 
-    if let Some(variable_name) = variable_name {
+    if let Some(op_arg_name) = operation_argument_name {
         if let Some(description) = arg.description.as_deref() {
             if !description.trim().is_empty() {
                 let descriptions = tree_shaker
                     .arguments_descriptions
-                    .entry(variable_name.to_string())
+                    .entry(op_arg_name.to_string())
                     .or_default();
                 descriptions.push(description.trim().to_string())
             }


crates/apollo-mcp-server/src/schema_tree_shake.rs
```diff
@@ -531,6 +535,22 @@ fn selection_set_to_fields(
     }
 }
 
+fn retain_variable_descriptions(
```
alocay: Function added to handle retaining descriptions for directives and fields

crates/apollo-mcp-server/src/schema_tree_shake.rs
```diff
@@ -664,6 +690,10 @@ fn retain_type(
                         }
                         if let Some(field_selection_directives) = field_selection_directives {
                             field_selection_directives.iter().for_each(|directive| {
+                                directive.arguments.iter().for_each(|arg| {
```
alocay: Would like to use the same function for directive once i can figure out how to get the description

crates/apollo-mcp-server/src/schema_tree_shake.rs
```diff
@@ -1,17 +1,12 @@
 //! Tree shaking for GraphQL schemas
 
-use apollo_compiler::ast::{
-    Definition, DirectiveDefinition, DirectiveList, Document, EnumTypeDefinition, Field,
-    FieldDefinition, FragmentDefinition, InputObjectTypeDefinition, InterfaceTypeDefinition,
-    NamedType, ObjectTypeDefinition, OperationDefinition, OperationType, ScalarTypeDefinition,
-    SchemaDefinition, Selection, Type, UnionTypeDefinition,
-};
+use apollo_compiler::ast::{Definition, DirectiveDefinition, DirectiveList, Document, EnumTypeDefinition, Field, FieldDefinition, FragmentDefinition, InputObjectTypeDefinition, InterfaceTypeDefinition, NamedType, ObjectTypeDefinition, OperationDefinition, OperationType, ScalarTypeDefinition, SchemaDefinition, Selection, Type, UnionTypeDefinition};
```
alocay: Removed unused imports

crates/apollo-mcp-server/src/schema_tree_shake.rs
```diff
@@ -11,7 +11,7 @@ use apollo_compiler::validation::WithErrors;
 use apollo_compiler::{Name, Node, Schema};
 use std::collections::HashMap;
 
-struct RootOperationNames {
+pub struct RootOperationNames {
```
Jephuff: I think we don't need to add pub to this, `new` or `name_for_operation_type` anymore, it might just been from our earlier testing?

crates/apollo-mcp-server/src/schema_tree_shake.rs
```diff
@@ -170,6 +176,10 @@ impl<'schema> SchemaTreeShaker<'schema> {
         document: &Document,
         depth_limit: DepthLimit,
     ) {
+        operation.variables.iter().for_each(|v| {
+            self.arguments_descriptions
+                .insert(v.name.to_string(), Vec::default());
+        });
```
Jephuff: I don't think we actually need to do this, we already initialize with a hashmap. And everywhere that writes or reads from this is already handling the case where it's not defined yet
```suggestion
```

alocay: You're right, the retain function is already using a default vec if not found 👍🏼

crates/apollo-mcp-server/src/schema_tree_shake.rs
alocay: You're probably right, this likely just old artifacts

crates/apollo-mcp-server/src/schema_tree_shake.rs
```diff
@@ -531,6 +538,35 @@ fn selection_set_to_fields(
...
+    operation_arguments: &HashMap<&str, &Name>,
+) {
+    let variable_name = operation_arguments.get(field_type_argument_name);
+
+    if let Some(variable_name) = variable_name {
+        let descriptions = tree_shaker
+            .arguments_descriptions
+            .entry(variable_name.to_string())
+            .or_default();
+        if let Some(description) = field_type_argument_description {
```
Jephuff:
```suggestion
    arg: &Node<InputValueDefinition>,
    operation_arguments: &HashMap<&str, &Name>,
) {
    let variable_name = operation_arguments.get(arg.name.as_str());

    if let Some(variable_name) = variable_name {
        let descriptions = tree_shaker
            .arguments_descriptions
            .entry(variable_name.to_string())
            .or_default();
        if let Some(description) = arg.description.as_deref() {
```

crates/apollo-mcp-server/src/schema_tree_shake.rs
```diff
@@ -531,6 +538,35 @@ fn selection_set_to_fields(
...
+}
+
+fn build_argument_name_to_value_map(arguments: &[Node<Argument>]) -> HashMap<&str, &Name> {
+    arguments
+        .iter()
+        .filter_map(|a| a.value.as_variable().map(|v| (a.name.as_str(), v)))
+        .collect::<HashMap<_, _>>()
+}
+
+#[unsafe(no_mangle)]
```
Jephuff:
```suggestion
```

crates/apollo-mcp-server/src/schema_tree_shake.rs
```diff
@@ -664,6 +715,24 @@ fn retain_type(
...
                         if let Some(field_selection_directives) = field_selection_directives {
                             field_selection_directives.iter().for_each(|directive| {
+                                let directive_args_map =
+                                    build_argument_name_to_value_map(&directive.arguments);
+
+                                if let Some(directive_definition) = tree_shaker
+                                    .schema
+                                    .directive_definitions
+                                    .get(&directive.name)
+                                {
```
Jephuff: Small tweak, we can make the directive_args_map a little lazier if we move it inside the if
```suggestion
                                if let Some(directive_definition) = tree_shaker
                                    .schema
                                    .directive_definitions
                                    .get(&directive.name)
                                {
                                    let directive_args_map =
                                        build_argument_name_to_value_map(&directive.arguments);
```

crates/apollo-mcp-server/src/schema_tree_shake.rs
alocay: Thanks, this was left over from debugging.

crates/apollo-mcp-server/src/schema_tree_shake.rs
alocay: Good catch. Originally I had the separate values passed in directly rather than the argument itself because I was under the impression that directive arguments and field arguments were different types although now that I know they ar the same this is a better way.

crates/apollo-mcp-server/src/operations.rs
```diff
@@ -41,6 +41,7 @@ use std::sync::{Arc, Mutex};
 use tracing::{debug, info, warn};
 
 const OPERATION_DOCUMENT_EXTENSION: &str = "graphql";
+const SEPARATOR: &str = "\n---\n";
```
pubmodmatt: Nitpick - this separator seems a bit heavyweight for an argument - I've used this to separate larger things like type definitions, but even then I kind of regret it since it's too many tokens and newlines aren't displayed well in some clients

crates/apollo-mcp-server/src/schema_tree_shake.rs
```diff
@@ -607,7 +642,13 @@ fn retain_type(
                     def.fields
                         .iter()
                         .map(|(name, field_definition)| {
-                            (name.as_str(), Some(field_definition), None, None)
+                            (
+                                name.as_str(),
+                                Some(field_definition),
+                                None,
+                                None,
+                                HashMap::default(),
```
pubmodmatt, line 650: Why not build the argument map in this case?

crates/apollo-mcp-server/src/schema_tree_shake.rs
```diff
@@ -531,6 +538,33 @@ fn selection_set_to_fields(
...
+) {
+    let variable_name = operation_arguments.get(arg.name.as_str());
+
+    if let Some(variable_name) = variable_name {
+        let descriptions = tree_shaker
+            .arguments_descriptions
+            .entry(variable_name.to_string())
+            .or_default();
+        if let Some(description) = arg.description.as_deref() {
+            if !description.trim().is_empty() {
```
Jephuff: 1 tiny little optimization, we can also do this only need to default the entry if we have a description, so this can be moved inside the ifs

crates/apollo-mcp-server/src/schema_tree_shake.rs
```diff
@@ -531,6 +538,33 @@ fn selection_set_to_fields(
     }
 }
 
+fn retain_variable_descriptions(
```
pubmodmatt: I find the mix of `variable` and `argument` names confusing, especially since variables have a different meaning in GraphQL. We're really only concerned with arguments here, right?

crates/apollo-mcp-server/src/schema_tree_shake.rs
```diff
@@ -531,6 +538,33 @@ fn selection_set_to_fields(
...
+    tree_shaker: &mut SchemaTreeShaker,
+    arg: &Node<InputValueDefinition>,
+    operation_arguments: &HashMap<&str, &Name>,
+) {
+    let variable_name = operation_arguments.get(arg.name.as_str());
+
+    if let Some(variable_name) = variable_name {
+        let descriptions = tree_shaker
+            .arguments_descriptions
+            .entry(variable_name.to_string())
```
pubmodmatt: Won't this result in collisions if more than one field has a variable with the same name?

crates/apollo-mcp-server/src/schema_tree_shake.rs
alocay: Yeah the input arguments of the operation. I can adjust the names used.

crates/apollo-mcp-server/src/schema_tree_shake.rs
alocay: Do you mean a case like the following:
```
query TestQuery($id: ID, $other: String) {
    someQuery(id: $id, other: $other, otherArg: $other) {
        value
    }
}
```
I use a vector to store all the descriptions found for potential matching arguments. In this case `$other` could have the schema description for `other` or `otherArg`.

crates/apollo-mcp-server/src/schema_tree_shake.rs
alocay: I'll move it

crates/apollo-mcp-server/src/schema_tree_shake.rs
alocay: So I might be misunderstanding something here since the naming throws me off. The `field` used in the `.map()` above is the field from the the operation (I think that's right) of type `Argument` and the `field_definition` in the `.unwrap_or()` is the field from the schema of type `InputValueDefinition`. The build map function is creating a mapping between the field's argument to the operation input argument, which I can't do within the `.unwrap_or()` block.

crates/apollo-mcp-server/src/operations.rs
alocay: I can use something else like `#` or `;`. Not sure if a character like `;` is likely to be found inside of a description so the `#` might be safer.

Jephuff: Feels like theres no real good separate here, so what we really need is the override for cases where there are a bunch of argument descriptions being joined together  .

crates/apollo-mcp-server/src/schema_tree_shake.rs
Jephuff: yep thats right! the arguments map is argument to variables, all be have in this case is the schema definition which doesn't have variables.

crates/apollo-mcp-server/src/schema_tree_shake.rs
Jephuff: I think in this case we are building a "variable" description based on "argument" descriptions. 
so we are retaining arguments and organizing them into vecs for each variable.

crates/apollo-mcp-server/src/schema_tree_shake.rs
alocay: Just FYI I'm currently working on grabbing variable descriptions from operation comments to override schema descriptions. For cases like these it'll likely be helpful.

PR #149, opened 2025-06-20
release: v0.4.1

alocay:
[0.4.1] - 2025-06-20
🐛 Fixes
Fix tool update on every poll - @Jephuff PR #146
Only update the tool list if an operation was removed, changed, or added.


alocay:
> Looks good, but instead of chore, I'd title this PR (and the squashed commit) "release: v0.4.1" since that's the style we've been using previously.

Sounds good, I amended the message

PR #151, opened 2025-06-20
Add default collection support

Jephuff:
This adds support for reading the default mcp server tool collection. Right now it uses the `--collection default` arg as the way to configure this. But I'm open to alternatives.

### DIFF
--- a/crates/apollo-mcp-registry/src/platform_api/operation_collections/collection_poller.rs
+++ b/crates/apollo-mcp-registry/src/platform_api/operation_collections/collection_poller.rs
@@ -9,10 +9,14 @@ use tokio_stream::wrappers::ReceiverStream;
 
 use super::{error::CollectionError, event::CollectionEvent};
 use crate::platform_api::PlatformApiConfig;
-use operation_collection_default_polling_query::OperationCollectionDefaultPollingQueryVariant;
+use operation_collection_default_polling_query::{
+    OperationCollectionDefaultPollingQueryVariant as PollingDefaultGraphVariant,
+    OperationCollectionDefaultPollingQueryVariantOnGraphVariantMcpDefaultCollection as PollingDefaultCollection,
+};
 use operation_collection_default_query::{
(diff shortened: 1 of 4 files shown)


crates/apollo-mcp-registry/src/platform_api/operation_collections/collection_poller.rs
```diff
@@ -318,6 +378,134 @@ impl CollectionSource {
...
+                            tracing::debug!(
+                                "failed to send error to collection stream. This is likely to be because the server is shutting down: {e}"
+                            );
+                            return;
+                        }
+                    }
+                    None => {
+                        if let Err(e) = sender
+                            .send(CollectionEvent::CollectionError(CollectionError::Response(
+                                "{graphRef} not found".to_string(),
```
pubmodmatt:
```suggestion
                                format!("{graph_ref} not found"),
```

crates/apollo-mcp-registry/src/platform_api/operation_collections/collection_poller.rs
```diff
@@ -218,6 +275,9 @@ impl CollectionSource {
             CollectionSource::Id(id, platform_api_config) => {
                 self.collection_id_stream(id.clone(), platform_api_config.clone())
             }
+            CollectionSource::Default(graph_ref, platform_api_config) => {
+                self.default_collection_stream(graph_ref.clone(), platform_api_config.clone())
+            }
```
pubmodmatt: If you use `ref` here and in the other match arm, you can change the signature to `into_stream(self)`:

```suggestion
            CollectionSource::Default(ref graph_ref, ref platform_api_config) => {
                self.default_collection_stream(graph_ref.clone(), platform_api_config.clone())
            }
```

pubmodmatt (approved):
Couple small things but otherwise looks good. I can't think of a better way than `--collection default` for now.

PR #153, opened 2025-06-23
Add token forwarding to sse call_tool

maxiozer:


DaleSeo:
Hi @maxiozer, Token passthrough is explicitly forbidden in the MCP spec: https://modelcontextprotocol.io/specification/draft/basic/security_best_practices#token-passthrough

PR #154, opened 2025-06-23
Change default schema and operation source behaviour

Jephuff:
- removes `--uplink`
    - uplink schema will be attempted automatically if no schema is passed in
- add `--uplink-manifest` as a new way to explicitly get persisted query manifest from uplink
- default to the default mcp tools if no other operation source is passed in

### DIFF
--- a/crates/apollo-mcp-server/src/main.rs
+++ b/crates/apollo-mcp-server/src/main.rs
@@ -91,7 +91,7 @@ struct Args {
     operations: Vec<PathBuf>,
 
     /// The path to the persisted query manifest containing operations
-    #[arg(long, conflicts_with_all(["operations", "collection, uplink_manifest"]))]
+    #[arg(long, conflicts_with_all(["operations", "collection", "uplink_manifest"]))]
     manifest: Option<PathBuf>,
 
     /// collection id to expose as MCP tools, or `default` to expose the default tools for the variant (requires APOLLO_KEY)
--- a/crates/apollo-mcp-server/src/main.rs
+++ b/crates/apollo-mcp-server/src/main.rs
@@ -72,14 +72,9 @@ struct Args {
     #[arg(long, short = 'i')]
     introspection: bool,
 
-    /// Enable use of uplink to get the schema and persisted queries (requires APOLLO_KEY and APOLLO_GRAPH_REF)
-    #[arg(
-        long,
-        short = 'u',
-        requires = "apollo_key",
-        requires = "apollo_graph_ref"
-    )]
-    uplink: bool,
+    /// Enable use of uplink to get the persisted queries (requires APOLLO_KEY and APOLLO_GRAPH_REF)
+    #[arg(long, requires = "apollo_key", requires = "apollo_graph_ref")]
+    uplink_manifest: bool,
 
     /// Expose a tool to open queries in Apollo Explorer (requires APOLLO_GRAPH_REF)
     #[arg(long, short = 'x', requires = "apollo_graph_ref")]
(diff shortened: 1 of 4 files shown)


pubmodmatt:
This will break `rover dev`, since that uses `--uplink`. May also impact the AIO container.

Jephuff:
> This will break rover dev, since that uses --uplink. May also impact the AIO container.  

Yeah, this definitely is a bigger breaking change. If we are good with the direction, what I was thinking we should do is release with keeping the `--uplink` flag as deprecated that will act like an alias for `--uplink-manifest`. That should make it a no op for existing usages.

crates/apollo-mcp-server/src/main.rs
```diff
@@ -237,34 +243,36 @@ async fn main() -> anyhow::Result<()> {
...
         }
-    } else if args.uplink {
+    } else if args.apollo_graph_ref.is_some() {
         SchemaSource::Registry(args.uplink_config()?)
     } else {
         bail!(ServerError::NoSchema);
     };
 
+    let collection_id = args.collection.as_ref().and_then(|c| {
+        if c == "default" {
```
pubmodmatt: Should we handle variations in case here like `Default` or `DEFAULT`?

pubmodmatt (approved):

crates/apollo-mcp-server/src/main.rs
Jephuff: I like that idea! no need for this to be case sensitive 👍

crates/apollo-mcp-server/src/main.rs
```diff
@@ -72,27 +72,41 @@ struct Args {
     #[arg(long, short = 'i')]
     introspection: bool,
 
-    /// Enable use of uplink to get the schema and persisted queries (requires APOLLO_KEY and APOLLO_GRAPH_REF)
+    /// Expose a tool to open queries in Apollo Explorer (requires APOLLO_GRAPH_REF)
```
DaleSeo: This sounds like the old behavior.

```suggestion
    /// Expose a tool that returns the URL to open a GraphQL operation in Apollo Explorer (requires APOLLO_GRAPH_REF)
```

crates/apollo-mcp-server/src/main.rs
```diff
@@ -72,27 +72,41 @@ struct Args {
...
-    explorer: bool,
+    uplink_manifest: bool,
 
     /// Operation files to expose as MCP tools
     #[arg(long = "operations", short = 'o', num_args=0..)]
     operations: Vec<PathBuf>,
 
     /// The path to the persisted query manifest containing operations
-    #[arg(long, conflicts_with_all(["operations", "collection"]))]
+    #[arg(long, conflicts_with_all(["operations", "collection", "uplink_manifest"]))]
```
DaleSeo, line 104: Why it conflicts with `--uplink_manifest`? It didn't conflict with `--uplink`.

crates/apollo-mcp-server/src/main.rs
```diff
@@ -237,34 +247,36 @@ async fn main() -> anyhow::Result<()> {
             path: path.clone(),
             watch: true,
         }
-    } else if args.uplink {
+    } else if args.apollo_graph_ref.is_some() {
```
DaleSeo, line 260: What if someone sets `APOLLO_GRAPH_REF` not to get the persisted queries, but just for the `--explorer` option?

crates/apollo-mcp-server/src/main.rs
Jephuff: uplink did schema + operations, so `--uplink` with an operation source still made sense because you were getting the schema. But now `--uplink-manifest` only gets the operations, so it doesn't make sense with another source because we only allow 1. 

Though, this reminds me that --uplink is an alias, and this might break existing code. I'm going to test and update to just log a warning if it breaks 👍

crates/apollo-mcp-server/src/main.rs
Jephuff: then they can pass a schema in with `--schema`. they need one either way, so this is just the default if a schema isn't passed

crates/apollo-mcp-server/src/main.rs
DaleSeo: Thanks for clarifying, @Jephuff. Just to confirm my understanding, `--uplink` only polled persisted queries when there were no other operation sources specified?

crates/apollo-mcp-server/src/main.rs
DaleSeo: Makes sense, @Jephuff! They would pass a schema.

DaleSeo (approved):

crates/apollo-mcp-server/src/main.rs
Jephuff: yep 👍

PR #158, opened 2025-06-25
Document why we don't support token passthrough

DaleSeo:
This adds a section in our docs that explicitly call out why we're not supporting token passthrough.

### DIFF
--- a/docs/source/best-practices.mdx
+++ b/docs/source/best-practices.mdx
@@ -39,4 +39,3 @@ To maintain clear trust boundaries, MCP servers must only accept tokens explicit
 Forwarding client tokens downstream is not allowed.
 
 Our team is actively working on a robust authentication mechanism that follows OAuth 2.1 best practices and aligns with the MCP authorization model.
-Our goal is to deliver a secure, standards-compliant, and maintainable solution.


docs/source/best-practices.mdx
```diff
@@ -16,11 +16,27 @@ GraphOS [contract variants](/graphos/platform/schema-management/delivery/contrac
...
+
+- **Audience confusion**: If the MCP Server accepts tokens not intended for it, it can violate OAuth’s trust boundaries.
+- **Confused deputy problem**: If an unvalidated token is passed downstream, a downstream API may incorrectly trust it as though it were validated by the MCP Server.
+
+To maintain clear trust boundaries, MCP servers must only accept tokens explicitly issued for themselves and must act as independent OAuth clients when calling upstream services.
+Forwarding client tokens downstream is not allowed.
 
-Use the `--header` option when running the MCP Server to pass the header to the router. The default name of the header expected by the router is `apollographql-client-name`. To use a different header name, configure `telemetry.apollo.client_name_header` in router YAML configuration. 
+Our team is actively working on a robust authentication mechanism that follows OAuth 2.1 best practices and aligns with the MCP authorization model.
+Our goal is to deliver a secure, standards-compliant, and maintainable solution.
```
mabuyo: Curious about this part because we're essentially promising something in the docs. Are we confident about this? Would we be able to link to a GitHub issue or post where they can find out more about this or partner with us? We can put the link in "actively working on". Also suggest shortening to just the first line.

```suggestion
Our team is actively working on a robust authentication mechanism that follows OAuth 2.1 best practices and aligns with the MCP authorization model.
```

DaleSeo: @mabuyo Thanks for the quick review and feedback! We track our MCP work on Jira instead of using GitHub issues. This is [the link](https://apollographql.atlassian.net/browse/GT-204) to the epic, but it is not accessible from outside.

PR #160, opened 2025-06-26
docs: update mcp spec version

DaleSeo:
Just updating the limitations docs to reference the latest version of the MCP spec.


swcollard (approved):
👍

PR #161, opened 2025-06-26
Fix Rust 1.88 clippy lint warnings

pubmodmatt:


Jephuff (approved):

DaleSeo:
I'm using v1.86 here. How can we make sure we're all using the same version of Rust for the project?

DaleSeo:
Oh, found this:
https://github.com/apollographql/apollo-mcp-server/blob/297bc1712af755c186989046f7bcba145f1f9621/rust-toolchain.toml#L2

I guess "stale" refers to something that's always changing? 🙄

pubmodmatt:
> I'm using v1.86 here. How can we make sure we're all using the same version of Rust for the project?

You can set a specific Rust version in the `rust-toolchain.toml` file, but then you have to periodically update it. It's set to `stable` for now until there's a reason to do otherwise, so that just means we're always picking up the latest stable release. On your local machine, `rustup update` will pull latest.

PR #162, opened 2025-06-26
feat!: replace args with config file

nicholascioli:
This commit replaces all cli arguments with a single config file to simplify the deployment process and make future iterations on config easier to sync across projects using the MCP server.

BREAKING: Users now need to create a config file instead of using CLI arguments.


nicholascioli:
I still need to add doc changes for this, but I'm not sure if they'll get deployed immediately if so. Should that be a separate PR?

nicholascioli:
Some notable differences from the previous cli argument approach:

- Parse, don't validate: Several arguments were untyped strings that were later validated, which polluted a lot of main with fallible operations. Many of these were changed to be the typed representation so that `serde` can serve as the source of truth and unify fallible state.
- No more working directory: The config file makes it easier to have relative and absolute paths, so functionality around changing the current working directory was removed. It can be added back easily, so let me know if this shouldn't be the case.

crates/apollo-mcp-server/src/runtime/overrides.rs
```diff
@@ -0,0 +1,22 @@
...
+#[derive(Debug, Deserialize, Default, JsonSchema)]
+pub struct Overrides {
+    /// Disable type descriptions to save on context-window space
+    pub disable_type_description: bool,
+
+    /// Disable schema descriptions to save on context-window space
+    pub disable_schema_description: bool,
+
+    /// Expose the schema to the MCP client through `introspect` and `execute` tools
+    pub enable_introspection: bool,
```
pubmodmatt: I don't think these things make sense under overrides. Introspection is one of the two main ways to run the server, so I'd probably have an `introspection:` item, with separate items for `execute` and `introspect`, and each of those having a boolean `enabled` to start. There may be other options here in the future.

crates/apollo-mcp-server/src/runtime/overrides.rs
```diff
@@ -0,0 +1,22 @@
...
+use serde::Deserialize;
+
+/// Overridable flags
+#[derive(Debug, Deserialize, Default, JsonSchema)]
+pub struct Overrides {
+    /// Disable type descriptions to save on context-window space
+    pub disable_type_description: bool,
+
+    /// Disable schema descriptions to save on context-window space
+    pub disable_schema_description: bool,
```
pubmodmatt, line 13: These need to be optional - currently, if `orverrides:` is specified, *all* of the overrides must be set.

crates/apollo-mcp-server/src/runtime/config.rs
```diff
@@ -0,0 +1,135 @@
...
+pub struct Config {
+    /// Apollo-specific credential overrides
+    #[serde(default)]
+    pub graphos: GraphOSCredentials,
+
+    /// Path to a custom scalar map
+    pub custom_scalars: Option<PathBuf>,
+
+    /// The target GraphQL endpoint
+    pub endpoint: Url,
```
pubmodmatt, line 22: This should be optional and default to http://127.0.0.1:4000

crates/apollo-mcp-server/src/runtime/config.rs
```diff
@@ -0,0 +1,135 @@
...
+
+    /// List of hard-coded headers to include in all GraphQL requests
+    #[serde(default, deserialize_with = "parsers::map_from_str")]
+    #[schemars(schema_with = "super::schemas::header_map")]
+    pub headers: HeaderMap,
+
+    /// Operations
+    pub operations: OperationSource,
```
pubmodmatt, line 43: This should be optional - operations are not required if introspection is enabled, and when introspection is not enabled, should default to the default operation collection for the graph variant if APOLLO_GRAPH_REF is set.

crates/apollo-mcp-server/src/runtime/config.rs
```diff
@@ -0,0 +1,135 @@
...
+    /// The log level to use for tracing
+    #[serde(
+        default = "defaults::log_level",
+        deserialize_with = "parsers::from_str"
+    )]
+    #[schemars(schema_with = "super::schemas::level")]
+    pub log_level: Level,
+
+    /// The schema to load for operations
+    pub schema: SchemaSource,
```
pubmodmatt, line 51: Probably also want this to default to Upink if APOLLO_GRAPH_REF is set

crates/apollo-mcp-server/src/runtime/graphos.rs
```diff
@@ -0,0 +1,68 @@
...
+use schemars::JsonSchema;
+use serde::Deserialize;
+use url::Url;
+
+/// Credentials to use with GraphOS
+#[derive(Debug, Deserialize, Default, JsonSchema)]
+pub struct GraphOSCredentials {
+    /// The apollo key
+    #[schemars(with = "Option<String>")]
+    pub apollo_key: Option<SecretString>,
```
pubmodmatt: If this isn't set, it incorrectly gives an error that the APOLLO_KEY environment variable isn't set

crates/apollo-mcp-server/src/runtime/graphos.rs
```diff
@@ -0,0 +1,68 @@
...
+
+    /// The graph reference
+    pub apollo_graph_ref: Option<String>,
+
+    /// The URL to use for Apollo's registry
+    pub apollo_registry_url: Option<Url>,
+
+    /// List of uplink URL overrides
+    pub uplink_endpoints: Vec<Url>,
```
pubmodmatt: This is not optional, which would require customers to specify these internal URLs

crates/apollo-mcp-server/src/runtime/graphos.rs
```diff
@@ -0,0 +1,68 @@
...
+pub struct GraphOSCredentials {
+    /// The apollo key
+    #[schemars(with = "Option<String>")]
+    pub apollo_key: Option<SecretString>,
+
+    /// The graph reference
+    pub apollo_graph_ref: Option<String>,
+
+    /// The URL to use for Apollo's registry
+    pub apollo_registry_url: Option<Url>,
```
pubmodmatt: These aren't really credentials

pubmodmatt:
Add tests including minimal possible config file

pubmodmatt:
Add changelog entry including config file example

pubmodmatt (changes requested):

crates/apollo-mcp-server/src/runtime/overrides.rs
nicholascioli: Sounds good! The current main groups both of these features under the same introspection flag, so should I propagate these new flags through? Also, should these default to `true`?

crates/apollo-mcp-server/src/runtime/overrides.rs
nicholascioli: Whoops, good catch! I meant to mark these as `#[serde(default)]`. I'll add it to the other optional fields as well.

crates/apollo-mcp-server/src/runtime/config.rs
nicholascioli: Sounds good!

crates/apollo-mcp-server/src/runtime/config.rs
nicholascioli: Do we really want this level of indirection here when configuring? Operation source has a variant for uplink which is meant to explicitly outline that it requires valid uplink configuration to the end user. I can make that the default case, but let me know what you prefer.

crates/apollo-mcp-server/src/runtime/config.rs
nicholascioli: See comment  for `OperationSource` default.

crates/apollo-mcp-server/src/runtime/graphos.rs
nicholascioli: Hmm, this shouldn't be the case when not using uplink-related options. Let me see if I missed a case of this somewhere.

crates/apollo-mcp-server/src/runtime/graphos.rs
nicholascioli: I'll mark this as `#[serde(default)]` to zero initialize it if not specified.

crates/apollo-mcp-server/src/runtime/graphos.rs
nicholascioli: Whoops, I had meant to rename the struct to just `GraphOSConfig`. Thanks for catching this!

crates/apollo-mcp-server/src/runtime/overrides.rs
pubmodmatt: The introspection and execute tools are currently enabled by the same `--introspection` flag, but we've wanted to split that up. All of these booleans should default to `false`. We want the introspection tools turned off by default, and the things that are truly overrides (`disable_*`) should be `false` and to override it you'd set it to `true`.

crates/apollo-mcp-server/src/runtime/config.rs
pubmodmatt: We need to think about what we want the experience to be if the user doesn't specify any config, and just goes with default. Right now I believe that's getting the schema from uplink and the operations from the default operation collection, but those things only work if the APOLLO_GRAPH_REF is set. Generally it's good to be explicit rather than making assumptions, but the counterpoint to that is the getting started experience, where the user doesn't know what any of this config is and we just need to do something sensible.

crates/apollo-mcp-server/src/runtime/graphos.rs
nicholascioli: Whoops, turns out I forgot to actually extract these from the env. Fixed.

crates/apollo-mcp-server/src/runtime/config.rs
nicholascioli: Got it! Fixed.

crates/apollo-mcp-server/src/runtime/config.rs
nicholascioli: Fixed

crates/apollo-mcp-server/src/runtime/overrides.rs
nicholascioli: Done.

pubmodmatt:
I was just [reading this](https://www.anthropic.com/engineering/desktop-extensions), and it may change our strategy here. The templating can use command line arguments and environment variables, but wouldn't work with a config file.

.changesets/breaking_nc_config_file.md
```diff
@@ -0,0 +1,45 @@
+### Replace CLI flags with a configuration file - @nicholascioli PR #162
+
+All command line arguments are now removed and replaced with equivalent configuration
+options. The Apollo MCP server only accepts a single argument which is a path to a
+configuration file. An empty file may be passed, as all options have sane defaults
+that follow the previous argument defaults.
+
+Below is a valid configuration file with all options filled out:
```
swcollard: Out of scope for this change, but do we have a way to generate documentation about the config file schema from the data classes we serialize the info to?

crates/apollo-mcp-server/src/runtime/graphos.rs
```diff
@@ -0,0 +1,234 @@
...
+            .clone()
+            .or(std::env::var(APOLLO_GRAPH_REF_ENV).ok())
+            .ok_or_else(|| ServerError::EnvironmentVariable(APOLLO_GRAPH_REF_ENV.to_string()))
+    }
+
+    /// Extract the apollo key from the config or from the current env
+    fn key(&self) -> Result<SecretString, ServerError> {
+        self.apollo_key
+            .clone()
+            .or(std::env::var(APOLLO_KEY_ENV).map(Into::into).ok())
```
swcollard: In this case, we're only using the environment variable if the config field is not being set? Which I guess matches the existing parameter behavior. Or should the environment should be an override? I'm trying to think what is the more secure way to pass this around

swcollard (commented):
This looks really good to me. Thanks for taking on such a big refactor. Left a couple questions inline

swcollard:
Will we also need updates in the apollo runtime container to pass along a config file to the running mcp server?

.changesets/breaking_nc_config_file.md
nicholascioli: The config file derives `JsonSchema` which allows for generating a JSON documentation of all of the options and their descriptions. We should probably add a command for outputting this spec at some point.

crates/apollo-mcp-server/src/runtime/graphos.rs
nicholascioli: These are very good questions. The environment variable should probably be an override, so I'll refactor this to show that. I'll probably need to allow for overriding other options as well, so I'll look into it.

nicholascioli:
The latest changes add support for reading all options (preferentially) from environment variables with the prefix `APOLLO_MCP_`. Nested options can be specified by separating the field names by `__` (e.g. The yaml option `introspection.execute.enabled` can be specified by `APOLLO_MCP_INTROSPECTION__EXECUTE__ENABLED`)

Cargo.toml
```diff
@@ -21,8 +21,10 @@ reqwest = { version = "0.12.15", default-features = false, features = [
   "json",
   "native-tls-vendored",
 ] }
+secrecy = { version = "0.10.3", features = ["serde"] }
 serde = { version = "1.0.219", features = ["derive"] }
 serde_json = "1.0.140"
+serde_yaml = "0.9.34"
```
pubmodmatt: Do we actually end up needing this deprecated crate? It looks like all YAML is through figment instead?

crates/apollo-mcp-server/src/runtime/operation_source.rs
```diff
@@ -0,0 +1,104 @@
...
+    Manifest { path: PathBuf },
+
+    /// Load operations from uplink manifest
+    Uplink,
+}
+
+/// Either a custom ID or the default variant
+#[derive(Debug, PartialEq, Eq)]
+pub enum IdOrDefault {
+    /// The ddefault tools for the variant (requires APOLLO_KEY)
```
pubmodmatt:
```suggestion
    /// The default tools for the variant (requires APOLLO_KEY)
```

crates/apollo-mcp-server/src/runtime/operation_source.rs
```diff
@@ -0,0 +1,104 @@
+use std::path::PathBuf;
+
+use schemars::JsonSchema;
+use serde::Deserialize;
+
+/// Source for loaded operations
+#[derive(Debug, Default, Deserialize, JsonSchema)]
+#[serde(tag = "type", rename_all = "snake_case")]
```
pubmodmatt: "type" isn't really clear in the YAML file - maybe "source" would be better?
```suggestion
#[serde(tag = "source", rename_all = "snake_case")]
```

crates/apollo-mcp-server/src/runtime/schema_source.rs
```diff
@@ -0,0 +1,16 @@
+use std::path::PathBuf;
+
+use schemars::JsonSchema;
+use serde::Deserialize;
+
+/// Source for upstream GraphQL schema
+#[derive(Debug, Default, Deserialize, JsonSchema)]
+#[serde(tag = "type", rename_all = "snake_case")]
```
pubmodmatt:
```suggestion
#[serde(tag = "source", rename_all = "snake_case")]
```

.changesets/breaking_nc_config_file.md
```diff
@@ -0,0 +1,45 @@
...
+```yaml
+custom_scalars: /path/to/custom/scalars
+endpoint: http://127.0.0.1:4000
+graphos:
+  apollo_key: some.key
+  apollo_graph_ref: example@graph
+  apollo_registry_url: https://some.url
+  apollo_uplink_endpoints:
+    - http://uplink.endpoint.1
+    - http://uplink.endpoint.2
```
pubmodmatt: These are sort of internal, and having them in the example may confuse users because they won't know what to put here.

.changesets/breaking_nc_config_file.md
```diff
@@ -0,0 +1,45 @@
...
+overrides:
+  disable_type_description: false
+  disable_schema_description: false
+  enable_explorer: false
+  mutation_mode: all
+schema:
+  type: local
+  path: /path/to/schema.graphql
+transport:
+  type: sse
```
pubmodmatt: Maybe change the example to HTTP since SSE is deprecated

crates/apollo-mcp-server/src/runtime/config.rs
```diff
@@ -0,0 +1,189 @@
...
+    #[serde(default)]
+    pub introspection: Introspection,
+
+    /// The log level to use for tracing
+    #[serde(
+        default = "defaults::log_level",
+        deserialize_with = "parsers::from_str"
+    )]
+    #[schemars(schema_with = "super::schemas::level")]
+    pub log_level: Level,
```
pubmodmatt: Looking to the future, we're going to need to support things like logging to a file. So this may merit a logging section in the YAML.

Cargo.toml
nicholascioli: Whoops, I'll remove this!

crates/apollo-mcp-server/src/runtime/config.rs
nicholascioli: Done!

.changesets/breaking_nc_config_file.md
nicholascioli: Done!

.changesets/breaking_nc_config_file.md
nicholascioli: Removed!

crates/apollo-mcp-server/src/runtime/schema_source.rs
nicholascioli: Done!

crates/apollo-mcp-server/src/runtime/operation_source.rs
nicholascioli: Done!

crates/apollo-mcp-server/src/runtime/operation_source.rs
nicholascioli: Done!

PR #164, opened 2025-06-27
Allow argument description overrides

alocay:
Extracting operation comments for input variables and allowing these to override schema descriptions.


crates/apollo-mcp-server/src/operations.rs
```diff
@@ -433,6 +433,52 @@ pub fn operation_defs(
...
+    if start <= end {
+        Some(string[start..end].trim().to_string())
+    } else {
+        None
+    }
+}
+
+pub fn variable_description_overrides(
+    source_text: &str,
+) -> Result<Option<HashMap<String, String>>, OperationError> {
```
Jephuff: We can keep this a separate function and avoid reparsing the document. `operation_defs` returns the operation  so we just need to accept it here and then we can also remove the Result and Option from the return 👍 
```suggestion
    source_text: &str,
    operation_definition: &Node<OperationDefinition>,
) -> HashMap<String, String> {
```

crates/apollo-mcp-server/src/operations.rs
```diff
@@ -433,6 +433,52 @@ pub fn operation_defs(
...
+    let document = Parser::new()
+        .parse_ast(source_text, "operation.graphql")
+        .map_err(|e| OperationError::GraphQLDocument(Box::new(e)))?;
+    let mut argument_overrides_map: HashMap<String, String> = HashMap::new();
+    let mut last_offset: Option<usize> = Some(0);
```
Jephuff: We need to initialize this with the start of the variable definitions
```
query Something( $var
                ^ we need this to be the last_offset start value
```
We already error if the operation doesn't have a name, so the name is going to be the previous location we can use it to get an offset and then look for the opening `(` to get the offset

crates/apollo-mcp-server/src/operations.rs
```diff
@@ -433,6 +433,52 @@ pub fn operation_defs(
...
+                if let Some(description) = description.filter(|d| !d.is_empty()) {
+                    if let Some(description) = extract_overridden_variable_description(description)
+                    {
+                        argument_overrides_map
+                            .insert(v.name.to_string(), description.trim().to_string());
+                    }
+                }
+
+                last_offset = Some(source_span.end_offset());
+            };
```
Jephuff: if `v.location()` is `None` we need to set `last_offset` to `None` because it means we can't do the slice for the next variable
```suggestion
            } else {
              last_offset = None;
            }
```

crates/apollo-mcp-server/src/operations.rs
```diff
@@ -433,6 +433,52 @@ pub fn operation_defs(
     Ok(Some((document, operation, comments.map(|c| c.to_string()))))
 }
 
+pub fn extract_overridden_variable_description(string: &str) -> Option<String> {
```
Jephuff: We should copy what the operation override description formatting is, we can move it to this helper and use it in both spots. 

If you fix the start offset, this special casing for the first one won't been needed anymore and we also need to support multiple line comments like 

```
# This variable is for 
# - Accounts
# - Graphs
```

crates/apollo-mcp-server/src/operations.rs
alocay: Oh of course, good idea.

crates/apollo-mcp-server/src/operations.rs
alocay: Ah okay. Technically with it at 0 we can just ignore the everything up to the last instance of the `#` symbol and it'll ignore the name, the operation comment etc but I do prefer your suggestion.

Jephuff: So that only works if we only want to support single line comments, but we need to be able to have multiline ones like 
```
query Something(
  # Line one
  # Line two
  $var: ID
) {...}
```

crates/apollo-mcp-server/src/operations.rs
```diff
@@ -433,6 +433,67 @@ pub fn operation_defs(
...
+        None
+    }
+}
+
+pub fn variable_description_overrides(
+    source_text: &str,
+    operation_definition: &Node<OperationDefinition>,
+) -> HashMap<String, String> {
+    let mut argument_overrides_map: HashMap<String, String> = HashMap::new();
+    let mut last_offset: Option<usize> = Some(0);
```
Jephuff: Doing the special case inside the function is going to be a little to fragile, it would break if there was a `(` in a comment before the operation, or if there was a field with arguments in a fragment before the operation. 
We should initialize it here with something like 
```suggestion 
    let mut last_offset = operation_definition
        .name
        .and_then(|name| name.location())
        .and_then(|source_span| {
            source_text[source_span.end_offset()..]
                .find('(')
                .map(|pos| 1 + pos + source_span.end_offset())
        })
```

crates/apollo-mcp-server/src/operations.rs
```diff
@@ -433,6 +433,67 @@ pub fn operation_defs(
     Ok(Some((document, operation, comments.map(|c| c.to_string()))))
 }
 
+pub fn extract_and_format_overridden_variable_description(string: &str) -> Option<String> {
```
Jephuff: I think we can reference the operation description cleanup in `fn tool_description(`
So this function would essentially become this
```
comments.and_then(|comments| {
            let content = Regex::new(r"(\n|^)\s*#")
                .ok()?
                .replace_all(comments.as_str(), "$1");
            let trimmed = content.trim();

            if trimmed.is_empty() {
                None
            } else {
                Some(trimmed.to_string())
            }
        })
```
And be used in both spots.

crates/apollo-mcp-server/src/operations.rs
alocay: 🤯 I did not know you could get the location of the name so easily... that changes everything.

Jephuff: yeeah, it's pretty nice you can get the location of anything thats in the apollo rs stuff 🎉

crates/apollo-mcp-server/src/operations.rs
```diff
@@ -433,6 +433,67 @@ pub fn operation_defs(
...
+) -> HashMap<String, String> {
+    let mut argument_overrides_map: HashMap<String, String> = HashMap::new();
+    let mut last_offset: Option<usize> = Some(0);
+    operation_definition
+        .variables
+        .iter()
+        .for_each(|v| match v.location() {
+            Some(source_span) => {
+                let mut substring = last_offset
+                    .map(|start_offset| &source_text[start_offset..source_span.end_offset()]);
```
Jephuff: Right here, it should be `offset` to slice to the start instead of the end
```suggestion
                    .map(|start_offset| &source_text[start_offset..source_span.offset()]);
```

crates/apollo-mcp-server/src/operations.rs
```diff
@@ -433,6 +433,61 @@ pub fn operation_defs(
...
+}
+
+pub fn variable_description_overrides(
+    source_text: &str,
+    operation_definition: &Node<OperationDefinition>,
+) -> HashMap<String, String> {
+    let mut argument_overrides_map: HashMap<String, String> = HashMap::new();
+    let mut last_offset = operation_definition
+        .name
+        .clone()
```
Jephuff: Does `as_ref` work here? 👀 clone is a-ok if as_ref doesn't work! 
```suggestion
        .as_ref()
```

crates/apollo-mcp-server/src/operations.rs
```diff
@@ -433,6 +433,61 @@ pub fn operation_defs(
...
+        });
+    operation_definition
+        .variables
+        .iter()
+        .for_each(|v| match v.location() {
+            Some(source_span) => {
+                let comment = last_offset
+                    .map(|start_offset| &source_text[start_offset..source_span.offset()]);
+
+                if let Some(description) = comment.filter(|d| !d.is_empty() && d.contains('#')) {
```
Jephuff, line 487: Do we need to have this filter? `description` should either be an whitespace, none, or a comment I think?  Maybe we can pass `comment` directly to `extract_and_format_comments`?

crates/apollo-mcp-server/src/operations.rs
```diff
@@ -433,6 +433,61 @@ pub fn operation_defs(
...
+    operation_definition: &Node<OperationDefinition>,
+) -> HashMap<String, String> {
+    let mut argument_overrides_map: HashMap<String, String> = HashMap::new();
+    let mut last_offset = operation_definition
+        .name
+        .clone()
+        .and_then(|n| n.location())
+        .and_then(|span| {
+            source_text[span.end_offset()..]
+                .find('(')
```
Jephuff: I realized an edge case here, this will break if theres a comment before the variables that includes a (, which is totally valid
```
query MyQuery # This is a query (not a hat)
($arg: ID) {
  __typename
}
```

I think we can fix it with ✨ regular expressions ✨  but I'll leave it up to you if you want to fix that in this PR or the next one! at a minimum can we add a test that shows how it's failing?

Jephuff: can we also add a test that doesn't have variables but does have a field with arguments? I'm not sure if we haven't any tests like that yet, I think that should be fine because the result of this is ignore, but just to make sure!

```
query Something {
  field(id: "123") 
}
```

Jephuff (commented):
looking great! left a couple requests for some edge case tests :)

crates/apollo-mcp-server/src/operations.rs
alocay: That works, I'll change it.

crates/apollo-mcp-server/src/operations.rs
alocay: The reason I have the check for contains `#` is when there is no comment, the substring from line 470 returns just the delimiter comma (on the variables after the first) which then the code tries to do all the regex, formatting, etc. If there is no comment in the substring I just ignore it and continue. The `is_empty()` check can probably be removed.

Jephuff: Ahh I see, That sounds like it might be an issue with the end offset in general? If I understand correctly, will it include the `,` in this example
```
query Test(
  $arg: ID,
  # This is a description
  $arg2: ID,
) {...}
```
So would this description in that case be `,\n  This is a description`?

alocay: So in that case yes it would pick up `This is a description\n` but what I was referring to was a schema with no descriptions like so:
```gql
query Test($arg: ID, $arg2: ID) { ... }
```

The way it's currently done with the offsets, the first variable comment sliced out would be the empty string and the second variable comment sliced would be `,` (the comma at the end of variable one) since it's before variable 2's location and after variable 1. To keep from passing this string to the comment processing I just make sure first that the sliced string contains the `#` comment symbol.

crates/apollo-mcp-server/src/operations.rs
alocay: I have solved this with the regex `(?m)^\s*\(`, I'll put up an updated PR for this soon. The regex will look for the first line that starts with 0-1+ whitespaces and an opening parens. It runs on multiple lines and we run this on the substring pulled from `source_text[span.end_offset()..]` where the span is the `SourceSpan` from the op name location. Making the last_offset initialized with this solves the issue above. I'll also add some unit tests for these cases.

crates/apollo-mcp-server/src/operations.rs
Jephuff: Right, I think the problem is it's always going to pick up the `,`
when theres no comment we can ignore it, but if there is a comment we are going to pick it up as part of that string. 
So I think we need to handle the position and get the string after any `,`

If you add a test for that operation I think you will see what I mean. Things get a little weird here too haha
```
# Override for operation description
query 
# ignored comment
Test
# ignored comment
(
  # Override for arg description
  $arg: ID 
  # ignored comment
  ,
    # Override for arg2 description
  $arg2: ID
) {...}
```
I think we want to do something like split on `\s*,` and take the second half? might be too simplistic...

alocay: Just pushed a commit that I believe deals with this comma

nicholascioli:
Does the parsed schema from the `apollo_compiler` not have a way to extract comments from various parts of the AST? I'm uninformed here, but I just wanted to make sure that we've considered traversing the typed AST versus running magic regular expressions on raw strings.

crates/apollo-mcp-server/src/operations.rs
```diff
@@ -434,6 +434,74 @@ pub fn operation_defs(
     Ok(Some((document, operation, comments.map(|c| c.to_string()))))
 }
 
+pub fn extract_and_format_comments(comments: Option<String>) -> Option<String> {
+    comments.and_then(|comments| {
+        let content = Regex::new(r"(\n|^)\s*,*\s*#")
```
Jephuff: So the edge case I was foreshadowing in the meeting earlier you are already half handling here!
`,` are whitespace, so you can have `,,,,,,,` between arguments, which this will handle, but you also could have `,,  , , , , , ` with mixed spaces and commas. Not sure we really need to worry about people doing that, because... why hahah
but we could simplify this regex to this to consume all the commas before the first comment.
```suggestion
        let content = Regex::new(r"(\n|^)[\s,]*#")
```

The one thing this won't handle is a comment between the commas, but I don't think we need to worry about that!
```
query Name($arg: ID,
# comment
,
# real comment
$arg2: ID
)
```

alocay:
> Does the parsed schema from the `apollo_compiler` not have a way to extract comments from various parts of the AST? I'm uninformed here, but I just wanted to make sure that we've considered traversing the typed AST versus running magic regular expressions on raw strings.

As far as I'm aware the apollo compiler does not provide a way to get comments from the operation. Ofc this changes once the support for descriptions in the operation is merged in.

crates/apollo-mcp-server/src/operations.rs
Jephuff: ooh actually we can handle that case too, I think making the # optional would clear out those lines? 
```suggestion
        let content = Regex::new(r"(\n|^)[\s,]*#?")
```

crates/apollo-mcp-server/src/operations.rs
```diff
@@ -434,6 +434,74 @@ pub fn operation_defs(
...
+            .ok()?
+            .replace_all(comments.as_str(), "$1");
```
Jephuff, line 441: I don't this we need to keep the leading \n here, "" should be fine? 
```suggestion
            .replace_all(comments.as_str(), "");
```

crates/apollo-mcp-server/src/operations.rs
Jephuff: With your change and I think if you accept this change https://github.com/apollographql/apollo-mcp-server/pull/164/files#r2183443166
 These checks can go away now?

crates/apollo-mcp-server/src/operations.rs
alocay: This is taken directly from what is already being done, I just moved it so it can be used for both operation comments and variable comments. Wasn't sure if we want to change this too.

crates/apollo-mcp-server/src/operations.rs
alocay: That regex does not work 100% for me. I had to adjust it to `(\n|^)(\s*,*)*#` for it to pass a test case of mine.

crates/apollo-mcp-server/src/operations.rs
Jephuff: My comment didn't make sense, we want it still. Sorry for the distraction! haha

Jephuff:
> As far as I'm aware the apollo compiler does not provide a way to get comments from the operation. Ofc this changes once the support for descriptions in the operation is merged in.  

@alocay @nicholascioli 
Thats correct, apollo-rs doesn't give us a way to get comments from the AST, so the only way is to extract ranges of the space between tokens. 
Operations also don't currently support descriptions, but that is coming soon so that will replace this once graphql-js and apollo-rs support it fully 👍

PR #166, opened 2025-06-30
Include Operation Name with graphql requests

DaleSeo:
This PR updates the code to include the operation name with GraphQL requests if it's available.

```diff
{
   "query":"query GetAlerts($state: String!) { alerts(state: $state) { severity description instruction } }",
   "variables":{
      "state":"CO"
   },
   "extensions":{
      "ApolloClientMetadata":{
         "type":"mcp",
         "version":"0.4.2"
      }
   },
+  "operationName":"GetAlerts"
}
```

### DIFF
--- a/crates/apollo-mcp-server/src/graphql.rs
+++ b/crates/apollo-mcp-server/src/graphql.rs
@@ -11,22 +11,25 @@ pub struct Request<'a> {
     pub headers: HeaderMap,
 }
 
+#[derive(Debug, PartialEq)]
+pub struct OperationDetails {
+    pub query: String,
+    pub operation_name: Option<String>,
+}
+
 /// Able to be executed as a GraphQL operation
 pub trait Executable {
     /// Get the persisted query ID to be executed, if any
     fn persisted_query_id(&self) -> Option<String>;
 
-    /// Get the operation to execute
-    fn operation(&self, input: Value) -> Result<String, McpError>;
+    /// Get the operation to execute and its name
+    fn operation(&self, input: Value) -> Result<OperationDetails, McpError>;
 
     /// Get the variables to execute the operation with
     fn variables(&self, input: Value) -> Result<Value, McpError>;
 
     /// Get the headers to execute the operation with
     fn headers(&self, default_headers: &HeaderMap<HeaderValue>) -> HeaderMap<HeaderValue>;
-
-    /// Get the operation name
-    fn operation_name(&self) -> Option<String>;
 
     /// Execute as a GraphQL operation using the endpoint and headers
     async fn execute(&self, request: Request<'_>) -> Result<CallToolResult, McpError> {
@@ -52,20 +55,22 @@ pub trait Executable {
                 }),
             );
         } else {
-            request_body.insert(
-                String::from("query"),
-                Value::String(self.operation(request.input)?),
-            );
+            let OperationDetails {
+                query,
+                operation_name,
+            } = self.operation(request.input)?;
+
+            request_body.insert(String::from("query"), Value::String(query));
             request_body.insert(
                 String::from("extensions"),
                 serde_json::json!({
                     "ApolloClientMetadata": client_metadata,
                 }),
             );
-        }
 
-        if let Some(op_name) = self.operation_name() {
-            request_body.insert("operationName".to_string(), Value::String(op_name));
+            if let Some(op_name) = operation_name {
+                request_body.insert(String::from("operationName"), Value::String(op_name));
+            }
         }
 
         reqwest::Client::new()
--- a/crates/apollo-mcp-server/src/graphql.rs
+++ b/crates/apollo-mcp-server/src/graphql.rs
@@ -1,10 +1,9 @@
 //! Execute GraphQL operations from an MCP tool
 
 use crate::errors::McpError;
-use apollo_compiler::response::serde_json_bytes::serde_json;
-use apollo_compiler::response::serde_json_bytes::serde_json::Value;
 use reqwest::header::{HeaderMap, HeaderValue};
 use rmcp::model::{CallToolResult, Content, ErrorCode};
+use rmcp::serde_json::{self, Map, Value};
 
 pub struct Request<'a> {
     pub input: Value,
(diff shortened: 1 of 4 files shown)


crates/apollo-mcp-server/src/graphql.rs
```diff
@@ -26,37 +26,50 @@ pub trait Executable {
...
+        };
+
+        if let Some(op_name) = self.operation_name() {
+            if let Some(obj) = request_body.as_object_mut() {
+                obj.insert(
+                    "operationName".to_string(),
+                    serde_json::Value::String(op_name),
+                );
+            }
+        }
```
Jephuff: I'm pretty sure serde will omit `None` keys from the output so I thiiink we can inline these to match the other values?
```suggestion
                },
                "operationName": self.operation_name()
            })
        } else {
            serde_json::json!({
                "query": self.operation(request.input.clone())?,
                "variables": self.variables(request.input)?,
                "extensions": {
                    "ApolloClientMetadata": client_metadata,
                },
                "operationName": self.operation_name()
            })
        };
```

crates/apollo-mcp-server/src/operations.rs
```diff
@@ -1019,6 +1019,21 @@ impl graphql::Executable for Operation {
...
+                for definition in document.definitions {
+                    if let Definition::OperationDefinition(operation) = definition {
+                        return operation.name.as_ref().map(|name| name.to_string());
+                    }
+                }
+                None
+            }
+            Err(_) => None,
+        }
+    }
```
Jephuff: Operation::from_document already parsed out the name so we can add `operation_name` to the `Operation` struct and return that
```suggestion
       Some(self.operation_name)
```

crates/apollo-mcp-server/src/introspection/tools/execute.rs
```diff
@@ -79,6 +79,12 @@ impl graphql::Executable for Execute {
     fn headers(&self, default_headers: &HeaderMap<HeaderValue>) -> HeaderMap<HeaderValue> {
         default_headers.clone()
     }
+
+    fn operation_name(&self) -> Option<String> {
+        // For ad-hoc operations, we don't have access to the input here,
+        // so we can't extract the operation name. This is acceptable for ad-hoc queries.
+        None
```
Jephuff: Hmmm, I kind of wonder if we should allow the LLM to run operations without names, or if we should make name the generated ones with a generic name like `LLMGeneratedOperation`.
Thats a separate ticket though! 

For this one, we do have the operation, we just need to get the name here. 
On line 54 above we parse the document and ignore the result, but the return value from `operation_defs` includes the operation_definition which has the name, so we can return that from the `operation` method. We might want to not add this method and just have `operation` return the name + operation?

crates/apollo-mcp-server/src/graphql.rs
DaleSeo: @Jephuff It doesn't seem to omit them. It becomes `"operationName":null`.

```json
{
   "query":"query($state: String!) { alerts(state: $state) { severity description instruction } }",
   "variables":{
      "state":"CO"
   },
   "extensions":{
      "ApolloClientMetadata":{
         "type":"mcp",
         "version":"0.4.2"
      }
   },
   "operationName":null
}
```

crates/apollo-mcp-server/src/operations.rs
DaleSeo: The operation name is parsed there but only stored in a variable. Are you suggesting we add the `operation_name` field to the `Operation` struct and include it when `from_document` returns so that it's available here?

crates/apollo-mcp-server/src/graphql.rs
Jephuff: dang! I think I was thinking about we mark up a struct to get rid of the `None`s :(

crates/apollo-mcp-server/src/operations.rs
Jephuff: yep!

crates/apollo-mcp-server/src/introspection/tools/execute.rs
DaleSeo: The `operation` method comes from the `Executable` trait.

https://github.com/apollographql/apollo-mcp-server/blob/aed1b6e8909e9d3d9b843b0843f9e99a8bd58601/crates/apollo-mcp-server/src/graphql.rs#L20-L21

So, I guess we need to change the method to return a tuple containing two strings? One string would represent the operation text itself, while the other would indicate the operation name?

Jephuff: I think thats probably the easiest version of this, yep!

crates/apollo-mcp-server/src/graphql.rs
Jephuff: Thinking about this some more, if we care going to break this up we can make a little cleaner. (or we could turn this into a struct) I think the `json!` macro makes sense if we aren't going to manipulate the data after, but the fact that we are having to check to make sure it's an object after makes this code a little weird.  So heres my new proposal! 

1. return `serde_json::Map::from_iter([])` from this if instead of json!
2. only include extensions + query (the things that are different) in the initial from_iter
3. insert variables after the if because they are the same, this removes the duplication
4. Fix the errors in the operation_name part to not if let Some(obj)
5. fix the stringify in the .body, something like `.body(serde_json::Value::Object(request_body).to_string())`

DaleSeo: Hey @Jephuff, I’ve refactored the code as suggested. I hope it looks cleaner now! 😃

crates/apollo-mcp-server/src/introspection/tools/execute.rs
DaleSeo: Hey @Jephuff, I removed the `operation_name` method from the `Executable` trait and modified the `operation` method so that its return value includes both the operation and its name. I tried using a tuple as the return type, but I found it cleaner to use a structure instead. So, I ended up adding `OperationDetails`.

https://github.com/apollographql/apollo-mcp-server/blob/d35cfa8a390589c70818ef27c8de840b46bf73e8/crates/apollo-mcp-server/src/graphql.rs#L15-L18

crates/apollo-mcp-server/src/introspection/tools/execute.rs
```diff
@@ -45,16 +48,22 @@ impl graphql::Executable for Execute {
...
-            .map_err(|e| McpError::new(ErrorCode::INVALID_PARAMS, e.to_string(), None))?;
-
-        Ok(input.query)
+        let operation_defs =
+            operation_defs(&input.query, self.mutation_mode == MutationMode::All, None)
+                .map_err(|e| McpError::new(ErrorCode::INVALID_PARAMS, e.to_string(), None))?;
+
+        Ok(OperationDetails {
+            query: input.query,
+            operation_name: operation_defs.and_then(|(_, operation_def, source_path)| {
```
Jephuff, line 63: Hmm, if `operation_defs` is `None` thats actually an error because it either means that the LLM tried to run a subscription or it tried to run a mutation when it shouldn't. 
We should probably return an invaid params error if `operation_defs` is None as well. (I know thats changing the original behaviour, so if you wanted to push that to another PR thats ok too! :) )

crates/apollo-mcp-server/src/operations.rs
```diff
@@ -642,7 +644,7 @@ impl Operation {
     }
 }
 
-fn operation_name(
+pub fn operation_name(
```
Jephuff, line 647: Does this still need to be pub?
```suggestion
fn operation_name(
```

DaleSeo: Yes, I changed it to be public so that I can reuse it `execute.rs`.

https://github.com/apollographql/apollo-mcp-server/blob/d5105003ed8b5134c7c01821aedaf0f7ca21113d/crates/apollo-mcp-server/src/introspection/tools/execute.rs#L64

Jephuff: Right! ignore me!

crates/apollo-mcp-server/src/introspection/tools/execute.rs
DaleSeo: OK, I’ll follow up on that in a separate PR.

PR #169, opened 2025-07-03
Fix client metadata extension name

pubmodmatt:
I had used a client metadata key from an outdated doc. The actual extension name [used by the router](https://github.com/apollographql/router/blob/413ec9c4b25ec03705237b4ceb1215d8a4b17c44/apollo-router/src/plugins/enhanced_client_awareness/mod.rs#L13) is `clientLibrary` and the key is `name` instead of `type`.

### DIFF
--- a/crates/apollo-mcp-server/src/graphql.rs
+++ b/crates/apollo-mcp-server/src/graphql.rs
@@ -34,7 +34,7 @@ pub trait Executable {
     /// Execute as a GraphQL operation using the endpoint and headers
     async fn execute(&self, request: Request<'_>) -> Result<CallToolResult, McpError> {
         let client_metadata = serde_json::json!({
-            "type": "mcp",
+            "name": "mcp",
             "version": std::env!("CARGO_PKG_VERSION")
         });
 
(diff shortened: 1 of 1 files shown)


nicholascioli:
This looks good and works for now, but is there any way that we could extract this structure information from the router as a library? The use of raw JSON spooks me since it can lead to these sort of untyped issues.

nicholascioli:
There are other parts of this codebase that have raw JSON literals that we should also break out into typed structures, so this isn't a problem with this feature in specific.

alocay (approved):

pubmodmatt:
> This looks good and works for now, but is there any way that we could extract this structure information from the router as a library? The use of raw JSON spooks me since it can lead to these sort of untyped issues.

Agree, but it isn't so straightforward since clients are in a variety of languages, like [swift](https://github.com/apollographql/apollo-ios/blob/d591c1dd55824867877cff4f6551fe32983d7f51/Sources/Apollo/RequestClientMetadata.swift#L25) and [kotlin](https://github.com/apollographql/apollo-kotlin/blob/65dd7a44a74ed016988777521248da164de32ea0/libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/http/DefaultHttpRequestComposer.kt#L180).

PR #171, opened 2025-07-03
Add search tool

pubmodmatt:
* Add a new `apollo-search-index` crate to index and search GraphQL schemas
* Add a new `search` tool to the MCP server that uses the new crate to search for type information. Enabled with a new `--search` option.
* Fix the schema tree shaker to not add a stub `Query` type

There is still work to be done to improve the new tool, but this is a start. For example, it currently includes all type information for types along each root path. It could just include the field that was traversed to get from parent to child type in the path, which would greatly reduce the returned size. That will be done as follow-on work.


crates/apollo-mcp-server/src/server/states/starting.rs
```diff
@@ -85,10 +80,23 @@ impl Starting {
...
+                Some(Execute::new(self.config.mutation_mode)),
+                Some(Introspect::new(
+                    schema.clone(),
+                    root_query_type,
+                    root_mutation_type,
+                )),
+                Some(Search::new(
+                    schema.clone(),
+                    matches!(self.config.mutation_mode, MutationMode::All),
+                )?),
```
DaleSeo: How can a client make the search tool preferred over the introspect tool?

crates/apollo-mcp-server/src/introspection/tools/search.rs
```diff
@@ -0,0 +1,207 @@
...
+        for root_path in root_paths {
+            let types = root_path.inner.types.clone();
+            let path_len = types.len();
+            for (i, type_name) in types.into_iter().enumerate() {
+                if let Some(extended_type) = schema.types.get(type_name.as_ref()) {
+                    let depth = if i == path_len - 1 {
+                        LEAF_DEPTH
+                    } else {
+                        DepthLimit::Limited(1)
+                    };
```
DaleSeo, line 109: I'm a bit confused here. Do both branches result in exactly the same value since `LEAF_DEPTH` is set to `DepthLimit::Limited(1)` on L24?

crates/apollo-mcp-server/src/introspection/tools/search.rs
```diff
@@ -0,0 +1,207 @@
...
+
+        assert!(!result.is_error.unwrap_or(false));
+        insta::assert_snapshot!(content_to_snapshot(result));
+    }
+
+    #[rstest]
+    #[tokio::test]
+    async fn test_referencing_types_are_collected(schema: Valid<Schema>) {
+        let schema = Arc::new(Mutex::new(schema));
+        let search = Search::new(schema.clone(), true).expect("Failed to create search tool");
```
DaleSeo, line 191: It doesn't matter if we pass `true` to `allow_mutations` in the test, right? Would it help to add tests to confirm that the search tool is able to find mutations?

crates/apollo-mcp-server/src/introspection/tools/introspect.rs
```diff
@@ -87,31 +86,12 @@ impl Introspect {
...
                             .root_operation(OperationType::Mutation)
                             .is_none_or(|root_name| {
                                 extended_type.name() != root_name
                                     || (type_name == root_name.as_str() && self.allow_mutations)
                             })
-                        && schema
-                            .root_operation(OperationType::Subscription)
-                            .is_none_or(|root_name| {
-                                extended_type.name() != root_name || type_name == root_name.as_str()
-                            })
```
DaleSeo: No longer filter out subscriptions?

crates/apollo-mcp-server/src/introspection/tools/introspect.rs
```diff
@@ -87,31 +86,12 @@ impl Introspect {
...
-                                | ExtendedType::Scalar(_)
-                                | ExtendedType::Enum(_)
-                                | ExtendedType::Interface(_)
-                                | ExtendedType::Union(_)
-                        )
-                        && schema
-                            .root_operation(OperationType::Query)
-                            .is_none_or(|root_name| {
-                                extended_type.name() != root_name || type_name == root_name.as_str()
-                            })
```
DaleSeo, line 103: Why is this removed?

DaleSeo (approved):
Reviewing 2,500+ lines in one sitting is a bit much. 😅 I'll pick up the new crate review on Monday.

crates/apollo-schema-index/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+license-file = "../LICENSE"
+repository = "https://github.com/apollographql/apollo-mcp-server"
+description = "GraphQL schema indexing"
+
+[dependencies]
+apollo-compiler.workspace = true
+enumset = "1.1.6"
+itertools = "0.14.0"
+tantivy = "0.24.1"
+thiserror = "2.0.12"
```
DaleSeo: Can we also inherit this package?

```suggestion
thiserror.workspace = true
```

https://github.com/apollographql/apollo-mcp-server/blob/7a395ca9491a0cf2ca37ffae320b2fd546ef7948/Cargo.toml#L31

crates/apollo-mcp-server/src/server/states/starting.rs
pubmodmatt: I'm open for input on the direction here. I'm planning on having the search tool just outright replace the introspect tool, but I haven't removed introspect yet here. Currently one or the other can be disabled on the client side (Claude Desktop lets you disable tools, for example), or it can be left to the LLM to decide which tool to use. Another option would be to have config that enables these separately (the current config PR #162 has the introspect tools broken out separately).

crates/apollo-mcp-server/src/introspection/tools/search.rs
pubmodmatt: I ended up not using this in this first version, but it's there for future changes. Providing more information about leaf nodes can be beneficial to the LLM, since it can see all the fields that it can query on the matching type. This is not used in this current implementation because it returns all fields for all types on the root path, which is already too much data. There is a TODO in the code (and I'll create a task) to reduce that data, and then we could consider increasing the leaf depth.

crates/apollo-schema-index/src/lib.rs
```diff
@@ -0,0 +1,527 @@
...
+use apollo_compiler::validation::Valid;
+use enumset::{EnumSet, EnumSetType};
+use error::{IndexingError, SearchError};
+use itertools::Itertools;
+use path::{RootPath, Scored};
+use std::collections::{HashMap, HashSet, VecDeque};
+use std::time::Instant;
+use tantivy::collector::TopDocs;
+use tantivy::query::{BooleanQuery, Occur, Query, TermQuery};
+use tantivy::schema::document::Value as TantivyValue;
```
DaleSeo: Where is it used?

crates/apollo-mcp-server/src/introspection/tools/search.rs
pubmodmatt: Sure, I can add that.

crates/apollo-mcp-server/src/introspection/tools/introspect.rs
pubmodmatt: I may have removed too much code here, I'll take a look.

crates/apollo-schema-index/src/lib.rs
```diff
@@ -0,0 +1,527 @@
...
+            // Create a document for each type
+            let mut doc = TantivyDocument::default();
+            doc.add_text(type_name_field, extended_type.name());
+            doc.add_text(raw_type_name_field, extended_type.name());
+            doc.add_text(
+                description_field,
+                extended_type
+                    .description()
+                    .map(|d| d.to_string())
+                    .unwrap_or(String::from("")),
```
DaleSeo: nit

```suggestion
                    .unwrap_or_default(),
```

crates/apollo-schema-index/src/lib.rs
```diff
@@ -0,0 +1,527 @@
...
+    type_name_field: Field,
+    description_field: Field,
+    fields_field: Field,
+    referencing_types_field: Field,
+}
+
+impl SchemaIndex {
+    pub fn new(
+        schema: &Valid<Schema>,
+        root_types: EnumSet<OperationType>,
```
DaleSeo, line 124: I'm just curious about the value that `EnumSet` offers compared to simply using `HashSet`.

crates/apollo-schema-index/src/lib.rs
```diff
@@ -0,0 +1,527 @@
...
+
+        // Create the index
+        let index_schema = index_schema.build();
+        let index = Index::create_in_ram(index_schema);
+        index
+            .tokenizers()
+            .register("en_stem", text_analyzer.clone());
+
+        // Map every type in the schema to the types referencing it
+        let mut index_writer = index.writer(INDEX_MEMORY_BYTES)?;
```
DaleSeo, line 173: Oh, do we need 50 MB of memory just for indexing?

crates/apollo-schema-index/src/lib.rs
```diff
@@ -0,0 +1,527 @@
...
+                .entry(extended_type.name().to_string())
+                .or_default();
+            if let Some(ref_type) = path.referencing_type() {
+                entry.push(ref_type.to_string());
+            }
+        }
+
+        if tracing::enabled!(Level::DEBUG) {
+            for (type_name, references) in &type_references {
+                debug!("Type '{}' is referenced by: {:?}", type_name, references);
```
DaleSeo, line 186: I've been calling `debug!()` without this if statement so far. Should we check the log level before calling it?

crates/apollo-schema-index/src/lib.rs
```diff
@@ -0,0 +1,527 @@
...
+                })
+                .map(|term| {
+                    (
+                        Occur::Should,
+                        Box::new(TermQuery::new(term, IndexRecordOption::Basic)) as Box<dyn Query>,
+                    )
+                })
+                .collect(),
+        );
+        query.set_minimum_number_should_match(1);
```
DaleSeo, line 487: Why is this set to `1`?

crates/apollo-schema-index/src/snapshots/apollo_schema_index__tests__search.snap
```diff
@@ -0,0 +1,10 @@
+---
+source: crates/apollo-schema-index/src/lib.rs
+expression: "results.iter().take(10).map(ToString::to_string).collect::<Vec<_>>().join(\"\\n\")"
+---
+Mutation -> Media -> MediaMetadata (6.1192875)
+Mutation -> Media -> MediaMetadata -> Dimensions (5.984665)
```
DaleSeo, line 6: Why does the first one receive a higher score when it doesn't contain the search term?

crates/apollo-mcp-server/src/server/states/starting.rs
DaleSeo: Thanks for clarifying, @pubmodmatt! My two cents: so far, we've mostly used CLI options to enable tools like `-x` for the explorer tool and `-i` for the inspect and execute tools. If I were a user, I would find it more consistent to set this up directly in the MCP server. I'm also a bit concerned that not all users will know how to disable tools on MCP clients, and documenting that for each client in our user guide might be challenging.

crates/apollo-mcp-server/src/introspection/tools/introspect.rs
pubmodmatt: The `matches` expression was nonsensical, as it matched every case. The remaining cases were about not including root types, but it always included a stub Query type - that was just wierd complexity. The tree shaker should be agnostic to this, and just add the types that were added by callers, including root types. Root paths, for example, need to start at the root. The stub Query has also been removed - the schema returned doesn't strictly need to be "valid" - we're just using it to communicate type information to the LLM.

crates/apollo-schema-index/src/lib.rs
pubmodmatt: It has special support for enums and lets you or them together, for example.

crates/apollo-schema-index/src/lib.rs
pubmodmatt: We should make this configurable. Our customers can have *very* large schemas.

crates/apollo-schema-index/src/lib.rs
pubmodmatt: The macro itself only emits if the log level is enabled, so you only need to check fi the code is doing some additional work, such as looping over all the references as it's doing here.

crates/apollo-schema-index/src/lib.rs
pubmodmatt: All the terms are "should" clauses, so they don't need to match. This says at least one needs to match.

crates/apollo-schema-index/src/snapshots/apollo_schema_index__tests__search.snap
pubmodmatt: It does match, as it has a field named "dimensions". It's also the shortest path to root.

DaleSeo: Oh, sorry! I just found that field in the schema. 😆

crates/apollo-schema-index/src/lib.rs
DaleSeo: Thanks, @pubmodmatt. Now I get that it's for performance.

crates/apollo-mcp-server/src/server/states/starting.rs
pubmodmatt: I've added a `--search` option to specifically enable this tool and the associated schema indexing.

PR #172, opened 2025-07-04
docs: clarify requirements for `--collection` option

DaleSeo:
I just ran into this using `--collection default`. 😆

```
error: the following required arguments were not provided:
  --apollo-graph-ref <APOLLO_GRAPH_REF>
```

We should clarify requirements for the `--collection` option in the command reference.

### Preview

Shot 2025-07-04 at 11 05 41


docs/source/command-reference.mdx
```diff
@@ -105,11 +105,11 @@ apollo-mcp-server [OPTIONS] --directory <DIRECTORY>
...
 | `-o, --operations [<OPERATIONS>...]`                  | Operation files to expose as MCP tools. [Learn more](/apollo-mcp-server/guides/#from-operation-files).                                                                                                                               |
 | `--manifest <MANIFEST>`                               | The path to the persisted query manifest containing operations.                                                                                                                                                                      |
-| `--collection <COLLECTION_ID>`                        | The ID of an operation collection to use as the source for operations                                                                                                                                                                |
+| `--collection <COLLECTION_ID>`                        | The ID of an operation collection to use as the source for operations (requires `APOLLO_KEY`). When set to `default`, also requires `APOLLO_GRAPH_REF`.                                                                              |
```
Jephuff: we don't really call out what default is here either, so maybe we could do something like 
```suggestion
| `--collection <COLLECTION_ID>`                        | The ID of an operation collection to use as the source for operations (requires `APOLLO_KEY`). If `default` is used instead of an ID, the MCP Tools for `APOLLO_GRAPH_REF` will be used as the source.                                                                              |
```

DaleSeo: Makes sense, @Jephuff!

PR #173, opened 2025-07-04
Skip unnamed operations and log a warning instead of crashing

DaleSeo:
image

### DIFF
--- a/crates/apollo-mcp-server/src/operations.rs
+++ b/crates/apollo-mcp-server/src/operations.rs
@@ -448,7 +448,18 @@ impl Operation {
             mutation_mode != MutationMode::None,
             raw_operation.source_path.clone(),
         )? {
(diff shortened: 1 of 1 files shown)


crates/apollo-mcp-server/src/operations.rs
```diff
@@ -448,7 +448,18 @@ impl Operation {
             mutation_mode != MutationMode::None,
             raw_operation.source_path.clone(),
         )? {
-            let operation_name = operation_name(&operation, raw_operation.source_path.clone())?;
+            let operation_name = match operation_name(&operation, raw_operation.source_path.clone())
+            {
+                Ok(name) => name,
+                Err(OperationError::MissingName {
+                    source_path: _,
```
swcollard: Nit, it would be good to log this path is if is present

DaleSeo: Sure @swcollard, I've included the path in the warning logs.

PR #175, opened 2025-07-07
Update to Rust MCP SDK version 2

pubmodmatt:
There are new [2.x versions of the Rust MCP SDK](https://crates.io/crates/rmcp/versions). Update to the latest 2.x version.

Previously, we had pinned an old version of the SDK due to shutdown problems. But I think that was a mistake. With the latest SDK, pressing Ctrl+C will not work if there are open connections. But I believe this is likely the intended behavior. If you close all open connections, then Ctrl+C will work. The Apollo MCP Server has the same behavior here as the MCP SDK examples.

Also note that `stateful_mode` seems to affect this behavior. If `false`, it would exit the same as it did with previous SDK versions. But I'm setting `stateful_mode` to `true` here, since we'd want the advantages of stateful mode in the server.

### DIFF
--- a/crates/apollo-mcp-server/src/server/states.rs
+++ b/crates/apollo-mcp-server/src/server/states.rs
@@ -1,7 +1,7 @@
 use apollo_compiler::{Schema, validation::Valid};
 use apollo_federation::{ApiSchemaOptions, Supergraph};
 use apollo_mcp_registry::uplink::schema::{SchemaState, event::Event as SchemaEvent};
-use futures::{FutureExt as _, Stream, StreamExt as _, future, stream};
+use futures::{FutureExt as _, Stream, StreamExt as _, stream};
 use reqwest::header::HeaderMap;
 
 use crate::{
@@ -144,34 +144,36 @@ impl StateMachine {
         }
     }
 
-    #[allow(clippy::expect_used)]
     fn ctrl_c_stream() -> impl Stream<Item = ServerEvent> {
-        #[cfg(not(unix))]
-        {
-            async {
-                tokio::signal::ctrl_c()
-                    .await
-                    .expect("Failed to install CTRL+C signal handler");
-            }
+        shutdown_signal()
             .map(|_| ServerEvent::Shutdown)
             .into_stream()
             .boxed()
-        }
+    }
+}
 
-        #[cfg(unix)]
-        future::select(
-            tokio::signal::ctrl_c().map(|s| s.ok()).boxed(),
-            async {
-                tokio::signal::unix::signal(tokio::signal::unix::SignalKind::terminate())
-                    .expect("Failed to install SIGTERM signal handler")
-                    .recv()
-                    .await
-            }
-            .boxed(),
-        )
-        .map(|_| ServerEvent::Shutdown)
-        .into_stream()
-        .boxed()
+#[allow(clippy::expect_used)]
+async fn shutdown_signal() {
+    let ctrl_c = async {
+        tokio::signal::ctrl_c()
+            .await
+            .expect("Failed to install CTRL+C signal handler");
+    };
+
+    #[cfg(unix)]
+    let terminate = async {
+        tokio::signal::unix::signal(tokio::signal::unix::SignalKind::terminate())
+            .expect("Failed to install SIGTERM signal handler")
+            .recv()
+            .await;
+    };
+
+    #[cfg(not(unix))]
+    let terminate = std::future::pending::<()>();
+
+    tokio::select! {
+        _ = ctrl_c => {},
+        _ = terminate => {},
     }
 }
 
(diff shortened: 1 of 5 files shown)


crates/apollo-mcp-server/src/server/states/starting.rs
```diff
@@ -115,14 +114,19 @@ impl Starting {
...
+                        stateful_mode: true,
+                    },
+                );
+                let router = axum::Router::new().nest_service("/mcp", service);
+                let tcp_listener = tokio::net::TcpListener::bind(listen_address).await?;
+                axum::serve(tcp_listener, router)
+                    .with_graceful_shutdown(shutdown_signal())
+                    .await?;
             }
             Transport::SSE { address, port } => {
```
DaleSeo, line 128: Looking at the example, we can similarly update the SSE transport code as we did for the Streamable HTTP transport: https://github.com/modelcontextprotocol/rust-sdk/blob/main/examples/servers/src/counter_sse.rs

crates/apollo-mcp-server/src/server/states/starting.rs
```diff
@@ -115,14 +114,19 @@ impl Starting {
...
-                    ct: cancellation_token,
-                    sse_keep_alive: None,
-                })
-                .await?
-                .with_service(move || running.clone());
+                let service = StreamableHttpService::new(
+                    move || Ok(running.clone()),
+                    LocalSessionManager::default().into(),
+                    StreamableHttpServerConfig {
+                        sse_keep_alive: None,
```
DaleSeo: This option name looks weird. Why `sse` in the Stream HTTP config?

crates/apollo-mcp-server/src/server/states.rs
```diff
@@ -1,7 +1,7 @@
 use apollo_compiler::{Schema, validation::Valid};
 use apollo_federation::{ApiSchemaOptions, Supergraph};
 use apollo_mcp_registry::uplink::schema::{SchemaState, event::Event as SchemaEvent};
-use futures::{FutureExt as _, Stream, StreamExt as _, future, stream};
+use futures::{FutureExt as _, Stream, StreamExt as _, stream};
```
DaleSeo, line 4: Just out of curiosity, why do we import these traits as `_`?

DaleSeo (commented):
Thanks for the update, @pubmodmatt! I tested it with MCP Inspector, and I was able to shut down the server using `Ctrl + C` after disconnecting from MCP Inspector. However, when I pressed `Ctrl + C` first on the server, I couldn't disconnect or reconnect from MCP Inspector. The only way to shut down the server was to terminate MCP Inspector. I've recorded the issue in a video:

https://github.com/user-attachments/assets/2fecf677-5c32-4996-9e58-4275a3500225

crates/apollo-mcp-server/src/server/states/starting.rs
pubmodmatt: Streamable HTTP uses SSE internally.

DaleSeo: Huh... makes sense!

crates/apollo-mcp-server/src/server/states.rs
pubmodmatt: This is called an "extension trait" in Rust, and it adds functionality to an existing trait, but isn't used directly. It needs to be brought into scope, but it doesn't need a name.

crates/apollo-mcp-server/src/server/states/starting.rs
pubmodmatt: The existing SSE code still works, so I wasn't going to mess with it to ensure it stays the same. But if you feel strongly that it should be updated, I can do that.

pubmodmatt: As for the Ctrl+C behavior, it works the same for the Apollo MCP Server as it does for the examples. Although inconvenient, it's how the SDK operates now.

DaleSeo: @pubmodmatt I'm fine with leaving the SSE code as it is. I just wanted to share the example code in case you missed it.

I think I liked the old behavior for Ctrl+C. My main concern is that when the server is running in production, there are times when you need to shut it down, no matter the open connections. This could be during deployments, when updating the binary version, or when changing the configuration.

pubmodmatt: It shouldn't be a problem in production, where you'd be shutting down the container. It's really only a dev issue, and there you can always use `kill` to end the process if necessary. My view is that we need to get on the lastest SDK as it's critical that we can implement the latest MCP features. The Ctrl+C issue is just a very minor annoyance. We need to merge this PR no matter what. But separately, either creating an issue or PR with a fix for modelcontextprotocol/rust-sdk would be the way forward to fix the Ctrl+C issue.

DaleSeo: @pubmodmatt You just made me realize that kill and Ctrl+C aren't the same thing. 😂 That sounds good!

PR #176, opened 2025-07-07
Invalid operation should not crash the MCP Server

DaleSeo:
Currently, the MCP server crashes when it encounters syntactically incorrect GraphQL operations. Since users can create such operations in Operation Collections or static files, this can lead to a frustrating experience and reliability issues in production. This PR implements graceful error handling by skipping invalid operations, ensuring that only valid ones are available.

## Testing

Given the following operation: 

```graphql
query GetAlerts($state: String!) {
  alerts(state: $state) {
    severity
    description
    instruction
  }

```

(Notice there is no closing curly bracket at the end.)

### Before: Invalid GraphQL operations crashed the entire MCP server

Shot 2025-07-07 at 16 12 59

### After: Invalid operations are logged and skipped, while the server continues running

Shot 2025-07-07 at 16 13 56

Shot 2025-07-07 at 16 19 16

### DIFF
--- a/crates/apollo-mcp-server/src/server/states/running.rs
+++ b/crates/apollo-mcp-server/src/server/states/running.rs
@@ -13,7 +13,7 @@ use rmcp::{
 use serde_json::Value;
 use tokio::sync::{Mutex, RwLock};
 use tokio_util::sync::CancellationToken;
-use tracing::{debug, error, warn};
+use tracing::{debug, error};
 
 use crate::{
     custom_scalar_map::CustomScalarMap,
@@ -68,7 +68,7 @@ impl Running {
                 ) {
                     Ok(operation) => operation,
                     Err(error) => {
-                        warn!("Invalid operation: {}", error);
+                        error!("Invalid operation: {}", error);
                         None
                     }
                 }
@@ -111,7 +111,7 @@ impl Running {
                     ) {
                         Ok(operation) => operation,
                         Err(error) => {
-                            warn!("Invalid operation: {}", error);
+                            error!("Invalid operation: {}", error);
                             None
                         }
                     }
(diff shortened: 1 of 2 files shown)


crates/apollo-mcp-server/src/server/states/running.rs
```diff
@@ -58,18 +58,21 @@ impl Running {
...
                     &schema,
                     self.custom_scalar_map.as_ref(),
                     self.mutation_mode,
                     self.disable_type_description,
                     self.disable_schema_description,
-                )
+                ) {
+                    Ok(operation) => operation,
+                    Err(error) => {
+                        warn!("Invalid operation: {}", error);
```
pubmodmatt: This seems like an error, not a warning.

pubmodmatt:
My original philosophy was that the server should not start up if the operations are not valid. This avoids any case where the server is running in a broken state.

These changes mean that the server will start up and run with an invalid configuration. The only way to detect this case currently would be to directly monitor the log output. So if someone deployed a change to an operation used by a production MCP server, and there was an invalid operation, they may fail to detect that case. Their end users (using the MCP server from Claude, for example) would be the first to encouter the issue becasue the MCP tool for that operation would be unavailable. If we go this route, we need mechanisms to detect this case. For example, metrics could be emitted through OTEL indicating that there are invalid operations, which could trigger an alarm.

However, another concern is hot reloading operations. For example, if a user saves an invalid operation to an operation collection through Apollo Explorer, this would currently cause all the running MCP servers to exit, which is not good. So the motivation here is sound, I just have concerns that we need to make this more detectable.

DaleSeo:
@pubmodmatt Thank you for explaining the reasoning behind the initial implementation and the consequences of changing it. I think that's valid concern and we can make invalid operations more detectable. I'll open a [ticket](https://apollographql.atlassian.net/browse/GT-292) to address your concern.

pubmodmatt (approved):

PR #178, opened 2025-07-08
Minify introspect return value

pubmodmatt:
Add options to minify the type information returned by the `introspect` and `search` tools. This emulates the [GSL format defined here](https://github.com/hwillson/gsl) in Rust. This removes extraneous whitespace and braces, and shortens built-in type names and keywords. It also currently removes directives, though these may be considered useful to the LLM in some cases.

To enable minification, turn on the `minify` option in the config file, for example:

```yaml
introspection:
  introspect:
    enabled: true
    minify: true
  search:
    enabled: true
    minify: true
```

For the top level Query type in the weather example schema, the original SDL was:

```graphql
type Query {
  """Get the weather forecast for a coordinate"""
  forecast(coordinate: InputCoordinate!): Forecast
  """Get the weather alerts for a state"""
  alerts(
    """The two-letter state abbreviation (e.g., 'CO' for Colorado)"""
    state: String!,
  ): [Alert]
}
```

Minified, this becomes:

```
T:Query:"Gettheweatherforecastforacoordinate"forecast(InputCoordinate!):Forecast,"Gettheweatheralertsforastate"alerts("Thetwo-letterstateabbreviation(e.g.,'CO'forColorado)"state:s!):[Alert]
```

This reduces the size from 272 -> 190 bytes, about a 30% reduction.

For a more real-world example, the `Repository` type in the GitHub schema went from 35322 -> 24418 bytes, about a 31% reduction.


alocay:
Do you have any estimate on how much this reduces the instrospect return value by?

pubmodmatt:
> Do you have any estimate on how much this reduces the instrospect return value by?

I've updated the PR description with some examples.

crates/apollo-mcp-server/src/introspection/tools/introspect.rs
```diff
@@ -113,15 +110,42 @@ impl Introspect {
...
             is_error: None,
         })
     }
 }
 
+#[cfg_attr(feature = "minify", allow(unused_variables))]
+fn tool_description(root_query_type: Option<String>, root_mutation_type: Option<String>) -> String {
+    #[cfg(feature = "minify")]
+    {
+        "Get GraphQL type information;T=type,I=input,E=enum,U=union,F=interface;s=String,i=Int,f=Float,b=Boolean,d=ID;!=required,[]=list;".to_string()
```
DaleSeo: So... the tool description is hard-coded initially, and then later, when it's serialized, it gets interpolated somehow in `manify.rs`. Am I getting this right, @pubmodmatt?

crates/apollo-mcp-server/src/introspection/tools/introspect/minify.rs
```diff
@@ -0,0 +1,169 @@
+use apollo_compiler::schema::{ExtendedType, Type};
+
+pub trait MinifyExt {
+    /// Serialize in minified form
+    fn minify(&self) -> String;
+}
+
+impl MinifyExt for ExtendedType {
+    fn minify(&self) -> String {
```
DaleSeo: I can't understand the complex logic here... 😵

crates/apollo-mcp-server/src/introspection/tools/introspect.rs
pubmodmatt: `minify.rs` encodes responses into the small format. This string tells the LLM how to decode that format.

DaleSeo: Got it, thanks! Now I can see the shortened types. 👀

crates/apollo-mcp-server/src/introspection/tools/introspect/minify.rs
pubmodmatt: Unfortunately `apollo-compiler` ergonomics generally result in a lot of code duplcation since the various types don't have traits in common. But this is really pretty simple - instead of using the `ExtendedType.serialize()` method, which outputs the full GraphQL SDL, this alternative `ExtendedType.minify()` method returns the smaller form for each type. So for each variant, it's building up the small string, which includes descriptions, field names, and arguments.

pubmodmatt: @DaleSeo - I refactored the `minify.rs` file to (hopefully) make it simpler to read and maintain.

crates/apollo-mcp-server/src/introspection/tools/introspect.rs
```diff
@@ -97,13 +94,41 @@ impl Introspect {
...
+    }
+}
+
+fn tool_description(
+    root_query_type: Option<String>,
+    root_mutation_type: Option<String>,
+    minify: bool,
+) -> String {
+    if minify {
+        "Get GraphQL type information - T=type,I=input,E=enum,U=union,F=interface;s=String,i=Int,f=Float,b=Boolean,d=ID;!=required,[]=list,<>=implements;".to_string()
```
swcollard, line 120: Is this how the LLM knows how to un-minify to execute actual queries? Or does it not need that information because the type and field names are unchanged?

pubmodmatt: Yes, exactly! This tiny bit of text is all the model needs to undo the minification. A big caveat here is that I haven't done extensive testing of this across different models, and only really tested with Claude Sonnet 4. Smaller models may struggle with it. I consider this feature to be experimental for that reason. There are other ways more detailed decoding information could be given to a model, such as through resources or system prompts. Or if the spec for the minified format were publicly documented, eventually models would have it as part of their training.

swcollard (approved):

PR #181, opened 2025-07-08
Fix an issue with rmcp 0.2.x upgrade

pubmodmatt:
The changes to upgrade to rmcp 0.2.x caused the state machine to be blocked on startup, so no events could be processed until the axum server was shut down (for example, by Ctrl+C). Fix by spawning like the old 0.1.x method did.

### DIFF
--- a/crates/apollo-mcp-server/src/server/states/running.rs
+++ b/crates/apollo-mcp-server/src/server/states/running.rs
@@ -59,19 +59,16 @@ impl Running {
             .cloned()
             .map(|operation| operation.into_inner())
             .filter_map(|operation| {
-                match operation.into_operation(
+                operation.into_operation(
                     &schema,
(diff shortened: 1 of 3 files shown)


DaleSeo (approved):
Confirmed it fixes the issue for both local operation files and operation collections.

crates/apollo-mcp-server/src/server/states/running.rs
```diff
@@ -59,19 +59,18 @@ impl Running {
             .cloned()
             .map(|operation| operation.into_inner())
             .filter_map(|operation| {
-                match operation.into_operation(
```
Jephuff, line 62: Is this related to the fix  or just a style you like more? They look functionally identical I think?

Jephuff: Ahh found it in the lint fix commit!

PR #188, opened 2025-07-10
Gt 167 log to a file

alocay:
Add support for logging to a file. Defaults to `./logs` if a path is not given using the new `--out` argument. On any failure setting up the rolling file appender than logging fallbacks to stderr.


alocay:
I'll update the docs for this new argument if we agree on this change

PR #190, opened 2025-07-11
docs: add the mcp-explorer option to rover dev

DaleSeo:
Just to align with https://github.com/apollographql/rover/pull/2661


Jephuff:
Ooh I don't think we want to add --explorer to rover dev, it doesn't really make sense in that context because the schema might not match. I think I'd rather we wait until we have a --sandbox flag and add that to rover dev?

DaleSeo:
@Jephuff All these CLI options will likely be replaced with the config file in v0.6. For now, I'm just trying to document what's currently supported in v0.5.

PR #192, opened 2025-07-14
update config files for example graphs

DaleSeo:
This PR updates the configuration files for example graphs to make it easier to test all its features.


graphql/TheSpaceDevs/config.yaml
```diff
@@ -1,9 +1,19 @@
+endpoint: https://thespacedevs-production.up.railway.app/
```
pubmodmatt, line 1: Nice catch!

pubmodmatt (approved):

PR #195, opened 2025-07-15
Limit search results to just fields on root path

pubmodmatt:
These changes greatly reduce and improve the results of the `search` tool. Rather than returning *all* type information for all types on the root paths, it now returns only information related to the fields along the path. For leaf types, it still returns the full set of fields.

Additionally, it now includes input types related to each field along the path. This information is critical for the AI model to be able to succesfully query the field.

For example, previously, when querying about space launches in `TheSpaceDevs` example, it would return a huge amount of type information. It now limits this to exactly the information the model will need to query the matching types. The model now gets the information it needs, and just that information, in a single call to the `search` tool:

```graphql
type Launch {
  id: ID!
  name: String
  launchDesignator: String
  status: LaunchStatus
  lastUpdated: String
  net: String
  netPrecision: String
  window: LaunchWindow
  image: Image
  infographic: String
  probability: Float
  weatherConcerns: String
  failreason: String
  hashtag: String
  provider: Agency
  rocket: Rocket
  mission: Mission
  pad: Pad
  webcastLive: Boolean
  program: Program
  orbitalLaunchAttemps: Int
  locationLaunchAttemps: Int
  padLaunchAttemps: Int
  agencyLaunchAttemps: Int
  orbitalLaunchAttempsYear: Int
  locationLaunchAttempsYear: Int
  padLaunchAttempsYear: Int
  agencyLaunchAttempsYear: Int
}

type LaunchConnection {
    results: [Launch]
}

type Query {
    launch(id: ID!): Launch
    launches(search: String, limit: Int = 5, offset: Int = 0): LaunchConnection
    previousLaunces(search: String, limit: Int = 5, offset: Int = 0): LaunchConnection
    upcomingLaunches(search: String, limit: Int = 5, offset: Int = 0): LaunchConnection
}
```


crates/apollo-mcp-server/src/introspection/tools/search.rs
```diff
@@ -20,15 +20,13 @@ use tracing::debug;
 /// The name of the tool to search a GraphQL schema.
 pub const SEARCH_TOOL_NAME: &str = "search";
 
-/// The depth of nested types to include for leaf nodes on matching root paths.
-pub const LEAF_DEPTH: DepthLimit = DepthLimit::Limited(1);
```
DaleSeo, line 24: How about we extract the magic number `5` on L87 and put it here?

```rust
const MAX_SEARCH_RESULTS: usize = 5;

// L87
root_paths.truncate(MAX_SEARCH_RESULTS);
```

clippy.toml
```diff
@@ -1,3 +1,4 @@
 allow-expect-in-tests = true
 allow-panic-in-tests = true
 allow-unwrap-in-tests = true
+allow-indexing-slicing-in-tests = true
```
DaleSeo, line 4: Does this mean we have something like `vec[0...3]` in the test?

crates/apollo-mcp-server/src/main.rs
```diff
@@ -17,6 +17,12 @@ use tracing_subscriber::EnvFilter;
 
 mod runtime;
 
+/// The default search leaf depth
+const DEFAULT_SEARCH_LEAF_DEPTH: usize = 1;
+
+/// The default amount of memory to use for GraphQL schema indexing
+const DEFAULT_INDEX_MEMORY_BYTES: usize = 50_000_000;
```
DaleSeo: Would this be the best place for the default values? How about putting them together with the `SearchConfig` struct like this?

```rust
impl SearchConfig {
    /// The default search leaf depth
    pub const DEFAULT_LEAF_DEPTH: usize = 1;

    /// The default amount of memory to use for GraphQL schema indexing
    pub const DEFAULT_INDEX_MEMORY_BYTES: usize = 50_000_000;
```

crates/apollo-schema-index/src/lib.rs
```diff
@@ -170,14 +171,27 @@ impl SchemaIndex {
             .register("en_stem", text_analyzer.clone());
 
         // Map every type in the schema to the types referencing it
-        let mut index_writer = index.writer(INDEX_MEMORY_BYTES)?;
+        let mut index_writer = index.writer(index_memory_bytes)?;
```
DaleSeo, line 174: Nice 👍

crates/apollo-schema-index/src/lib.rs
```diff
@@ -170,14 +171,27 @@ impl SchemaIndex {
...
         let mut type_references: HashMap<String, Vec<String>> = HashMap::default();
         for (extended_type, path) in schema.traverse(root_types) {
             let entry = type_references
                 .entry(extended_type.name().to_string())
                 .or_default();
-            if let Some(ref_type) = path.referencing_type() {
```
DaleSeo, line 179: I'm curious about how it even worked before if `path.referencing_type()` returned a tuple. 🤔
Should it have been like this?

```rust
if let Some((ref_type)) = path.referencing_type() {
```

clippy.toml
pubmodmatt: Or just v[1]. Any index or slice can panic. See: https://rust-lang.github.io/rust-clippy/master/index.html#indexing_slicing

crates/apollo-schema-index/src/lib.rs
pubmodmatt: It wasn't a tuple previously, just a `NamedType`

crates/apollo-mcp-server/src/introspection/tools/search.rs
pubmodmatt: Sure, I'll update that. Ultimately, some of these should probably be exposed as config, but there's a balance between exposing enough config and too much.

crates/apollo-schema-index/src/lib.rs
DaleSeo: which can be destructured like the updated code?

```rust
if let Some((ref_type, field_name, field_args)) = path.referencing_type() {
```

pubmodmatt: In the updated code, it's a tuple, so its just destructuring the tuple.

PR #200, opened 2025-07-16
feat: adding a log file support

alocay:
Adding support for logging to a file via config option `logging.path`. When provided the server will attempt to build the log path and then initialize a RollingFileAppender log writer. If any of this failed or if not provided, logging defaults to stderr. 

Note: Decided to default to stderr (when not provided) since the server will only be able to log out to specific places which it has permissions for. Since this isn't not easily known, the easiest thing is to defualt to stderr and let the user setup file logging, pointing to a location that it can log to.

### DIFF
--- a/crates/apollo-mcp-server/src/runtime/config.rs
+++ b/crates/apollo-mcp-server/src/runtime/config.rs
@@ -1,15 +1,14 @@
 use std::path::PathBuf;
 
 use apollo_mcp_server::{health::HealthCheckConfig, server::Transport};
-use crate::runtime::logging::Logging;
 use reqwest::header::HeaderMap;
 use schemars::JsonSchema;
 use serde::Deserialize;
 use url::Url;
 
 use super::{
-    OperationSource, SchemaSource, graphos::GraphOSConfig, introspection::Introspection,
-    logging::Logging, overrides::Overrides,
+    OperationSource, SchemaSource, endpoint::Endpoint, graphos::GraphOSConfig,
+    introspection::Introspection, logging::Logging, overrides::Overrides,
 };
 
 /// Configuration for the MCP server


crates/apollo-mcp-server/src/runtime.rs
```diff
@@ -35,6 +41,82 @@ pub fn read_config(yaml_path: impl AsRef<Path>) -> Result<Config, figment::Error
...
+    match ensure_log_dir_exists(log_path.clone()) {
+        Ok(..) => {}
+        Err(_err) => {
+            eprintln!("Could not build log path - falling back to stderr");
+            return setup_stderr_logging(env_filter);
+        }
+    }
+
+    let (non_blocking_writer, guard) = match RollingFileAppender::builder()
+        .rotation(Rotation::NEVER)
```
pubmodmatt: Can we add an option for rotation, and default it to something fairly small? Lack of log rotation causing disk space exhaustion is one of the most frequent causes of service outages.

crates/apollo-mcp-server/src/runtime.rs
```diff
@@ -35,6 +41,82 @@ pub fn read_config(yaml_path: impl AsRef<Path>) -> Result<Config, figment::Error
         .extract()
 }
 
+/// Sets up either file logging or stderr logging depending on provided configuration options
```
pubmodmatt: Small nit, but there's enough logging code now that you might consider moving this to a `logging` module.

alocay: Ok I moved it over although let me know if there is an issue with how i set it up... I find Rust's module system a bit wonky.

crates/apollo-mcp-server/src/runtime.rs
alocay: I added the `rotation` option, defaulting to the smallest unit (minute).

crates/apollo-mcp-server/src/log/logging.rs
```diff
@@ -12,30 +21,46 @@ pub struct Logging {
...
+    use super::LogRotationKind;
     use tracing::Level;
 
-    pub(super) const fn log_level() -> Level {
+    pub(crate) const fn log_level() -> Level {
         Level::INFO
     }
+
+    pub(crate) const fn default_rotation() -> LogRotationKind {
+        LogRotationKind::Minutely
```
pubmodmatt: NIt: this may be too far - I'd probably go with hourly here

crates/apollo-mcp-server/src/log.rs
```diff
@@ -0,0 +1,103 @@
+//! Logging utilities
+//!
+//! This module is only used by the main binary and provides logging setup
+
+pub(crate) mod logging;
```
pubmodmatt: Rather than having a `log` module with a `logging` submodule , just make a `logging` module directly.

alocay:
Moved logging.rs to src/ and combined with logging setup functions that were added

crates/apollo-mcp-server/src/main.rs
```diff
@@ -1,19 +1,18 @@
 use std::path::PathBuf;
 
+use crate::runtime::logging::init::setup_logging;
```
nicholascioli: Nit: Re-export the `setup_logging` method under the `logging` module and rename it to `setup` so that it doesn't embed the name of the module into it. Also merge it with the existing `use` on line 14. 

Then you could use it like:

```rust
// On line 14
use runtime::{logging, IdOrDefault};

// Wherever you used `setup_logging`
logging::setup(...);
```

nicholascioli: Actually, refer to the comment on the logging struct. This should probably be a member function of logging.

crates/apollo-mcp-server/src/runtime.rs
```diff
@@ -1,12 +1,12 @@
...
+//! Runtime utilities
 //!
 //! This module is only used by the main binary and provides helper code
 //! related to runtime configuration.
 
 mod config;
 mod graphos;
 mod introspection;
-mod logging;
+pub(crate) mod logging;
```
nicholascioli: Nit: This can probably be not public if the setup function is a member on the logging struct.

crates/apollo-mcp-server/src/runtime/config.rs
```diff
@@ -1,16 +1,16 @@
 use std::path::PathBuf;
 
+use super::{
+    OperationSource, SchemaSource, graphos::GraphOSConfig, introspection::Introspection,
+    overrides::Overrides,
+};
+use crate::runtime::logging::Logging;
```
nicholascioli: Nit: These can be merged by just using `super::logging:Logging`

nicholascioli: Also, any reason why the super imports were moved here?

crates/apollo-mcp-server/src/runtime/logging.rs
```diff
@@ -1,5 +1,14 @@
+//! Logging config and utilities
+//!
+//! This module is only used by the main binary and provides logging config structures and setup
+//! helper functions
+pub(crate) mod init;
```
nicholascioli: Nit: This can be private.

crates/apollo-mcp-server/src/runtime/logging.rs
```diff
@@ -1,5 +1,14 @@
...
+mod log_rotation_kind;
+
+use crate::runtime::logging::log_rotation_kind::LogRotationKind;
```
nicholascioli: Nit: This should probably just be `log_rotation_kind::LogRotationKind`

crates/apollo-mcp-server/src/runtime/logging.rs
```diff
@@ -10,32 +19,48 @@ pub struct Logging {
...
         }
     }
 }
 
 mod defaults {
+    use super::LogRotationKind;
     use tracing::Level;
 
-    pub(super) const fn log_level() -> Level {
+    pub(crate) const fn log_level() -> Level {
```
nicholascioli: Nit: This should probably stay as `pub(super)` since I don't think anything else needs it?

crates/apollo-mcp-server/src/runtime/logging.rs
```diff
@@ -10,32 +19,48 @@ pub struct Logging {
...
         Level::INFO
     }
+
+    pub(crate) const fn default_rotation() -> LogRotationKind {
+        LogRotationKind::Hourly
+    }
```
nicholascioli: Nit: Same as above

crates/apollo-mcp-server/src/runtime/schemas.rs
nicholascioli, line 1: Nit: This doesn't have to be done in this PR, but these should probably just live in their corresponding files since they aren't reused across different modules.

crates/apollo-mcp-server/src/runtime/logging/init.rs
```diff
@@ -0,0 +1,85 @@
...
+            .add_directive("rmcp=warn".parse()?)
+            .add_directive("tantivy=warn".parse()?);
+    }
+
+    if let Some(path) = &config.logging.path {
+        setup_file_logging(path, env_filter, config.logging.rotation.clone())
+    } else {
+        setup_stderr_logging(env_filter)
+    }
+}
```
nicholascioli: This method should probably be a member function on `Logging`. You can either have an `impl Logging` block in this file or merge it with the parent module.

crates/apollo-mcp-server/src/runtime/logging/init.rs
```diff
@@ -0,0 +1,85 @@
...
+        )
+        .init();
+
+    Ok(None)
+}
+
+/// Creates any missing directories in the log output path
+fn ensure_log_dir_exists(dir: PathBuf) -> std::io::Result<()> {
+    std::fs::create_dir_all(dir)
+}
```
nicholascioli: Nit: This can probably be deleted.

crates/apollo-mcp-server/src/runtime/logging/init.rs
```diff
@@ -0,0 +1,85 @@
...
+        .with(env_filter)
+        .with(
+            tracing_subscriber::fmt::layer()
+                .with_writer(non_blocking_writer)
+                .with_ansi(false)
+                .with_target(false),
+        )
+        .init();
+
+    Ok(Some(guard))
```
nicholascioli: Nit: This is mostly to show that this could be simplified slightly by functional chaining, but feel free to ignore.

```suggestion
    let (writer, guard) = std::fs::create_dir_all(log_path)
        .and_then(RollingFileAppender::builder()
            .rotation(log_rotation.into())
            .filename_prefix("apollo_mcp_server")
            .filename_suffix("log")
            .build())
        .and_then(|(non_blocking, guard)| (non_blocking, Some(guard)))
        .unwrap_or_else(|| {
            eprintln!("Log file setup failed - falling back to stderr");
            
            (std::io::stderr, None)
        })
    
    tracing_subscriber::registry()
        .with(env_filter)
        .with(
            tracing_subscriber::fmt::layer()
                .with_writer(writer)
                .with_ansi(true)
                .with_target(false),
        )
        .init();
    
    Ok(guard)
```

crates/apollo-mcp-server/src/runtime/logging/init.rs
```diff
@@ -0,0 +1,85 @@
...
+}
+
+/// Sets up stderr logging
+fn setup_stderr_logging(env_filter: EnvFilter) -> Result<Option<WorkerGuard>, anyhow::Error> {
```
nicholascioli: Nit: this is probably not needed.

crates/apollo-mcp-server/src/runtime/logging/log_rotation_kind.rs
```diff
@@ -0,0 +1,42 @@
...
+    #[case(LogRotationKind::Hourly, Rotation::HOURLY)]
+    #[case(LogRotationKind::Daily, Rotation::DAILY)]
+    #[case(LogRotationKind::Never, Rotation::NEVER)]
+    fn it_maps_to_rotation_correctly(
+        #[case] log_rotation_kind: LogRotationKind,
+        #[case] expected: Rotation,
+    ) {
+        let actual: Rotation = log_rotation_kind.into();
+        assert_eq!(expected, actual);
+    }
```
nicholascioli, line 45: Nit: This test seems weird to me since the `From` impl is just a straight mapping, but feel free to leave it.

nicholascioli (commented):
Looks good! I left a few nits.

crates/apollo-mcp-server/src/runtime/config.rs
alocay: Hmm I was altering things but I don't remember moving any imports myself. Maybe it was RustRover or clippy after I made some change (which was removed later).

crates/apollo-mcp-server/src/runtime/logging/log_rotation_kind.rs
alocay: I added it mainly for coverage and to break in case someone changes something with the mapping. The test will then at least alert you that you broke some expected mapping between the two.

alocay:
> Assuming all other comments have been address this looks good to me. Tested locally and got everything working great.
> 
> Can you expand why falling back to stderr instead of failing to start up the server was chosen? For example if I set the log directory to be only accessible by root, I get the error
> 
> ```
> Failed to setup logging: InitError { context: "failed to create initial log file", source: Os { code: 13, kind: PermissionDenied, message: "Permission denied" } }
> Log file setup failed - falling back to stderr
> ```
> 
> And then the server keeps going as if nothing was wrong

I did not think that failing to start the server completely simply because log file logging could not be setup would be a great experience. This felt especially true since understanding MCP Server failures when using LLM clients such as Claude Desktop are not very obvious. We can easily alter tis behavior though if the consenus to go this route though.

PR #202, opened 2025-07-17
Add binary to output JSON schema for config

pubmodmatt:
Adds a new bin to generate the JSON Schema for Apollo MCP Server config:
```
cargo run --bin config-schema > config.schema.json
```

This can be used to configure editors for the config file, to enable things like validation and auto-completion.

For example, in VS Code, you can install the [YAML Language Support extension](https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml), then configure something like the following in the VS Code user settings:

```json
    "yaml.schemas": {
        "./config.schema.json": [
            "**/config.yaml",
        ],
    }
```

or in the file itself:

```yaml
# yaml-language-server: $schema=./config.schema.json
```

Another alternative might be to allow users to specify the config file in JSON format directly (which the figment crate [supports](https://docs.rs/figment/0.10.19/figment/providers/struct.Json.html)). Most editors have support for this without requiring addiional extensions or configuration. For example, you can specify the following in a JSON file:

```json
"$schema": "./config.schema.json",
```

Most editors will pick this up and use it without any further configuration. The file could be specified as a URL hosted on Apollo's domain or directly to the file in GitHub, for example.

It'd also be desirable to add genration of the schema file to the build so any changes get picked up automatically, but that hasn't been done here.


nicholascioli (approved):
LGTM! Thanks for adding this.

PR #203, opened 2025-07-17
feat: Validate tool for verifying graphql queries before executing them

swcollard:
The introspection options in the mcp server provide introspect, execute, and search tools. The LLM often tries to validate its queries by just executing them. This may not be desired (there might be side effects, for example). This PR adds a `validate` tool so the LLM can validate the operation without actually hitting the GraphQL endpoint. It first validates the syntax of the operation, and then checks it against the introspected schema for validation.


crates/apollo-mcp-server/src/introspection/tools/execute.rs
```diff
@@ -36,7 +36,7 @@ impl Execute {
             mutation_mode,
             tool: Tool::new(
                 EXECUTE_TOOL_NAME,
-                "Execute a GraphQL operation. Use the `introspect` tool to get information about the GraphQL schema. Always use the schema to create operations - do not try arbitrary operations. DO NOT try to execute introspection queries.",
+                "Execute a GraphQL operation. Use the `introspect` tool to get information about the GraphQL schema. Always use the schema to create operations - do not try arbitrary operations. Use the `validate` tool to validate operations. DO NOT try to execute introspection queries.",
```
DaleSeo: How will this be interpreted by LLMs when the validate option is off?

```suggestion
                "Execute a GraphQL operation. Use the `introspect` tool to get information about the GraphQL schema. Always use the schema to create operations - do not try arbitrary operations. Use the `validate` tool to validate operations if is available. DO NOT try to execute introspection queries.",
```

Not sure if this will make any real difference, though. 😂

crates/apollo-mcp-server/src/introspection/tools/validate.rs
```diff
@@ -0,0 +1,115 @@
...
+                McpError::new(
+                    ErrorCode::INVALID_PARAMS,
+                    "Invalid operation type".to_string(),
+                    None,
+                )
+            })?;
+
+        let schema_guard = self.schema.lock().await;
+        Parser::new()
+            .parse_executable(&schema_guard, input.operation.as_str(), "operation.graphql")
```
DaleSeo, line 66: Interesting... I also noticed a method called `parse_mixed_validate` and wondered if it was intended for this scenario?

crates/apollo-mcp-server/src/introspection/tools/validate.rs
```diff
@@ -0,0 +1,115 @@
...
+mod tests {
+    use serde_json::json;
+
+    use super::*;
+    static SCHEMA: std::sync::LazyLock<Arc<Mutex<Valid<Schema>>>> =
+        std::sync::LazyLock::new(|| {
+            Arc::new(Mutex::new(
+                Schema::parse_and_validate("type Query { id: ID! }", "schema.graphql").unwrap(),
+            ))
+        });
```
DaleSeo, line 85: This is pretty challenging for a Rust newbie. 😹

crates/apollo-mcp-server/src/introspection/tools/validate.rs
alocay, line 66: if `parse_executable` is only reading and there is no concurrent writes happening to the schema, should we use a `RwLock` instead of `Mutex`? I believe Nick was telling me that the `Mutex` will block on both reads and writes and `RwLock` only blocks on writes and allows reads to go through. If there is no worry of writes happening, then `RwLock` should be safe.

crates/apollo-mcp-server/src/introspection/tools/validate.rs
alocay: I find testing/mocking in general in Rust a bit challenging and odd. In my opinion in regards to unit testing (and if it was another language) I would simply mock out `Parser::new().parse_executable()` in some way and have the mock return the valid or invalid result. Then validate that the function was called with the right inputs. By actually passing in invalid graphql and checking that it's invalid we are dipping into the realm of testing the `apollo_compiler` itself which depending on who you ask is a bit of a no-no. Either way though, I don't know how to do what I'm suggesting in Rust 🤷🏼‍♂️ ... I'm too new.

crates/apollo-mcp-server/src/operations.rs
```diff
@@ -370,8 +370,12 @@ pub fn operation_defs(
     allow_mutations: bool,
     source_path: Option<String>,
 ) -> Result<Option<(Document, Node<OperationDefinition>, Option<String>)>, OperationError> {
+    let source_path_clone = source_path.clone();
     let document = Parser::new()
-        .parse_ast(source_text, "operation.graphql")
+        .parse_ast(
+            source_text,
+            source_path_clone.unwrap_or_else(|| "operation.graphql".to_string()),
```
alocay, line 377: What does this do?

crates/apollo-mcp-server/src/runtime/introspection.rs
```diff
@@ -57,6 +60,14 @@ impl Default for SearchConfig {
     }
 }
 
+/// Validation tool configuration
+#[derive(Debug, Default, Deserialize, JsonSchema)]
+#[serde(default)]
+pub struct ValidateConfig {
+    /// Enable validation tool
+    pub enabled: bool,
```
alocay, line 75: I think I missed it, where is this used? It seems like it's in `starting.rs` via
```
let validate_tool = self
            .config
            .validate_introspection
            .then(|| Validate::new(schema.clone()));
```
but i figured int he config it would have been accessed via config.validate but I'm probably misunderstanding something.

crates/apollo-mcp-server/src/introspection/tools/validate.rs
swcollard: I looked into `parse_mixed_validate` and it is expecting the schema and the queries/mutations to be in the same document (String)

In our purpose we are getting the schema separately from the Query that the LLM is generating so we'd have to concatenate them together to use the other method

swcollard: I have also never used Rust Mutex's before, I mostly pulled this in because its a clone of the same wrappers around the schema type the introspection and execute tools are using.

crates/apollo-mcp-server/src/runtime/introspection.rs
swcollard: The config object is used in main.rs where we set this `validate_introspection` bool

```rust
.validate_introspection(config.introspection.validate.enabled)
```

crates/apollo-mcp-server/src/operations.rs
swcollard: This was a miss from a previous PR where we used the file path to indicate where invalid operations were. If we have the path, this change uses it instead of a generic operation.graphql filename

PR #209, opened 2025-07-24
Health Check Support

### LINKED ISSUE #168: Health check for MCP Server
There currently is no health check for the Apollo MCP Server if I wanted to deploy it isolated from my GraphQL API.

If you plan on using the Apollo Router, you could use the [Apollo Runtime Container](https://github.com/apollographql/apollo-runtime) that includes both the MCP Server and an Apollo Router instance that can be used. The Apollo Router can be configured to have a [health check ](https://www.apollographql.com/docs/graphos/routing/configuration/yaml#health_check).

DaleSeo:
fixes #168

This PR adds health reporting functionality similar to Apollo Router. This makes the MCP server ready for production deployment with proper health monitoring and Kubernetes integration.

### Configuration

You can configure it like Apollo Router:

```yaml
health_check:
  enabled: true
  path: "/health"
  readiness:
    allowed: 100 # Maximum number of rejections allowed per sampling interval
    interval:
      sampling: "5s" # How often to sample for readiness
      unready: "10s" # How long to stay unready when threshold is exceeded
```

Minimum configuration:

```yaml
health_check:
  enabled: true
```

### Test

Basic health check

```bash
$ curl http://127.0.0.1:8088/health
{"status":"UP"}
```

Kubernetes Liveness Probe

```bash
$ curl "http://127.0.0.1:8088/health?live"
{"status":"UP"}
```

Kubernetes Readiness Probe

```bash
$ curl "http://127.0.0.1:8088/health?ready"
{"status":"UP"}
```

### DIFF
--- a/crates/apollo-mcp-server/src/health.rs
+++ b/crates/apollo-mcp-server/src/health.rs
@@ -100,7 +100,7 @@ pub struct HealthCheckConfig {
 impl Default for HealthCheckConfig {
     fn default() -> Self {
         Self {
-            enabled: true,
+            enabled: false,
             path: "/health".to_string(),
             readiness: Default::default(),
         }
@@ -220,7 +220,7 @@ mod tests {
     #[test]
     fn test_health_check_default_config() {
         let config = HealthCheckConfig::default();
-        assert!(config.enabled);
+        assert!(!config.enabled);
         assert_eq!(config.path, "/health");
         assert_eq!(config.readiness.allowed, 100);
         assert_eq!(config.readiness.interval.sampling, Duration::from_secs(5));


crates/apollo-mcp-server/src/health.rs
```diff
@@ -0,0 +1,259 @@
...
+pub struct HealthCheckConfig {
+    /// Set to false to disable the health check
+    pub enabled: bool,
+
+    /// Optionally set a custom healthcheck path
+    /// Defaults to /health
+    pub path: String,
+
+    /// Optionally specify readiness configuration
+    pub readiness: ReadinessConfig,
```
swcollard, line 97: Do we need separate configuration for liveness probe checks like we do for readiness?

swcollard (commented):
Don't forget to add a Changelog entry for this new feature ❤️

crates/apollo-mcp-server/src/health.rs
DaleSeo: @swcollard, I borrowed this pattern from Router to provide a familiar health check configuration experience. Are there any specific options you think we should support for the liveness probe?

You can check out the approach I mentioned here: https://github.com/apollographql/router/blob/96e4ca1fabb2cd1f4c86b524aa65783932aa2bec/apollo-router/src/plugins/healthcheck/mod.rs#L114

crates/apollo-mcp-server/src/health.rs
alocay, line 1: We should consider moving this to the `src/runtime` directory where the other config options are found. I believe the healthcheck  is only when the MCP server is running as a binary and not appropriate for use a lib. If I'm not mistaken I think the runtime section is for those non-lib specific modules.

crates/apollo-mcp-server/src/health.rs
swcollard: Aligning with router was in the acceptance criteria of the ticket, so I think what you have makes sense. Thanks!

crates/apollo-mcp-server/src/health.rs
```diff
@@ -0,0 +1,255 @@
...
+
+    pub fn config(&self) -> &HealthCheckConfig {
+        &self.config
+    }
+
+    pub fn get_health_state(&self, query: Option<&str>) -> (Health, StatusCode) {
+        let mut status_code = StatusCode::OK;
+
+        let health = if let Some(query) = query {
+            let query_upper = query.to_ascii_uppercase();
```
alocay, line 176: nit: I think you could shadow query here to make it cleaner:
```
let query = query.to_ascii_uppercase();
...
```

crates/apollo-mcp-server/src/health.rs
```diff
@@ -0,0 +1,255 @@
...
+    }
+
+    pub fn config(&self) -> &HealthCheckConfig {
+        &self.config
+    }
+
+    pub fn get_health_state(&self, query: Option<&str>) -> (Health, StatusCode) {
+        let mut status_code = StatusCode::OK;
+
+        let health = if let Some(query) = query {
```
alocay, line 175: nit: depending on who you ask (i find it looks cleaner) we can also change this chain of if's to a match statement:
```
let health = match query.map(|q| q.to_ascii_uppercase()) {
    Some(q) if q.starts_with("READY") => {
        /* handle READY */
    }
    Some(q) if q.starts_with("LIVE") => {
        /* handle LIVE */
     }
     None => { /* handle no query */
}
```
I'm not 100% on all the syntax but the general idea should work.

crates/apollo-mcp-server/src/health.rs
```diff
@@ -0,0 +1,255 @@
...
+            let query_upper = query.to_ascii_uppercase();
+
+            if query_upper.starts_with("READY") {
+                let status = if self.ready.load(Ordering::SeqCst) {
+                    HealthStatus::Up
+                } else {
+                    status_code = StatusCode::SERVICE_UNAVAILABLE;
```
alocay, line 182: I don't love the side effect here of setting `status_code` within this block. What if this changed to 
```
let (health, status_code) = if let Some()...
```
and this way we don't have the side effect of altering status_code inside the big if chain. It would also remove the immutability needed for `status_code`.

crates/apollo-mcp-server/src/lib.rs
```diff
@@ -3,6 +3,7 @@ pub mod errors;
 pub mod event;
 mod explorer;
 mod graphql;
+pub mod health;
```
alocay, line 6: I don't think the healthcheck is for the lib right? Should this be in `src/runtime.rs` instead?

crates/apollo-mcp-server/src/server/states/starting.rs
```diff
@@ -144,9 +160,21 @@ impl Starting {
                     LocalSessionManager::default().into(),
                     Default::default(),
                 );
-                let router = axum::Router::new().nest_service("/mcp", service);
+                let mut router = axum::Router::new().nest_service("/mcp", service);
+
+                // Add health check endpoint if configured (only supported for StreamableHttp transport)
+                if let Some(ref health_check_instance) = health_check {
+                    if health_check_instance.config().enabled {
```
alocay: You can condense these two i believe maybe with a guard or if not then with a filter. I'm not sure if a guard works with `if let Some()` but it's worth a try:
```
if let Some(ref health_check_instance) = health_check if health_check_instance.config().enabled { ... }
```
if this doesn't work then I believe this will:
```
if let Some(hc) = health_check.map(|h| h.config()).filter(|config| config.enabled) { ... }
```

alocay: We can maybe even get by without the `if let Some()` at all if we just use:
```
health_check.filter(|h| h.config().enabled).map(|h| { /* do work */ });
```

DaleSeo: Thanks @alocay, the refactored code is much clearer now.

crates/apollo-mcp-server/src/health.rs
DaleSeo: Hmm... 🤔 Can you clarify what should and shouldn’t go in `src/runtime`, based on how things are currently organized? I’m a bit confused since the `Transport` struct is also outside `src/runtime` though it contains config options.

https://github.com/apollographql/apollo-mcp-server/blob/b541f7221b9ba7f89664162cda6f7394bfc1b5f4/crates/apollo-mcp-server/src/runtime/config.rs#L53-L55

https://github.com/apollographql/apollo-mcp-server/blob/b541f7221b9ba7f89664162cda6f7394bfc1b5f4/crates/apollo-mcp-server/src/server.rs#L41-L72

crates/apollo-mcp-server/src/health.rs
DaleSeo: Hey @alocay, thanks for the suggestions. For now, I'd like to keep the code closely aligned with [the Router's implementation](https://github.com/apollographql/router/blob/96e4ca1fabb2cd1f4c86b524aa65783932aa2bec/apollo-router/src/plugins/healthcheck/mod.rs#L266). This will make it easier to borrow improvements or potentially consolidate the logic into a shared library in the future. If we diverge too much, it could complicate maintenance and code sharing.

crates/apollo-mcp-server/src/lib.rs
DaleSeo: Again, I'm not totally clear on what the criteria are for placing modules in `src/runtime` versus elsewhere. Do you think other modules should also move into `src/runtime`, or is there a specific guideline you follow?

crates/apollo-mcp-server/src/health.rs
alocay: I might be getting things mixed up. Let me see if @nicholascioli can chime to clear up the thinking behind the organization since he's the one who made this change. The one thing I believe is true though is that modules used in `lib.rs` should be for the library and not the binary. If it's for the binary to use then that should be used within `src/runtime.rs` instead of `src/lib.rs`.

crates/apollo-mcp-server/src/lib.rs
alocay: I believe `lib.rs` is where we use modules that are for the library implementation. If it's for the binary implementation then I think those are used within `runtime.rs`. That way if Apollo MCP server is ever used as a library then the we likely don't need to import binary specific modules such as logging (which is currently only used within `runtime.rs`).

riginoommen:
Enabling this Readiness and liveness checks will help a lot. thanks for this updates

riginoommen:
Something like this

```
LIVENESS_HTTP_PATH: "/health?live"
 READINESS_HTTP_PATH: "/health?ready"
```

crates/apollo-mcp-server/src/lib.rs
swcollard: ~Testing this locally we need the health import here to be used in `server.rs` but we don't need to declare it public, which, as @alocay mentions, would expose it as part of a library implementation.~

This was incorrect

DaleSeo: @swcollard I needed to expose it so that `HealthCheckConfig` is used in `config.rs`. I adopted this approach from how `Transport` is currently exposed from the `server` module.

https://github.com/apollographql/apollo-mcp-server/blob/6c677649cfec4d897187665f4ab0d3074cf6c9b7/crates/apollo-mcp-server/src/runtime/config.rs#L3

This is actually the main source of my confusion, as I mentioned in https://github.com/apollographql/apollo-mcp-server/pull/209#discussion_r2229145522.

I'm a bit lost on organizing the code. 😅 Please let me know if you have any specific restructuring suggestions!

swcollard: Weird my IDE didn't catch that issue... You're right. Ignore my comment.

crates/apollo-mcp-server/src/health.rs
swcollard: I am also a bit confused why the Transport was not also moved? But I think there is an argument to be made that a health check only applies for the Streamable HTTP transport type and wouldn't make sense when running the MCP server in StdIn mode. 

Which makes me also lean towards the health check configuration living in `src/runtime`

swcollard: I however, cannot get this refactor to work

crates/apollo-mcp-server/src/health.rs
```diff
@@ -0,0 +1,255 @@
...
+    pub path: String,
+
+    /// Optionally specify readiness configuration
+    pub readiness: ReadinessConfig,
+}
+
+impl Default for HealthCheckConfig {
+    fn default() -> Self {
+        Self {
+            enabled: true,
```
swcollard: Should the default be not enabled? So that users will need to opt in to the new functionality?

DaleSeo: Hey @swcollard, I don't know. I set it to true by default just because that was the router's default setting. Do you think we should default it to false?

> GraphOS Router and Apollo Router Core support a basic HTTP-level health check. This is enabled by default and is served on port 8088 at the URL path /health.

https://www.apollographql.com/docs/graphos/routing/self-hosted/health-checks

swcollard: I think it is again complicated by the multiple transport options. But the healthcheck only seems to work for me in `streamable_http` mode where it makes the most sense for it to be enabled by default. I get a 403 in `stdio` mode and 404 in `sse` mode.

So yes, I think we can leave it on by default, but we should make sure to call it out in the documentation

swcollard (approved):

crates/apollo-mcp-server/src/health.rs
DaleSeo: @swcollard I’ve decided to disable it for now. I’ll take some time to figure out the best way to describe this transport-specific behavior in the documentation. I’ll follow up with another PR soon.

PR #210, opened 2025-07-28
feat: add mcp auth

nicholascioli:
This commit adds support for OAuth authn/z flows at the MCP protocol level. This follows the [official specification](https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization) and makes no other assumptions.

### DIFF
--- a/crates/apollo-mcp-server/src/auth.rs
+++ b/crates/apollo-mcp-server/src/auth.rs
@@ -1,7 +1,7 @@
 use axum::{
     Json, Router,
     extract::{Request, State},
-    http::{HeaderMap, HeaderValue, StatusCode},
+    http::StatusCode,
     middleware::Next,
     response::Response,
     routing::get,
@@ -12,12 +12,11 @@ use axum_extra::{
 };
 use http::Method;
 use jsonwebtoken::{Algorithm, Validation, decode, decode_header};
-use jwks::{Jwks, JwksError};
-use reqwest::header::WWW_AUTHENTICATE;
+use jwks::Jwks;
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 use tower_http::cors::{Any, CorsLayer};
-use tracing::{error, info, warn};
+use tracing::{info, warn};
 use url::Url;
 
 mod www_authenticate;
@@ -132,7 +131,7 @@ async fn oauth_validate(
         _ => Err((
             StatusCode::UNAUTHORIZED,
             TypedHeader(WwwAuthenticate::Bearer {
-                resource_metadata: auth_config.resource,
+                resource_metadata: resource_url,
             }),
         )),
     }
--- a/crates/apollo-mcp-server/src/auth.rs
+++ b/crates/apollo-mcp-server/src/auth.rs
@@ -1,0 +1,235 @@
+use axum::{
+    Json, Router,
+    extract::{Request, State},
+    http::{HeaderMap, HeaderValue, StatusCode},
+    middleware::Next,
+    response::Response,
+    routing::get,
+};
+use http::Method;
+use jsonwebtoken::{Algorithm, Validation, decode, decode_header};
+use jwks::{Jwks, JwksError};
+use reqwest::header::{AUTHORIZATION, WWW_AUTHENTICATE};
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+use tower_http::cors::{Any, CorsLayer};
+use tracing::{error, info, warn};
+use url::Url;
+
+/// Auth configuration options
+#[derive(Debug, Clone, Deserialize, JsonSchema)]
+pub struct Config {
+    /// Disable authentication
+    #[serde(default)]
+    pub disable: bool,
+
+    /// List of upstream OAuth servers to delegate auth
+    pub servers: Vec<Url>,
+
+    /// List of accepted audiences for the OAuth tokens
+    pub audiences: Vec<String>,
+
+    /// The resource to protect.
+    ///
+    /// Note: This is usually the publically accessible URL of this running MCP server
+    pub resource: String,
+}
+
+/// A validated token string
+#[derive(Clone)]
+pub struct ValidToken(String);
+
+impl ValidToken {
+    /// Read the contents of the token, consuming it.
+    pub fn read(self) -> String {
+        self.0
+    }
+}
+
+// Note: Here we implement default for Config such that it disables auth when not
+// specified. If it is specified, however, the above `#[serde(default)]` ensures that
+// disabled is set to false.
+impl Default for Config {
+    fn default() -> Self {
+        Self {
+            disable: true,
+            servers: Vec::new(),
+            audiences: Vec::new(),
+            resource: String::new(),
+        }
+    }
+}
+
+impl Config {
+    pub fn enable_middleware(&self, router: axum::Router) -> axum::Router {
+        #[derive(Serialize)]
+        struct ProtectedResource {
+            resource: String,
+            authorization_servers: Vec<Url>,
+            bearer_methods_supported: Vec<String>,
+            scopes_supported: Vec<String>,
+            resource_documentation: String,
+        }
+
+        impl From<Config> for ProtectedResource {
+            fn from(value: Config) -> Self {
+                Self {
+                    resource: value.resource,
+                    authorization_servers: value.servers,
+                    bearer_methods_supported: vec!["header".to_string()],
+                    scopes_supported: ["profile", "email", "phone", "mcp-flow"]
+                        .into_iter()
+                        .map(str::to_string)
+                        .collect(),
+                    resource_documentation: "TODO".to_string(),
+                }
+            }
+        }
+
+        async fn protected_resource(State(auth_config): State<Config>) -> Json<ProtectedResource> {
+            Json(auth_config.into())
+        }
+
+        // Set up auth routes
+        let cors = CorsLayer::new()
+            // allow `GET` and `POST` when accessing the resource
+            .allow_methods([Method::GET])
+            // allow requests from any origin
+            .allow_origin(Any);
+        let auth_router = Router::new()
+            .route(
+                "/.well-known/oauth-protected-resource",
+                get(protected_resource),
+            )
+            .with_state(self.clone())
+            .layer(cors);
+
+        // Merge with MCP server routes
+        Router::new()
+            .merge(auth_router)
+            .merge(router.layer(axum::middleware::from_fn_with_state(
+                self.clone(),
+                oauth_validate,
+            )))
+    }
+}
+
+/// Helper for returning internal server errors
+macro_rules! internal_error {
+    ($msg:literal, $e:ident) => {{
+        error!("INTERNAL ERROR (This should not happen). {}: {}", $msg, $e);
+
+        (StatusCode::INTERNAL_SERVER_ERROR, Default::default())
+    }};
+}
+
+async fn oauth_validate(
+    State(auth_config): State<Config>,
+    headers: HeaderMap,
+    mut request: Request,
+    next: Next,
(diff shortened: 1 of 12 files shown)


crates/apollo-mcp-server/src/auth.rs
```diff
@@ -0,0 +1,235 @@
...
+use reqwest::header::{AUTHORIZATION, WWW_AUTHENTICATE};
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+use tower_http::cors::{Any, CorsLayer};
+use tracing::{error, info, warn};
+use url::Url;
+
+/// Auth configuration options
+#[derive(Debug, Clone, Deserialize, JsonSchema)]
+pub struct Config {
```
alocay, line 32: I'm guessing this is not under `runtime` since it's used by the lib? Should there also be a reference to this under `src/runtime/config.rs`? I figured we would want to have auth configs in the config file.

crates/apollo-mcp-server/src/auth.rs
```diff
@@ -0,0 +1,235 @@
...
+        pub aud: String,
+        pub sub: String,
+    }
+
+    // The token should be in the form 'Bearer ...'
+    let bearer_prefix = "bearer ";
+    let Some((bearer, jwt)) = token.split_at_checked(bearer_prefix.len()) else {
+        return Ok(false);
+    };
+    if bearer.to_lowercase() != bearer_prefix {
```
alocay: nit: would it be any better to use `eq_ignore_ascii_case` here instead?

crates/apollo-mcp-server/src/auth.rs
alocay: Oh nevermind, I see now that it's part of the server transport config.

crates/apollo-mcp-server/src/server/states/running.rs
```diff
@@ -223,10 +226,27 @@ impl ServerHandler for Running {
...
+                // Optionally extract the validated token and propagate it to upstream servers
+                // if found
+                let mut headers = self.headers.clone();
+                if let Some(token) = context.extensions.get::<ValidToken>() {
+                    headers.insert(
+                        AUTHORIZATION,
+                        HeaderValue::try_from(token.clone().read()).map_err(|e| {
+                            error!("Could not create auth token from validated token: {e}");
+                            McpError::new(
+                                ErrorCode::INTERNAL_ERROR,
```
swcollard: Odd that the MCP error codes don't have an Auth error for this situation.

crates/apollo-mcp-server/src/auth.rs
```diff
@@ -0,0 +1,235 @@
...
+use tower_http::cors::{Any, CorsLayer};
+use tracing::{error, info, warn};
+use url::Url;
+
+/// Auth configuration options
+#[derive(Debug, Clone, Deserialize, JsonSchema)]
+pub struct Config {
+    /// Disable authentication
+    #[serde(default)]
+    pub disable: bool,
```
swcollard: nit, most of our configs use `enabled` instead of `disable` 

It might be good to keep that consistent, unless this is a direct mapping to how the MCP spec configures Auth

crates/apollo-mcp-server/src/auth.rs
```diff
@@ -0,0 +1,235 @@
...
+            fn from(value: Config) -> Self {
+                Self {
+                    resource: value.resource,
+                    authorization_servers: value.servers,
+                    bearer_methods_supported: vec!["header".to_string()],
+                    scopes_supported: ["profile", "email", "phone", "mcp-flow"]
+                        .into_iter()
+                        .map(str::to_string)
+                        .collect(),
+                    resource_documentation: "TODO".to_string(),
```
swcollard: nit, these still look like placeholder values

crates/apollo-mcp-server/src/auth.rs
nicholascioli: I made this way more type safe, so hopefully this shouldn't even be an issue anymore

crates/apollo-mcp-server/src/auth.rs
nicholascioli: Oh, that makes sense. I just removed this flag entirely and made the auth section of the transport config optional

crates/apollo-mcp-server/src/auth.rs
nicholascioli: These definitely are. I'll add them as configurable options.

crates/apollo-mcp-server/src/server/states/running.rs
nicholascioli: The issue is that this error is if (somehow) the bearer token verified by the auth handler cannot be turned back into an auth token. I've messaged this to no longer be fallible.

swcollard (approved):

crates/apollo-mcp-server/src/auth.rs
```diff
@@ -0,0 +1,178 @@
...
+
+    /// Link to documentation related to the protected resource
+    pub resource_documentation: Option<Url>,
+
+    /// Supported OAuth scopes by this resource server
+    pub scopes: Vec<String>,
+}
+
+impl Config {
+    pub fn enable_middleware(&self, router: axum::Router) -> axum::Router {
```
DaleSeo:
```suggestion
    pub fn enable_middleware(&self, router: Router) -> Router {
```

`axum::Router` was already brought into the scope:

https://github.com/apollographql/apollo-mcp-server/blob/32ba4f3e1a1a879e1098619d4652834e31b69274/crates/apollo-mcp-server/src/auth.rs#L1-L2

crates/apollo-mcp-server/src/auth.rs
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    let Ok(header) = decode_header(jwt) else {
+        return false;
+    };
+    let Some(ref key_id) = header.kid else {
+        return false;
+    };
+
+    for server in &auth_config.servers {
```
DaleSeo: I’m curious if supporting multiple upstream OAuth servers is a common pattern in practice.

crates/apollo-mcp-server/src/auth/www_authenticate.rs
```diff
@@ -0,0 +1,42 @@
...
+
+    fn encode<E: Extend<http::HeaderValue>>(&self, values: &mut E) {
+        let encoded = match &self {
+            WwwAuthenticate::Bearer { resource_metadata } => format!(
+                r#"Bearer resource_metadata="{}""#,
+                resource_metadata.as_str()
+            ),
+        };
+
+        // TODO: This shouldn't error, but it can so we might need to do something else here
```
DaleSeo, line 36: How about we create backlog items for the TODOs and reference the ticket numbers here. This will make it easier to prioritize and track the follow-up work.

crates/apollo-mcp-server/src/auth/www_authenticate.rs
```diff
@@ -0,0 +1,42 @@
...
+    fn name() -> &'static http::HeaderName {
+        &WWW_AUTHENTICATE
+    }
+
+    fn decode<'i, I>(_values: &mut I) -> Result<Self, headers::Error>
+    where
+        Self: Sized,
+        I: Iterator<Item = &'i http::HeaderValue>,
+    {
+        // We don't care about decoding, so we do nothing here.
```
DaleSeo, line 24: Why we only care about encoding?

crates/apollo-mcp-server/src/lib.rs
```diff
@@ -1,3 +1,4 @@
+pub mod auth;
```
DaleSeo, line 1: This makes me feel better about `pub mod health` that I did. 😂

crates/apollo-mcp-server/src/server.rs
```diff
@@ -52,6 +53,10 @@ pub enum Transport {
     /// Note: This is deprecated in favor of HTTP streams.
     #[serde(rename = "sse")]
     SSE {
+        /// Authentication configuration
+        #[serde(default)]
+        auth: Option<auth::Config>,
+
```
DaleSeo, line 59: This reminds me that Matt once told me not to worry about supporting health checks for the SSE transport since we do not want to spend time on a deprecated transport. Maybe that same advice applies here for OAuth support?

crates/apollo-mcp-server/src/server/states/running.rs
```diff
@@ -180,7 +183,7 @@ impl ServerHandler for Running {
     async fn call_tool(
         &self,
         request: CallToolRequestParam,
-        _context: RequestContext<RoleServer>,
+        context: RequestContext<RoleServer>,
```
DaleSeo, line 186: I'm a bit confused about the terminology here. We use "state" above and "context" here. Are these meant to refer to the same concept, just with different names?

crates/apollo-mcp-server/src/server/states/starting.rs
```diff
@@ -150,8 +151,23 @@ impl Starting {
             health_check: health_check.clone(),
         };
 
+        // Helper to enable auth
+        macro_rules! with_auth {
```
DaleSeo, line 155: Oh, this is my first time seeing how a macro is defined in Rust! 💡

DaleSeo:
> The code looks good to me, but I am currently trying to test it locally before approving

@nicholascioli I’m running into trouble testing it locally as well. I’d like to try it out with an IDP provider like the one you used in the MVP demo (was it Keycloak?). Would you be able to add some testing instructions to the PR description so we can follow along?

DaleSeo:
Should we add documentation for our OAuth support? I think OAuth is notoriously difficult for most folks. 😅

crates/apollo-mcp-server/src/auth/www_authenticate.rs
nicholascioli: We only ever return this header when a client is unauthorized so that they can then fetch the needed metadata for auth delegation. This struct exists so that we can use a strongly-typed structure instead of raw text.

crates/apollo-mcp-server/src/server.rs
nicholascioli: I can remove it from here and then simplify the SSE code back to what it was, if you feel that makes sense.

crates/apollo-mcp-server/src/auth.rs
nicholascioli: This is just what the spec defines as being needed. We can diverge here if you prefer, but I figured that most people will probably just configure this to only have one upstream IdP.

crates/apollo-mcp-server/src/server/states/running.rs
nicholascioli: I'm not sure that I know what you mean by `state above`. The context here is access to the context of the MCP request, which embeds any state defined by axum.

DaleSeo: I was referring to this:

https://github.com/apollographql/apollo-mcp-server/blob/673cb43052689d532814d21a65ed51af3b133b2e/crates/apollo-mcp-server/src/auth.rs#L84

crates/apollo-mcp-server/src/server.rs
DaleSeo: This should make maintenance easier. It could also help us encourage users to switch to Streamable HTTP with OAuth and move away from SSE. I like it, but what do you all think? @swcollard @alocay @evansamek

nicholascioli: I can retroactively remove these changes once we come to a decision!

crates/apollo-mcp-server/src/server/states/running.rs
nicholascioli: Oh, yeah that's axum state which differs (but not really) from an RMCP context.

PR #212, opened 2025-07-29
fix: ensuring properties field exists even when there are no input args

alocay:
Due to `schemas::ObjectValidation` using the attribute `#[serde(skip_serializing_if = "Map::is_empty")]`, when there are no input arguments for an operation the resulting input schema JSON will omit the `properties` field. Addressing this by running a post `serde_json::to_value()` check to ensure that the properties field exists. If not, it will be populated with an empty map. 

Added a simple unit test for this case.


swcollard (approved):
I think technically the missing `properties` still matches the spec, but it can't hurt to have it be empty like you have here.

PR #213, opened 2025-07-29
Fix parsing headers from header file

tylerscoville:
I'm seeing an error like: `Error: invalid type: found string "X-Header", expected a borrowed string for key "default.headers.X-Header" in config.yaml YAML file` when running the mcp server with a config file with the headers key:
```yaml
...
headers:
   X-Header: Foo
...
```

This updates the while loop to borrow the string, avoiding the error. I'm sure there is a better way to handle this, but I wanted to take a quick poke at fixing it.


swcollard (approved):
Thanks for the fix!

PR #211, opened 2025-07-29
Handle environment variables when no config file is provided

DaleSeo:
When running the MCP server without a config file, environment variables like `APOLLO_GRAPH_REF` and `APOLLO_KEY` are ignored. This results in the server failing with the message "Missing environment variable: APOLLO_GRAPH_REF," even when the variables are properly set.

The root cause appears to be the configuration loading logic, which has two separate paths:
-  **With config file**: `read_config(path)` processes both the YAML file and environment variables.
-  **Without config file**: `Config::default()` only uses defaults, ignoring environment variables.

The processing of environment variables, including the mapping of `APOLLO_GRAPH_REF` to `GRAPHOS:APOLLO_GRAPH_REF`, occurs only in the `read_config()` function.

To resolve this issue, I added the `read_config_from_env()` function, which processes environment variables without requiring a config file. I then updated the `main` function to utilize environment variable processing regardless of the presence of a config file.

## Testing

Before:

```bash
$ APOLLO_GRAPH_REF=dale@demo APOLLO_KEY=🔑 cargo run
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.20s
     Running `target/debug/apollo-mcp-server`
2025-07-29T13:55:26.616148Z  INFO Apollo MCP Server v0.6.0 // (c) Apollo Graph, Inc. // Licensed under MIT
Error: Missing environment variable: APOLLO_GRAPH_REF
```

After:

```bash
$ APOLLO_GRAPH_REF=dale@demo APOLLO_KEY=🔑 cargo run
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.55s
     Running `target/debug/apollo-mcp-server`
2025-07-29T14:33:39.756760Z  INFO Apollo MCP Server v0.6.0 // (c) Apollo Graph, Inc. // Licensed under MIT
2025-07-29T14:33:39.756884Z  WARN No operations specified, falling back to the default collection in dale@demo
2025-07-29T14:33:40.131890Z  INFO Tool GoodQuery loaded with a character count of 188. Estimated tokens: 47
2025-07-29T14:33:40.134403Z  INFO Tool BadQuery loaded with a character count of 187. Estimated tokens: 46
2025-07-29T14:33:40.134492Z  INFO Starting MCP server in stdio mode
```

PR #217, opened 2025-07-30
Remove APOLLO_GRAPH_REF requirement for local configurations

### LINKED ISSUE #216: APOLLO_GRAPH_REF environment variable still required even with config file in v0.6.1
## Summary
  Despite the fix in PR #211 that was supposed to resolve "Missing environment variable: APOLLO_GRAPH_REF" errors, the issue persists when using a configuration file with apollo-mcp-server v0.6.1.

  ## Environment
  - **apollo-mcp-server version**: v0.6.1
  - **Platform**: Docker (ghcr.io/apollographql/apollo-mcp-server:v0.6.1)
  - **Configuration**: YAML config file with local schema and operations

  ## Expected Behavior
  According to PR #211, the server should start successfully when APOLLO_GRAPH_REF is provided via environment variable or when using a config file with local schema (without requiring GraphOS connection).

  ## Actual Behavior
  The server fails to start with the error:
  Error: Missing environment variable: APOLLO_GRAPH_REF
  [2025-07-30T01:51:47.042540Z] INFO Apollo MCP Server v0.6.1 // (c) Apollo Graph, Inc. // Licensed under MIT

  ## Configuration File
  ```yaml
  endpoint: "https://example.com/graphql"
  transport:
    type: http
  listen:
    address: "0.0.0.0"
    port: 5000
  schema:
    source: local
    path: "schema.graphql"
  operations:
    source: local
    paths:
      - "operations/queries"
    persisted_queries:
      source: local
      path: "persisted_queries/apollo.json"
  custom_scalars: "custom_scalars_config.json"
  introspection:
    execute:
      enabled: true
    introspect:
      enabled: true
    search:
      enabled: true

  Steps to Reproduce

  1. Create the above configuration file as config.yaml
  2. Run: docker run --rm ghcr.io/apollographql/apollo-mcp-server:v0.6.1 config.yaml
  3. Observe the error

  Additional Testing

  Test 1: With APOLLO_GRAPH_REF environment variable

  docker run --rm -e APOLLO_GRAPH_REF="test@dev" ghcr.io/apollographql/apollo-mcp-server:v0.6.1 config.yaml
  Result: Same error

  Test 2: With graphos section in config

  graphos:
    apollo_graph_ref: "test@dev"
  Result: Same error

  Test 3: Without config file (environment variables only)

  docker run --rm -e APOLLO_GRAPH_REF="test@dev" -e APOLLO_KEY="dummy" ghcr.io/apollographql/apollo-mcp-server:v0.6.1 apollo-mcp-server
  Result: Different error (schema-related), suggesting APOLLO_GRAPH_REF is recognized

  Analysis

  The fix in PR #211 appears to work for environment-variable-only configurations but not when using a configuration file. The issue seems to be that when a config file is provided, the server still requires APOLLO_GRAPH_REF environment variable regardless of the config file content.

  Use Case

  We're running apollo-mcp-server as a proxy to an existing GraphQL API with local schema and operations files, without needing GraphOS integration. This worked in v0.5.2 but broke in v0.6.0+ due to the APOLLO_GRAPH_REF requirement.

  Workaround

  Currently considering reverting to v0.5.2 until this issue is resolved.

  Related Issues

  - PR #211: Fix environment variable handling
  - Original issue that PR #211 was supposed to fix

DaleSeo:
Fixes #216 

Even with the fix in PR #211, the MCP server still requires `APOLLO_GRAPH_REF` when using local schema and operations files. This stops users from running the server without GraphOS integration.

It seems this issue is due to local operation configurations incorrectly falling back to GraphOS logic. I separated the pattern matching to handle local configurations independently from the inference logic.

I was also able to reproduce the same issue when `enable_explorer` was true, even by default, which unconditionally required `APOLLO_GRAPH_REF`. I updated the explorer logic to handle the missing `APOLLO_GRAPH_REF` more gracefully.

## Testing

testing now...

### DIFF
--- a/crates/apollo-mcp-server/src/main.rs
+++ b/crates/apollo-mcp-server/src/main.rs
@@ -103,9 +103,23 @@ async fn main() -> anyhow::Result<()> {
             OperationSource::from(ManifestSource::Uplink(config.graphos.uplink_config()?))
         }
 
+        // Handle empty local paths separately from infer logic
+        runtime::OperationSource::Local { .. } => {
+            // This is the case where Local was specified but paths were empty
+            // We should not fall back to GraphOS for explicitly local configurations
+            if config.introspection.any_enabled() {
+                warn!(
+                    "Local operations specified but no valid paths found, falling back to introspection"
+                );
+                OperationSource::None
+            } else {
+                anyhow::bail!(ServerError::NoOperations)
+            }
+        }
+
         // TODO: Inference requires many different combinations and preferences
         // TODO: We should maybe make this more explicit.
-        runtime::OperationSource::Local { .. } | runtime::OperationSource::Infer => {
+        runtime::OperationSource::Infer => {
             if config.introspection.any_enabled() {
                 warn!("No operations specified, falling back to introspection");
                 OperationSource::None
@@ -127,8 +141,8 @@ async fn main() -> anyhow::Result<()> {
     let explorer_graph_ref = config
         .overrides
         .enable_explorer
-        .then(|| config.graphos.graph_ref())
-        .transpose()?;
+        .then(|| config.graphos.graph_ref().ok())
+        .flatten();
 
     Ok(Server::builder()
         .transport(config.transport)


crates/apollo-mcp-server/src/main.rs
```diff
@@ -127,8 +141,8 @@ async fn main() -> anyhow::Result<()> {
     let explorer_graph_ref = config
         .overrides
         .enable_explorer
-        .then(|| config.graphos.graph_ref())
-        .transpose()?;
+        .then(|| config.graphos.graph_ref().ok())
```
alocay, line 144: Just curious, why the change from `.transpose()` to using `ok()`? The `transpose` along with the `?` operator seems to have the same effect as using `ok()` with the `flatten()`.

DaleSeo:
I noticed that I pointed at the wrong spot while testing. 😅
It looks like there's an issue with the Docker image. I'll bring it up with the team.

PR #220, opened 2025-07-30
fix: Add custom deserializer to handle APOLLO_UPLINK_ENDPOINTS environment variable parsing

swcollard:
The APOLLO_UPLINK_ENDPOINTS environment variables [has historically been a comma separated list of URL strings.](https://github.com/apollographql/router/blob/5b4d688a2ce52e83e1ffe861e98e8ecb87141e9f/README.md?plain=1#L48-L49) The move to yaml configuration allows us to more directly define the endpoints as a Vec. This change introduces a custom deserializer for the `apollo_uplink_endpoints` config field that can handle both the environment variable comma separated string, and the yaml-based list.

### DIFF
--- a/crates/apollo-mcp-server/src/runtime/graphos.rs
+++ b/crates/apollo-mcp-server/src/runtime/graphos.rs
@@ -7,7 +7,7 @@ use apollo_mcp_registry::{
 use apollo_mcp_server::errors::ServerError;
 use schemars::JsonSchema;
 use serde::de::Error;
-use serde::{Deserialize, Deserializer};
+use serde::{Deserialize, Deserializer, Serialize};
 use url::Url;
 
 const APOLLO_GRAPH_REF_ENV: &str = "APOLLO_GRAPH_REF";
@@ -39,21 +39,23 @@ where
 
 /// Credentials to use with GraphOS
 #[derive(Debug, Deserialize, Default, JsonSchema)]
+#[cfg_attr(test, derive(Serialize))]
 #[serde(default)]
 pub struct GraphOSConfig {
     /// The apollo key
     #[schemars(with = "Option<String>")]
-    pub apollo_key: Option<SecretString>,
+    #[cfg_attr(test, serde(skip_serializing))]
+    apollo_key: Option<SecretString>,
 
     /// The graph reference
-    pub apollo_graph_ref: Option<String>,
+    apollo_graph_ref: Option<String>,
 
     /// The URL to use for Apollo's registry
-    pub apollo_registry_url: Option<Url>,
+    apollo_registry_url: Option<Url>,
 
     /// List of uplink URL overrides
     #[serde(deserialize_with = "apollo_uplink_endpoints_deserializer")]
-    pub apollo_uplink_endpoints: Vec<Url>,
+    apollo_uplink_endpoints: Vec<Url>,
 }
 
 impl GraphOSConfig {
--- a/crates/apollo-mcp-server/src/runtime/graphos.rs
+++ b/crates/apollo-mcp-server/src/runtime/graphos.rs
@@ -6,11 +6,32 @@ use apollo_mcp_registry::{
 };
 use apollo_mcp_server::errors::ServerError;
 use schemars::JsonSchema;
-use serde::Deserialize;
+use serde::de::Error;
+use serde::{Deserialize, Deserializer};
 use url::Url;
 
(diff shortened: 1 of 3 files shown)


crates/apollo-mcp-server/src/runtime/graphos.rs
```diff
@@ -6,28 +6,50 @@ use apollo_mcp_registry::{
...
+    enum UrlListOrString {
+        List(Vec<Url>),
+        String(String),
+    }
+
+    match UrlListOrString::deserialize(deserializer)? {
+        UrlListOrString::List(urls) => Ok(urls),
+        UrlListOrString::String(s) => s
+            .split(',')
+            .map(|v| Url::parse(v.trim()).map_err(D::Error::custom))
```
nicholascioli: Do we want to add context to the error here that it failed while trying to decode the uplink list?

crates/apollo-mcp-server/src/runtime/graphos.rs
```diff
@@ -6,28 +6,50 @@ use apollo_mcp_registry::{
...
+    pub apollo_graph_ref: Option<String>,
 
     /// The URL to use for Apollo's registry
-    apollo_registry_url: Option<Url>,
+    pub apollo_registry_url: Option<Url>,
 
     /// List of uplink URL overrides
-    apollo_uplink_endpoints: Vec<Url>,
+    #[serde(deserialize_with = "apollo_uplink_endpoints_deserializer")]
+    pub apollo_uplink_endpoints: Vec<Url>,
```
nicholascioli: Why did visibility change here?

nicholascioli (changes requested):
Just left a quick comment about visibility. Looks great otherwise!

crates/apollo-mcp-server/src/runtime/graphos.rs
swcollard: This was changed so I could read the field in my unit test. Not sure if there is a better way to do that?

crates/apollo-mcp-server/src/runtime/graphos.rs
swcollard: Yes, will add a message to the error

crates/apollo-mcp-server/src/runtime/graphos.rs
```diff
@@ -6,28 +6,54 @@ use apollo_mcp_registry::{
...
+    match UrlListOrString::deserialize(deserializer)? {
+        UrlListOrString::List(urls) => Ok(urls),
+        UrlListOrString::String(s) => s
+            .split(',')
+            .map(|v| {
+                Url::parse(v.trim()).map_err(|e| {
+                    D::Error::custom(format!("Could not parse uplink endpoint URL: {e}"))
+                })
+            })
+            .collect(),
```
DaleSeo, line 39: What happens if we get an empty string? It seems like it would turn into an empty array. I guess this is intended since `GraphOSConfig.apollo_uplink_endpoints` is not optional?

crates/apollo-mcp-server/src/runtime/graphos.rs
nicholascioli: You could add the following to the derive list for `GraphOSConfig` and then use snapshot testing to make sure that the fields match up as expected.

```rust
#[derive(Debug, Deserialize, Default, JsonSchema)]
// Add the line below
#[cfg_attr(test, derive(Serialize))]
#[serde(default)
pub struct GraphOSConfig {
  //
}
```

Snapshot testing also supports `Debug` derived objects. Some examples of how the testing works are shown below:

```rust
// In your test
// For using `Serialize` snapshots
insta::assert_yaml_snapshot(expected, @"");

// For using the `Debug` derived snapshots
insta::assert_debug_snapshot(expected, @"");
```

In either case, after you run `cargo insta` it should let you know that there are some new snapshots which you can review with `cargo insta review`.

crates/apollo-mcp-server/src/runtime/graphos.rs
swcollard: For an empty string, it returns the error: 
```
Error: Could not parse uplink endpoint URL: relative URL without a base for key "GRAPHOS.APOLLO_UPLINK_ENDPOINTS" in `APOLLO_` environment variable(s)
```

crates/apollo-mcp-server/src/runtime/graphos.rs
swcollard: Updated the tests to use an insta snapshot

PR #223, opened 2025-07-31
docs: Provide example config file for documentation

jcaudle:
In order to make usage of the new configuration file options easier, this commit adds an example configuration file. It also adds a few notes about proper usage of some options.

### DIFF
--- a/docs/source/command-reference.mdx
+++ b/docs/source/command-reference.mdx
@@ -87,12 +87,13 @@ iwr 'https://mcp.apollo.dev/download/win/v0.6.1' | iex
 
 ## Usage
 
-Configure the Apollo MCP server with a YAML configuration file. The path to this file is the only argument.
-A configuration file is not required, but if one is not provided environment variables for your Apollo Graph Reference and Apollo Key are needed.
+Configure the Apollo MCP server with a YAML configuration file.
 
 ```sh showLineNumbers=false
 apollo-mcp-server [OPTIONS] <PATH/TO/CONFIG/FILE>
 ```
+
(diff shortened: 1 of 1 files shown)


docs/source/command-reference.mdx
```diff
@@ -100,6 +100,25 @@ apollo-mcp-server [OPTIONS] <PATH/TO/CONFIG/FILE>
 | `-h, --help`    | Print help information    |
 | `-V, --version` | Print version information |
 
+### Example config file
```
mabuyo, line 105: Is this the minimum configuration needed to get up and running with MCP quickly?

docs/source/command-reference.mdx
```diff
@@ -100,6 +100,25 @@ apollo-mcp-server [OPTIONS] <PATH/TO/CONFIG/FILE>
...
+
+Detailed options for your configuration file are below, but this is an example configuration file:
```
mabuyo: Could we provide a little more explanation on what this config file is doing? For example, it uses local operations, enables introspection, connects to a graph (? not sure what `graphos` does actually).

Here are some example docs we can use as well for style consistency:

- https://www.apollographql.com/docs/rover/commands/supergraphs#yaml-configuration-file (I like how this one explains the example).
- https://www.apollographql.com/docs/graphos/routing/configuration/yaml#example-yaml-with-all-properties (is it possible to list out all the properties like this?)

jcaudle:
@mabuyo I think I addressed some of the questions you had in my changes here. I don't know that putting an exhaustive list of configuration options in the example file would make sense, but I could reformat the tables below or provide example YAML for each of the subkeys, if you think it would better match stylistically with other documentation.

mabuyo (approved):
Thanks @jcaudle - LGTM!

PR #225, opened 2025-08-01
Add documentation for validate tool and health check support

DaleSeo:
This PR adds documentation for two recently implemented features.

### DIFF
--- a/crates/apollo-mcp-server/src/introspection/tools/execute.rs
+++ b/crates/apollo-mcp-server/src/introspection/tools/execute.rs
@@ -36,7 +36,7 @@ impl Execute {
             mutation_mode,
             tool: Tool::new(
                 EXECUTE_TOOL_NAME,
-                "Execute a GraphQL operation. Use the `introspect` tool to get information about the GraphQL schema. Always use the schema to create operations - do not try arbitrary operations. If available, first use the `validate` tool to validate operations. DO NOT try to execute introspection queries.",
+                "Execute a GraphQL operation. Use the `introspect` tool first to get information about the GraphQL schema. Always use the schema to create operations - do not try arbitrary operations. If available, first use the `validate` tool to validate operations. DO NOT try to execute introspection queries.",
                 schema_from_type!(Input),
             ),
         }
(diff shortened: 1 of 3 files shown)


crates/apollo-mcp-server/src/introspection/tools/execute.rs
```diff
@@ -36,7 +36,7 @@ impl Execute {
             mutation_mode,
             tool: Tool::new(
                 EXECUTE_TOOL_NAME,
-                "Execute a GraphQL operation. Use the `introspect` tool to get information about the GraphQL schema. Always use the schema to create operations - do not try arbitrary operations. If available, first use the `validate` tool to validate operations. DO NOT try to execute introspection queries.",
+                "Execute a GraphQL operation. Use the `introspect` tool first to get information about the GraphQL schema. Always use the schema to create operations - do not try arbitrary operations. If available, first use the `validate` tool to validate operations. DO NOT try to execute introspection queries.",
```
swcollard: This could get really complicated, but I wonder if we need to pass more configuration into the tools so that we can make descriptions that are aware of what else is enabled.

For example, if we have both search and introspection enabled, how can we better tell the LLM which one to use and for what purpose.

Same for when the validation tool is enabled, can we make it more explicit in the execute tool to use validate first?

DaleSeo: Oh, now I realize the intent! When I was testing 4 inspection tools at the same time on Claude Desktop, adding this was helpful, but that might've been just a coincidence. Let me change this back.

docs/source/guides/index.mdx
```diff
@@ -328,6 +325,54 @@ npx @modelcontextprotocol/inspector
...
+
+The health check provides different responses based on query parameters:
+
+| Endpoint            | Description        | Response                                                         |
+| :------------------ | :----------------- | :--------------------------------------------------------------- |
+| `GET /health`       | Basic health check | Always returns `{"status": "UP"}`                                |
+| `GET /health?live`  | Liveness check     | Returns `{"status": "UP"}` if server is alive                    |
+| `GET /health?ready` | Readiness check    | Returns `{"status": "UP"}` if server is ready to handle requests |
+
+### Readiness tracking
```
swcollard: Not blocking, but I was comparing what we have to the kubernetes docs that refer to these as "Probes" 

https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/

swcollard (approved):

PR #226, opened 2025-08-04
docs: adding logging.path and logging.rotation cmd reference docs

alocay:
Updating config reference docs for information on newly added `logging.path` and `logging.rotation` options.

### DIFF
--- a/docs/source/command-reference.mdx
+++ b/docs/source/command-reference.mdx
@@ -192,9 +192,11 @@ These fields are under the top-level `introspection` key. Learn more about the M
 
 These fields are under the top-level `logging` key.
 
-| Option  | Type                                                | Default  | Description                     |
-| :------ | :-------------------------------------------------- | :------- | :------------------------------ |
-| `level` | `oneOf ["trace", "debug", "info", "warn", "error"]` | `"info"` | The minimum log level to record |
+| Option     | Type                                                | Default    | Description                                                                       |
+|:-----------|:----------------------------------------------------|:-----------|:----------------------------------------------------------------------------------|
+| `level`    | `oneOf ["trace", "debug", "info", "warn", "error"]` | `"info"`   | The minimum log level to record                                                   |
+| `path`     | `FilePath`                                          |            | An output file path for logging. If not provided logging outputs to stdio/stderr. |
+| `rotation` | `oneOf ["minutely", "hourly", "daily", "never"]`    | `"hourly"` | The log file rotation interval (if file logging is used)                          |
 
 #### Operation source configuration
 


docs/source/command-reference.mdx
```diff
@@ -192,9 +192,11 @@ These fields are under the top-level `introspection` key. Learn more about the M
 
 These fields are under the top-level `logging` key.
 
-| Option  | Type                                                | Default  | Description                     |
-| :------ | :-------------------------------------------------- | :------- | :------------------------------ |
-| `level` | `oneOf ["trace", "debug", "info", "warn", "error"]` | `"info"` | The minimum log level to record |
+| Option     | Type                                                | Default    | Description                                                                       |
+|:-----------|:----------------------------------------------------|:-----------|:----------------------------------------------------------------------------------|
+| `level`    | `oneOf ["trace", "debug", "info", "warn", "error"]` | `"info"`   | The minimum log level to record                                                   |
+| `path`     | `FilePath`                                          |            | An output file path for logging. If not provided logging outputs to stdio/stderr. |
```
mabuyo, line 198: Based on the description, should the default column be set to `stdio/stderr`?

docs/source/command-reference.mdx
```diff
@@ -192,9 +192,11 @@ These fields are under the top-level `introspection` key. Learn more about the M
...
+| `rotation` | `oneOf ["minutely", "hourly", "daily", "never"]`    | `"hourly"` | The log file rotation interval (if file logging is used)                          |
```
mabuyo, line 199: The term `rotation` is unclear here for me, could you clarify what this is referring to?

alocay: @mabuyo Sure. So with file logging, a rotation can be set to indicate how often it should spin up a new file for logging. If it's set to hourly for example it will create a new file every hour and start logging there. If it's set to never it will use a single constantly growing file. I believe the term `rotation` is an older term but essentially just indicates the interval at which a new file should be created/rolled over to.

docs/source/command-reference.mdx
alocay: Yeah we could although technically the default for this config value is nothing. Internally the server translates no value for the file path to mean to route all logs to stderr (send to the standard error/diagnostics stream). So if we don't mind that it doesn't show the actual default of the config value, we can set that. I just don't want some users to think that `stderr` is valid option to set this value to in the config.

docs/source/command-reference.mdx
mabuyo: That makes sense, thanks! I was thinking of adding a bit more to the description but it seems like a standard term in this specific process so we can probably leave as-is

docs/source/command-reference.mdx
mabuyo: Right, let's leave it blank then! Since `stdio` isn't a _file path_, that makes sense to me

mabuyo (approved):
Thank you for documenting! 🙌🏽

PR #232, opened 2025-08-07
Docs: Auth guide with Auth0 and MCP Inspector example

mabuyo:
- Extracts auth guide into its own page in the sidebar
- Add auth guide for Auth0 and MCP Inspector example


docs/source/guides/auth.mdx
```diff
@@ -0,0 +1,164 @@
...
+            thumbnail
+          }
+        }
+      }
+    }
+    ```
+
+1. You should see an `UNAUTHENTICATED` error, which means the router is enforcing authentication. Great! Now you can proceed to test the authentication with an MCP client.
+
+### Make requests with MCP Inspector
```
mabuyo: Somewhere in this flow we need to add the correct callback URL in Auth0... so maybe it's better to actually do the "Guided OAuth Flow".

docs/source/guides/auth.mdx
```diff
@@ -0,0 +1,164 @@
...
+To follow along with this example, you need to:
+- Clone the [Apollo MCP Server repo](https://github.com/apollographql/apollo-mcp-server)
+- Install [Apollo Rover CLI](/rover/getting-started) v0.35 or later
+
+### Set up the Auth0 Identity Provider
+
+#### Create the Auth0 API
+
+1. Create an Auth0 API by navigating to `Applications -> APIs`. Click **Create API** and fill in the fields necessary. You'll need to note the `Identifier` you choose, as it will be used in the MCP server configuration later on as the `audience` property.
+1. Navigate to your dashboard **Settings**.
```
DaleSeo, line 52: I'm afraid users will get lost right from the start here because the Auth0 dashboard can be overwhelming. It might help to spell out the navigation steps, for example: "Navigate to `Dashboard -> Settings -> Advanced`."

docs/source/guides/auth.mdx
```diff
@@ -0,0 +1,164 @@
...
+    1. Toggle on **OIDC Dynamic Application Registration** to enable [dynamic client registration](https://auth0.com/docs/get-started/applications/dynamic-client-registration#enable-dynamic-client-registration).
+    1. Toggle on **Enable Application Connections**.
+    1. Under **API Authorization Settings**, set the **Default Audience** to the `Identifier` you chose when creating the API.
```
DaleSeo: Someone might be looking for this on the same tab. 😅 It might help to mention that they need to switch from the Advanced tab to the General tab.

docs/source/guides/auth.mdx
```diff
@@ -0,0 +1,164 @@
...
+    transport:
+      type: streamable_http
+      auth:
+        servers:
+          - https://<AUTH0 DOMAIN>
+        audiences:
+          - <AUTH0 DEFAULT AUDIENCE>
+        resource: http://localhost:5000
+        scopes:
+          - read:users
```
DaleSeo: How about highlighting these lines since the first two already exists?

DaleSeo (commented):
"User Guide > Guides" looks a bit awkward.

image

docs/source/guides/auth.mdx
```diff
@@ -0,0 +1,164 @@
...
+
+1. Start MCP Inspector by running:
+
+    ```sh
+    npx @modelcontextprotocol/inspector
+    ```
+
+    The browser should open automatically with a proxy auth token.
+
+1. In the MCP Inspector, select `Streamable HTTP` as the Transport Type and enter `http://127.0.0.1:5000/mcp` as the URL.
```
DaleSeo, line 197: Is it okay to use `http://127.0.0.1:5000/mcp` here while we had `resource: http://localhost:5000` in the config file above? I'm a bit fixated on this tiny detail because I’ve wasted a lot of time ignoring it before. 😂

docs/source/guides/auth.mdx
DaleSeo: What is the "Guided Token Refresh". You mean "Guided Token Refresh"?

2025-08-07 at 20 18 39

docs/source/guides/auth.mdx
DaleSeo: On second thought, I'd make this the first step, since the General tab is the default on the settings page and the order doesn’t matter.

docs/source/guides/auth.mdx
mabuyo: Ahh it changes text depending on whether or not you have a token there already. Without the token it says "Guided OAuth Flow"

Screenshot 2025-08-08 at 7 55 37 AM

mabuyo: And actually the step to add the correct callback URL in Auth0 is only needed if going through this Guided OAuth Flow process... if connecting directly, it should just work 🤔 (confirmed on my end)

docs/source/guides/auth.mdx
mabuyo: Good call, fixed!

docs/source/guides/auth.mdx
mabuyo: good catch, I'll make sure they're both consistent

mabuyo:
Still want to make larger docs organization changes, but opting to start small for now and just add a new "Authorization" page

DaleSeo (approved):

PR #235, opened 2025-08-08
docs: update curl command placeholders

DaleSeo:
Minor touches to the auth docs 😀

### DIFF
--- a/docs/source/guides/auth.mdx
+++ b/docs/source/guides/auth.mdx
@@ -64,11 +64,11 @@ The Auth0 Connection is the method clients use to authenticate. This guide uses 
 
 1. In your Auth0 dashboard, navigate to **Authentication** -> **Database**.
 1. Create the default **Username-Password-Authentication** connection. Click the **Try Connection** button to test it and set up a username and password for later.
-1. Back on your Auth0 dashboard, note the **Connection Identifier** at the top of the page. It should start with something like `con_`. Copy it into a scratch file. We'll refer to it later as `<CONNECTION ID>`.
+1. Back on your Auth0 dashboard, note the **Connection Identifier** at the top of the page. It should start with something like `con_`. Copy it into a temporary location. This guide refers to it as `<CONNECTION ID>`.
 1. Navigate to **Applications** -> **APIs** -> **Auth0 Management API**.
-1. Copy the domain name from the **Identifier** for this API into your scratch file. It should look something like `dev-123456.us.auth0.com`, where `dev-123456` is your Auth0 tenant ID. We'll refer to it as `<AUTH0 DOMAIN>`.
-1. Click the **API Explorer** tab. Copy the token value into your scratch file. We call it `<MGMT API ACCESS TOKEN>` in this guide.
-1. Run the following `curl` command to promote the connection to domain-level, replacing `<CONNECTION ID>`, `<AUTH0 DOMAIN>`, and `<MGMT API ACCESS TOKEN>` with the values you copied in the previous steps:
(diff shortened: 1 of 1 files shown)


jcaudle:
I jumped on @DaleSeo's PR to make some very closely related tweaks, but it expands what he intended a bit. If I need to pull this out for separate review, please let me know, but I think this will help readability.

docs/source/guides/auth.mdx
```diff
@@ -62,17 +62,18 @@ This guide uses [Auth0](https://auth0.com/) as the Identity Provider.
...
-1. Create the default **Username-Password-Authentication** connection. Click the **Try Connection** button to test it.
-1. Copy the **Connection Identifier** at the top of the page, you need it for the next step. It should start with something like `con_`.
+1. In your Auth0 dashboard, navigate to **Authentication** -> **Database**.
+1. Create the default **Username-Password-Authentication** connection. Click the **Try Connection** button to test it and set up a username and password for later.
+1. Back on your Auth0 dashboard, copy the **Connection Identifier** at the top of the page. You'll need it soon. It should start with something like `con_`.
 1. Navigate to **Applications** -> **APIs** -> **Auth0 Management API**.
-1. Click the **API Explorer** tab. Copy the token value, you also need it for the next step.
-1. Run the following `curl` command to promote the connection to domain-level, replacing `CONNECTION_ID` and `MGMT_API_ACCESS_TOKEN` with the values you copied in the previous steps:
+1. Copy the **Identifier** for this API. It should look something like `dev-123456.us.auth0.com`, where `dev-123456` is your Auth0 tenant ID.
```
mabuyo: @jcaudle Is "Identifier" the correct word here? It should be Auth0 domain, right?

jcaudle: The Identifier contains the domain, but it's not exactly the same thing. Here's a screenshot:

Screenshot 2025-08-11 at 15 38 37

I wanted to introduce the fact that this comes from your own Auth0 account and you can get it at this point, since there's nowhere it shows up (that I noticed) before this, but also, I think we need to change the placeholder in the command or the thing you need to copy here to make this correct...

DaleSeo: Technically, API identifiers on Auth0 includes `/api/v2` at the end, but in this context, we want users only grab the string before that, right?

DaleSeo:
@mabuyo @jcaudle It feels like we're asking users to copy too many times before they paste. I'm worried that their clipboard might get overwritten with whatever they copied last. 😅

2025-08-11 at 15 46 59

docs/source/guides/auth.mdx
mabuyo: OK so it looks like that's called the Auth0 Tenant ID https://auth0.com/docs/get-started/tenant-settings/find-your-tenant-name-or-tenant-id

I made reference to it in Step 2. Can we update to use that instead?

mabuyo:
@DaleSeo totally copy-paste overload.

@jcaudle we could try something like. "You will need three values to run the curl command: X, Y, Z" then bullet points for each outlining where to get those values?

docs/source/guides/auth.mdx
jcaudle: We need them to get the whole domain, so I changed the wording. I don't know if Auth0 has other regions where they run their servers, but just the Tenant ID isn't enough.

jcaudle:
I added some advice around saving things into a scratch file and the names we'll use throughout the guide for those values. Let me know what you think.

DaleSeo:
This looks good to me now but I can't approve my own PR 😂 @mabuyo

mabuyo (approved):
Thank you both for making these steps clearer!

PR #236, opened 2025-08-10
feat: Pass `remote-mcp` mcp-session-id header along to GraphQL request

damassi:
We're attempting to embed the apollo-mcp-server within a proxy setup for more flexible OAuth support but are running into an issue where we cannot correlate requests issued from the MCP client with outbound requests, through apollo-mcp-server, to our graphql endpoint. 

Inspecting the headers passed in from `remote-mcp`, there's an `mcp-session-id` header [that originates from the underlying `@modelcontextprotocol/sdk` library](https://github.com/modelcontextprotocol/typescript-sdk/blob/a1608a6513d18eb965266286904760f830de96fe/src/client/streamableHttp.ts#L182). Allowing this header through would not only solve our correlation issue, but provide a level of observability at the graphql service level for which clients performed what, and when. This feels generally useful and correct. So this PR updates things to allow for that.

### DIFF
--- a/crates/apollo-mcp-server/src/server/states/running.rs
+++ b/crates/apollo-mcp-server/src/server/states/running.rs
@@ -238,6 +238,7 @@ impl ServerHandler for Running {
                 // if found
                 let mut headers = self.headers.clone();
                 if let Some(axum_parts) = context.extensions.get::<axum::http::request::Parts>() {
+                    // Optionally extract the validated token and propagate it to upstream servers if present
                     if let Some(token) = axum_parts.extensions.get::<ValidToken>() {
                         headers.typed_insert(token.deref().clone());
                     }
--- a/crates/apollo-mcp-server/src/server/states/running.rs
+++ b/crates/apollo-mcp-server/src/server/states/running.rs
@@ -208,13 +208,21 @@ impl ServerHandler for Running {
                     .await
             }
             EXECUTE_TOOL_NAME => {
+                let mut headers = self.headers.clone();
+                if let Some(axum_parts) = context.extensions.get::<axum::http::request::Parts>() {
+                    // Forward the mcp-session-id header if present
+                    if let Some(session_id) = axum_parts.headers.get("mcp-session-id") {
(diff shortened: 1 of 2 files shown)


alocay:
LGTM although could you add a changeset file under [here](https://github.com/apollographql/apollo-mcp-server/tree/main/.changesets) ? The [README](https://github.com/apollographql/apollo-mcp-server/blob/main/.changesets/README.md) has some steps on how to generate it. 

Edit: I think you'll need to run `cargo fmt` as well to fix some linter issues.

damassi:
Updated 👍 Let me know if y'all need anything else. And thank you 🙏

crates/apollo-mcp-server/src/server/states/running.rs
```diff
@@ -229,10 +237,15 @@ impl ServerHandler for Running {
                 // Optionally extract the validated token and propagate it to upstream servers
                 // if found
                 let mut headers = self.headers.clone();
-                if let Some(axum_parts) = context.extensions.get::<axum::http::request::Parts>()
-                    && let Some(token) = axum_parts.extensions.get::<ValidToken>()
-                {
-                    headers.typed_insert(token.deref().clone());
+                if let Some(axum_parts) = context.extensions.get::<axum::http::request::Parts>() {
+                    if let Some(token) = axum_parts.extensions.get::<ValidToken>() {
```
DaleSeo, line 240: Moving the comment on L237-238 here might help prevent confusion for future readers.

```suggestion
                    // Optionally extract the validated token and propagate it to upstream servers if present
                    if let Some(token) = axum_parts.extensions.get::<ValidToken>() {
```

crates/apollo-mcp-server/src/server/states/running.rs
```diff
@@ -229,10 +237,16 @@ impl ServerHandler for Running {
                 // Optionally extract the validated token and propagate it to upstream servers
                 // if found
```
DaleSeo:
```suggestion
```

PR #239, opened 2025-08-11
it_rejects_different_audience is passing due to token expiry not audience mismatch

DaleSeo:
The test `it_rejects_different_audience` is passing because the JWT is expired, not because of an audience mismatch. This PR fixes this false positive, which could mask regressions in audience validation.

### DIFF
--- a/crates/apollo-mcp-server/src/introspection/tools/execute.rs
+++ b/crates/apollo-mcp-server/src/introspection/tools/execute.rs
@@ -247,24 +247,6 @@ mod tests {
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


crates/apollo-mcp-server/src/introspection/tools/execute.rs
```diff
@@ -246,24 +246,6 @@ mod tests {
         }
     }
 
-    #[test]
-    fn execute_query_err_with_subscription_when_mutation_mode_is_all() {
```
DaleSeo, line 250: Removed it because the above test already covers this scenario.

swcollard (approved):

PR #252, opened 2025-08-14
ci: adding a changeset check to github workflow

alocay:
Adding a github workflow to verify that a changeset file is created per change. If a changeset isn't found within the change list the check will fail and comment will be added to the PR reminding the author to create one. 

Ignoring changes that only involve more infra related configuration such as those under `.github`, `.cargo`, etc.

### DIFF
--- a/.github/workflows/verify-changeset.yml
+++ b/.github/workflows/verify-changeset.yml
@@ -11,15 +11,40 @@ on:
   workflow_dispatch:
 
 jobs:
-  check:
-    name: Run checks
+  verify-changeset:
+    name: Verify Changeset
     runs-on: ubuntu-24.04
     permissions:
       pull-requests: write
       contents: read
     steps:
+      - uses: actions/github-script@v7
+        id: filter
+        env:
+          IGNORE_DIRS: ".github/workflows/,.cargo/,.direnv/,.vscode/"
+        with:
+          script: |
+            const pr = context.payload.pull_request;
+            const files = await github.paginate(
+              github.rest.pulls.listFiles,
+              { owner: context.repo.owner, repo: context.repo.repo, pull_number: pr.number, per_page: 100 }
+            );
+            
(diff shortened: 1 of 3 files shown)


.github/workflows/verify-changeset.yml
```diff
@@ -0,0 +1,82 @@
...
+      pull-requests: write
+      contents: read
+    steps:
+      - uses: actions/github-script@v7
+        id: filter
+        env:
+          IGNORE_DIRS: ".github/workflows/,.cargo/,.direnv/,.vscode/"
+        with:
+          script: |
+            const pr = context.payload.pull_request;
```
Hunter-apollo: Can we not use `paths` and `paths-ignore` in the `on` block for this filtering?

alocay: Oh let me try that! 👀

alocay: Oh nice call out, this is much simpler!

.github/workflows/verify-changeset.yml
```diff
@@ -0,0 +1,62 @@
...
+    tags-ignore:
+      - "**"
+    branches:
+      - main
+  pull_request:
+    paths-ignore:
+      - '.github/**'
+      - '.cargo/**'
+      - '.direnv/**'
+      - '.vscode/**'
```
swcollard, line 15: This is a great list.

What about things like documentation changes, or dependency updates. Are those things we want to require changesets for too?

alocay: Hmmm good questions... I would say probably no changes that affect only docs or Cargo.toml/lock files. Let me update the list.

swcollard (approved):

.github/workflows/verify-changeset.yml
alocay: Not sure if we should adds docs to the list. I'm going to keep it out of the list for now.

PR #261, opened 2025-08-15
Add retry mechanism for GraphQL API requests to prevent servers

### LINKED ISSUE #208: MCP Server Is Deployment is restarted if the communication https://graphql.api.apollographql.com/api/graphql endpoint is failed. needs retry mechanisam
MCP Server Is restarted if the communication to the https://graphql.api.apollographql.com/api/graphql endpoint is failed. Ideally please implement a retry mechanism to address this errors.

```
Error: Failed to create operation: Error loading collection: error sending request for url (https://graphql.api.apollographql.com/api/graphql)
Caused by:
    Error loading collection: error sending request for url (https://graphql.api.apollographql.com/api/graphql)
```

DaleSeo:
Fixes #208

Right now, the MCP server restarts whenever there's a connectivity issue with the GraphQL API endpoint. This causes the entire server to restart instead of handling the error gracefully.

```
Error: Failed to create operation: Error loading collection: error sending request for url (https://graphql.api.apollographql.com/api/graphql)
Caused by:
    Error loading collection: error sending request for url (https://graphql.api.apollographql.com/api/graphql)
```

This PR adds a retry mechanism with exponential backoff in the collection poller. It differentiates between transient errors, which should be retried, and permanent errors, which should fail immediately.

### DIFF
--- a/crates/apollo-mcp-registry/src/platform_api/operation_collections/collection_poller.rs
+++ b/crates/apollo-mcp-registry/src/platform_api/operation_collections/collection_poller.rs
@@ -4,6 +4,7 @@ use reqwest::header::{HeaderMap, HeaderName, HeaderValue};
 use secrecy::ExposeSecret;
 use std::collections::HashMap;
 use std::pin::Pin;
+use std::time::Duration;
 use tokio::sync::mpsc::channel;
 use tokio_stream::wrappers::ReceiverStream;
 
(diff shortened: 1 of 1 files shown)


crates/apollo-mcp-registry/src/platform_api/operation_collections/collection_poller.rs
```diff
@@ -1,3 +1,4 @@
+use backon::{ExponentialBuilder, Retryable};
```
DaleSeo: I initially wrote some custom retry logic, but later found a great crate for this called [backon](https://crates.io/crates/backon). Switching to `backon` really cleaned up the code and made it easier to maintain, so I replaced my original implementation with it.

crates/apollo-mcp-registry/src/platform_api/operation_collections/collection_poller.rs
```diff
@@ -595,32 +611,41 @@ where
...
-    let response_body: graphql_client::Response<Query::ResponseData> =
-        res.json().await.map_err(CollectionError::Request)?;
-    response_body
-        .data
-        .ok_or(CollectionError::Response("missing data".to_string()))
+        res.json::<graphql_client::Response<Query::ResponseData>>()
+            .await
+    })
+    .retry(ExponentialBuilder::default())
+    .when(should_retry_error)
```
Jephuff: Should we retry ever? We are currently polling every 30s, so if there is a failure we at most getting the operations seconds. I'm thinking it might be best to handle the error and log it, and let the next 30 poll pick up the changes?

DaleSeo: That's a good question, @Jephuff. While retrying does help shorten recovery time, the main benefit of this change is that it prevents the entire server from restarting due to transient connectivity issues. This is why issue #208 specifically requests a retry mechanism. Would you suggest we should only focus on preventing the error from bubbling up, rather than implementing retries?

Jephuff: yep, thats what I would suggest. We are already polling quite frequently so that should cover the retry part of this, so the core issue is that it crashes the server 👍

DaleSeo: Thanks, @Jephuff! I've addressed your feedback.

PR #263, opened 2025-08-17
fix: build error, let expressions unstable in while

ThoreKoritzius:
Running cargo build throws the following error

error[E0658]: `let` expressions in this position are unstable
   --> crates/apollo-schema-index/src/lib.rs:363:19
    |
363 |             while let Some(current_path) = queue.pop_front()
    |                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: see issue #53667 <https://github.com/rust-lang/rust/issues/53667> for more information

For more information about this error, try `rustc --explain E0658`.


DaleSeo:
Thank you for the fix, @ThoreKoritzius. Could you please add a changeset by running `cargo xtask changeset create`? This helps us keep track of changes in the release notes.

PR #265, opened 2025-08-18
feat: initial refactor for moving server serving from state machine

alocay:
Initial attempt at moving out serving the server from the state machine. Still a work in progress and needs testing.

### DIFF
--- a/crates/apollo-mcp-server/src/runtime/serve.rs
+++ b/crates/apollo-mcp-server/src/runtime/serve.rs
@@ -16,6 +16,8 @@ use rmcp::{RoleServer, ServiceExt};
 use serde_json::json;
 use std::io::Error;
 use std::net::{IpAddr, SocketAddr};
+use std::sync::Arc;
+use tokio::sync::RwLock;
 use tokio_util::sync::CancellationToken;
 use tracing::{Instrument, error, info, trace};
 
@@ -153,7 +155,7 @@ async fn serve_sse(
                 error!(error = %e, "mcp shutdown with error");
             }
         }
-            .instrument(tracing::info_span!("mcp-server", bind_address = %server.config.bind)),
+        .instrument(tracing::info_span!("mcp-server", bind_address = %server.config.bind)),
     );
 
     server.with_service(move || server_handler.clone());
--- a/crates/apollo-mcp-server/src/server.rs
+++ b/crates/apollo-mcp-server/src/server.rs
@@ -1,10 +1,10 @@
-use std::net::{IpAddr, Ipv4Addr};
-use std::sync::Arc;
 use apollo_mcp_registry::uplink::schema::SchemaSource;
 use bon::bon;
 use reqwest::header::{CONTENT_TYPE, HeaderMap, HeaderValue};
 use schemars::JsonSchema;
 use serde::Deserialize;
+use std::net::{IpAddr, Ipv4Addr};
+use std::sync::Arc;
 use tokio::sync::RwLock;
 use tokio_util::sync::CancellationToken;
 use url::Url;
@@ -26,7 +26,7 @@ use states::StateMachine;
 pub struct Server<T: McpServerHandler> {
     schema_source: SchemaSource,
     operation_source: OperationSource,
-    server_handler: Arc<RwLock<T>>,
+    server_handler: T,
     cancellation_token: CancellationToken,
     server_config: ServerConfig,
 }
@@ -83,12 +83,12 @@ impl Transport {
 }
 
 #[bon]
-impl<T: McpServerHandler> Server<T> {
+impl<T: McpServerHandler + Clone> Server<T> {
     #[builder]
     pub fn new(
         schema_source: SchemaSource,
         operation_source: OperationSource,
-        server_handler: Arc<RwLock<T>>,
+        server_handler: T,
         cancellation_token: CancellationToken,
         server_config: ServerConfig,
     ) -> Self {
@@ -97,7 +97,7 @@ impl<T: McpServerHandler> Server<T> {
             operation_source,
             server_handler,
             cancellation_token,
-            server_config
+            server_config,
         }
     }
 
--- a/crates/apollo-mcp-server/src/server/states/running.rs
+++ b/crates/apollo-mcp-server/src/server/states/running.rs
@@ -1,7 +1,7 @@
 use std::sync::Arc;
 
 use apollo_compiler::{Schema, validation::Valid};
-use tokio::sync::{Mutex, RwLock};
+use tokio::sync::{Mutex};
 use tracing::{debug, error};
 
 use crate::server_handler::{ApolloMcpServerHandler, McpServerHandler};
@@ -13,10 +13,11 @@ use crate::{
 
 #[derive(Clone)]
 pub struct Running<T = ApolloMcpServerHandler>
-where T: McpServerHandler
+where
+    T: McpServerHandler,
 {
     pub(super) schema: Arc<Mutex<Valid<Schema>>>,
-    pub(super) server_handler: Arc<RwLock<T>>,
+    pub(super) server_handler: T,
     pub(super) custom_scalar_map: Option<CustomScalarMap>,
     pub(super) mutation_mode: MutationMode,
     pub(super) disable_type_description: bool,
@@ -24,20 +25,21 @@ where T: McpServerHandler
 }
 
 impl<T> Running<T>
-where T: McpServerHandler
+where
+    T: McpServerHandler,
 {
     /// Update a running server with a new schema.
-    pub(super) async fn update_schema(self, schema: Valid<Schema>) -> Result<Running<T>, ServerError> {
+    pub(super) async fn update_schema(
+        mut self,
+        schema: Valid<Schema>,
+    ) -> Result<Running<T>, ServerError> {
         debug!("Schema updated:\n{}", schema);
 
         // Update the operations based on the new schema. This is necessary because the MCP tool
         // input schemas and description are derived from the schema.
         let operations: Vec<Operation> = self
             .server_handler
-            .read()
-            .await
             .operations()
-            .lock()
             .await
             .iter()
             .cloned()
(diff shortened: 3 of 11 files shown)


crates/apollo-mcp-server/src/runtime/serve.rs
```diff
@@ -0,0 +1,183 @@
...
+use rmcp::transport::{SseServer, StreamableHttpService, stdio};
+use rmcp::{RoleServer, ServiceExt};
+use serde_json::json;
+use std::io::Error;
+use std::net::{IpAddr, SocketAddr};
+use tokio_util::sync::CancellationToken;
+use tracing::{Instrument, error, info, trace};
+
+// Helper to enable auth
+macro_rules! with_auth {
```
Hunter-apollo, line 25: Unless we are going to disable/modify this macro based on feature flags I would just move this to a function.

crates/apollo-mcp-server/src/runtime/serve.rs
```diff
@@ -0,0 +1,183 @@
...
+            }
+        }
+
+        Ok(())
+    }
+}
+
+// Create health check if enabled (only for StreamableHttp transport)
+fn create_health_check(config: HealthCheckConfig) -> Option<HealthCheck> {
+    // let telemetry: Arc<dyn Telemetry> = Arc::new(InMemoryTelemetry::new());
```
Hunter-apollo, line 75: ?

crates/apollo-mcp-server/src/runtime/serve.rs
```diff
@@ -0,0 +1,183 @@
...
+        let mut router = $router;
+        if let Some(auth) = $auth {
+            router = auth.enable_middleware(router);
+        }
+
+        router
+    }};
+}
+
+pub struct Serve;
```
Hunter-apollo, line 36: Nit: The `Serve` name here is a little confusing as verbs are usually `Trait`s in rust.

crates/apollo-mcp-server/src/runtime/serve.rs
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    let tcp_listener = tokio::net::TcpListener::bind(listen_address).await?;
+    tokio::spawn(async move {
+        // Health check is already active from creation
+        if let Err(e) = axum::serve(tcp_listener, router)
+            .with_graceful_shutdown(shutdown_signal())
+            .await
+        {
+            // This can never really happen
```
Hunter-apollo, line 113: Why cant this happen? If it is really impossible we should mark it as `unreachable!`

crates/apollo-mcp-server/src/runtime/serve.rs
```diff
@@ -0,0 +1,183 @@
...
+    port: u16,
+    server_handler: ApolloMcpServerHandler,
+    cancellation_token: CancellationToken,
+) -> Result<(), Error> {
+    info!(port = ?port, address = ?address, "Starting MCP server in SSE mode");
+    let listen_address = SocketAddr::new(address, port);
+
+    let (server, router) = SseServer::new(SseServerConfig {
+        bind: listen_address,
+        sse_path: "/sse".to_string(),
```
Hunter-apollo, line 133: Nit: Implement `Default` for `SseServerConfig` then this can be reduced to:
```
SseServerConfig {
   bind: listen_address, 
   ct: canellation_token,
   ..Default::default()
}
```

crates/apollo-mcp-server/src/runtime/serve.rs
```diff
@@ -0,0 +1,183 @@
...
+                .with_state(health_check.clone());
+            router = router.merge(health_router);
+        }
+    }
+
+    let tcp_listener = tokio::net::TcpListener::bind(listen_address).await?;
+    tokio::spawn(async move {
+        // Health check is already active from creation
+        if let Err(e) = axum::serve(tcp_listener, router)
+            .with_graceful_shutdown(shutdown_signal())
```
Hunter-apollo, line 110: Should this also use the cancellation token?

crates/apollo-mcp-server/src/runtime/serve.rs
```diff
@@ -0,0 +1,183 @@
...
+
+    server.with_service(move || server_handler.clone());
+    Ok(())
+}
+
+async fn serve_stdio(
+    server_handler: ApolloMcpServerHandler,
+) -> Result<RunningService<RoleServer, ApolloMcpServerHandler>, ServerInitializeError<Error>> {
+    info!("Starting MCP server in stdio mode");
+    server_handler.serve(stdio()).await.inspect_err(|e| {
```
Hunter-apollo, line 169: Nit: I would use `inspect_err` for a number of the `if-let`s above.

crates/apollo-mcp-server/src/runtime/serve.rs
```diff
@@ -0,0 +1,183 @@
...
+        error!("serving error: {:?}", e);
+    })
+}
+
+/// Health check endpoint handler
+async fn health_endpoint(
+    axum::extract::State(health_check): axum::extract::State<HealthCheck>,
+    Query(params): Query<std::collections::HashMap<String, String>>,
+) -> Result<(StatusCode, Json<serde_json::Value>), StatusCode> {
+    let query = params.keys().next().map(|k| k.as_str());
```
Hunter-apollo, line 179: Q: Is it valid to have no params on a call to the health check endpoint?

crates/apollo-mcp-server/src/server.rs
```diff
@@ -13,32 +15,20 @@ use crate::errors::ServerError;
...
-    search_minify: bool,
-    search_introspection: bool,
-    explorer_graph_ref: Option<String>,
-    custom_scalar_map: Option<CustomScalarMap>,
-    mutation_mode: MutationMode,
-    disable_type_description: bool,
-    disable_schema_description: bool,
-    search_leaf_depth: usize,
-    index_memory_bytes: usize,
-    health_check: HealthCheckConfig,
```
Hunter-apollo, line 41: Yay! So many fewer bools

crates/apollo-mcp-server/src/server/states/running.rs
```diff
@@ -1,69 +1,42 @@
...
     pub(super) schema: Arc<Mutex<Valid<Schema>>>,
-    pub(super) operations: Arc<Mutex<Vec<Operation>>>,
-    pub(super) headers: HeaderMap,
-    pub(super) endpoint: Url,
-    pub(super) execute_tool: Option<Execute>,
-    pub(super) introspect_tool: Option<Introspect>,
-    pub(super) search_tool: Option<Search>,
-    pub(super) explorer_tool: Option<Explorer>,
-    pub(super) validate_tool: Option<Validate>,
+    pub(super) server_handler: Arc<RwLock<T>>,
```
Hunter-apollo: If we are going to use `Arc<RwLock<` all over the place could we create a type alias like `RwRef` or `MutexRef`. This allows use to potentially change pointer types and is just generally nicer to look at IMO. Just a personal preference thing though.

crates/apollo-mcp-server/src/server/states/running.rs
```diff
@@ -1,69 +1,42 @@
...
 
         // Update the operations based on the new schema. This is necessary because the MCP tool
         // input schemas and description are derived from the schema.
         let operations: Vec<Operation> = self
-            .operations
+            .server_handler
+            .read()
+            .await
+            .operations()
             .lock()
```
Hunter-apollo: Maybe Im just missing something, but if we already have a lock on the server handler, how could another thread modify the operations? Basically, do we need the inner RwLocks?

crates/apollo-mcp-server/src/server/states/running.rs
```diff
@@ -132,200 +109,24 @@ impl Running {
                 updated_operations.len(),
                 serde_json::to_string_pretty(&updated_operations)?
             );
-            *self.operations.lock().await = updated_operations;
+            *self.server_handler.write().await.operations().lock().await = updated_operations;
```
Hunter-apollo: See comment above on the overall write lock

crates/apollo-mcp-server/src/server/states/running.rs
```diff
@@ -132,200 +109,24 @@ impl Running {
...
         }
 
         // Notify MCP clients that tools have changed
-        Self::notify_tool_list_changed(self.peers.clone()).await;
+        self.server_handler
```
Hunter-apollo, line 116: Thought: I wonder if we could move all the mutable state into an inner RwLock object. Then you wouldnt might have some set of state that is readable cross-thread without the lock.

crates/apollo-mcp-server/src/server_config.rs
```diff
@@ -0,0 +1,71 @@
+use crate::custom_scalar_map::CustomScalarMap;
+use crate::operations::MutationMode;
+use bon::bon;
+use http::header::CONTENT_TYPE;
+use http::{HeaderMap, HeaderValue};
+use crate::health::HealthCheckConfig;
+
+/// Common configuration options for the server
+pub struct ServerConfig {
```
Hunter-apollo, line 9: Maybe I missed how this is created, but it seems odd to me that this doesnt impl `Deserialize`?

crates/apollo-mcp-server/src/server_handler.rs
```diff
@@ -0,0 +1,343 @@
...
+
+    fn operations(&self) -> Arc<Mutex<Vec<Operation>>> {
+        Arc::clone(&self.operations)
+    }
+
+    fn headers(&self) -> HeaderMap {
+        self.headers.clone()
+    }
+
+    fn endpoint(&self) -> Url {
```
Hunter-apollo: If these are public, should we even expose the fields in the struct via `pub(crate)`?

crates/apollo-mcp-server/src/server_handler.rs
```diff
@@ -0,0 +1,343 @@
...
+
+    fn explorer_tool(&self) -> Option<Explorer> {
+        self.explorer_tool.clone()
+    }
+
+    fn validate_tool(&self) -> Option<Validate> {
+        self.validate_tool.clone()
+    }
+
+    fn peers(&self) -> Arc<RwLock<Vec<Peer<RoleServer>>>> {
```
Hunter-apollo: This is quite a type signature lol

crates/apollo-mcp-server/src/runtime/serve.rs
alocay: This is left over from a previous change I was doing, ignore it for now.

crates/apollo-mcp-server/src/runtime/serve.rs
alocay: This is just copied over from what was there. I'd have to take a look to see what this comment was referring to.

PR #266, opened 2025-08-18
docs: replace SSE with Streamable HTTP in the architecture diagram

DaleSeo:
Updating the architecture diagram in the welcome docs to replace the old SSE transport with Streamable HTTP.

2025-08-18 at 11 59 54

### DIFF
--- a/docs/source/index.mdx
+++ b/docs/source/index.mdx
@@ -40,7 +40,7 @@ graph LR
 
     %% Connections
     AI <-->|"Natural Language\nRequests"| MCPClient
-    MCPClient <-->|"MCP Protocol\n(stdio/SSE)"| MCPServer
+    MCPClient <-->|"MCP Protocol\n(stdio/Streamable HTTP)"| MCPServer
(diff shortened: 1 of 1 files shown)


mabuyo:
cc @ronan-apollo I recall you wanted to improve this diagram! Is that work in flight?

ronan-apollo:
I did make a diagram for this purpose - it's more compact / stylized rather than a mermaid chart - WDYT?

redraw-mcp

DaleSeo:
> I did make a diagram for this purpose - it's more compact / stylized rather than a mermaid chart - WDYT?

@ronan-apollo Wow, I love it! Is that the only change, or is it part of a bigger update?

@mabuyo Would you like me to replace the mermaid chart with the new diagram in this PR?

mabuyo (approved):

mabuyo:
@ronan-apollo @DaleSeo I can't find any examples in the docs that uses the stylized format, usually that's only in presentations. I believe we use Mermaid for accessibility / consistency reasons. I'll confirm with Daniel separately. Let's merge this one for now so that conversation is not blocking!

PR #269, opened 2025-08-20
docs: Add quickstart variant for Docker

jcaudle:
This change provides an alternative method for running the Apollo MCP Server. Rather than using Rover CLI, these instructions allow running the MCP Server Docker container directly.

### DIFF
--- a/docs/source/quickstart.mdx
+++ b/docs/source/quickstart.mdx
@@ -105,7 +105,7 @@ You can run the MCP Server using the Rover CLI or Docker.
         enabled: true
     ```
 
-Then, from the root directory of your local repo, run this Docker command to start a local graph with an MCP Server:
+1. In a new terminal, from the root directory of your local repo, run this Docker command to start the MCP Server:
     ```sh showLineNumbers=false
     docker run \
       -it --rm \
@@ -121,22 +121,24 @@ Then, from the root directory of your local repo, run this Docker command to sta
     - Maps configuration files into the proper place for the Apollo MCP Server container
     - Forwards port 5000 for accessing the MCP Server
 
-2. Start MCP Inspector to verify the server is running:
+## Step 3: Verify the MCP Server with MCP Inspector
+
+1. Start MCP Inspector to verify the server is running:
 
     ```sh
     npx @modelcontextprotocol/inspector
     ```
 
-3. Open a browser and go to [`http://127.0.0.1:6274`](http://127.0.0.1:6274)
+1. Open a browser and go to [`http://127.0.0.1:6274`](http://127.0.0.1:6274)
 
-4. In Inspector:
+1. In Inspector:
     - Select `Streamable HTTP` as the Transport Type
     - Enter `http://127.0.0.1:5000/mcp` as the URL
     - Click **Connect**, then **List Tools**
 
     You should see the tools from your server listed.
 
-## Step 3: Connect Claude Desktop
+## Step 4: Connect Claude Desktop
 
 We're using [Claude](https://claude.ai/download) as our AI Assistant (acting as our MCP Client).
 
@@ -171,7 +173,7 @@ You need Node v18 or later installed for `mcp-remote` to work. If you have an ol
 
 2. Restart Claude.
 
-## Step 4: Test Your Setup
+## Step 5: Test Your Setup with Claude
 
 Let's verify everything is working:
 


alocay:
Thanks for the PR @jcaudle! Could you rebase this onto the `develop` branch instead of `main` and make `develop` the destination branch of the PR (can check out the [contributor guidelines](https://github.com/apollographql/apollo-mcp-server/blob/main/CONTRIBUTING.md))? It would be greatly appreciated!

docs/source/quickstart.mdx
```diff
@@ -61,6 +61,10 @@ The example files located in `graphql/TheSpaceDevs/` include:
 
 ## Step 2: Run the MCP Server
 
+There are two methods of running the MCP Server, Rover or Docker. Depending on which you prefer, follow those steps.
```
mabuyo:
```suggestion
You can run the MCP Server using the Rover CLI or Docker. 

> **Note**: We recommend using Docker if you typically use it in your developer workflow. Otherwise, use Rover.
```

What do you think of providing them with more guidance on which to use?

docs/source/quickstart.mdx
```diff
@@ -61,6 +61,10 @@ The example files located in `graphql/TheSpaceDevs/` include:
...
+
+### Rover deployment
```
mabuyo:
```suggestion
### Run with Rover
```

I don't think this is really _deploying_, we're talking about just running the server.

docs/source/quickstart.mdx
```diff
@@ -74,6 +78,47 @@ The example files located in `graphql/TheSpaceDevs/` include:
 
     See the [command reference](/apollo-mcp-server/command-reference#config-options) for a list of available configuration options. 
 
+### Docker deployment
```
mabuyo:
```suggestion
### Run with Docker
```

docs/source/quickstart.mdx
```diff
@@ -74,6 +78,47 @@ The example files located in `graphql/TheSpaceDevs/` include:
...
+
+1. Modify the configuration YAML for the example code to the following:
```
mabuyo:
```suggestion
1. Modify the MCP server configuration in `graphql/TheSpaceDevs/config.yaml` to the following:
```

Specifying what the configuration is for

mabuyo: Can we also add what exactly has changed here from the original file?

docs/source/quickstart.mdx
```diff
@@ -74,6 +78,47 @@ The example files located in `graphql/TheSpaceDevs/` include:
...
+    ```yaml graphql/TheSpaceDevs/config.yaml {7, 10}
```
mabuyo:
```suggestion
    ```yaml title="graphql/TheSpaceDevs/config.yaml" {7, 10}
```
I think this is the proper syntax

docs/source/quickstart.mdx
```diff
@@ -74,6 +78,47 @@ The example files located in `graphql/TheSpaceDevs/` include:
...
+    introspection:
+      execute:
+        enabled: true
+      introspect:
+        enabled: true
+      search:
+        enabled: true
+    ```
+
+Then, from the root directory of your local repo, run this Docker command to start a local graph with an MCP Server:
```
mabuyo:
```suggestion
1. In a new terminal, from the root directory of your local repo, run this Docker command to start a local graph with the MCP Server:
```
Continue with the steps, adding clarification

docs/source/quickstart.mdx
```diff
@@ -74,6 +78,47 @@ The example files located in `graphql/TheSpaceDevs/` include:
...
+    ```sh showLineNumbers=false
+    docker run \
+      -it --rm \
+      --name apollo-mcp-server \
+      -p 5000:5000 \
+      -v $PWD/graphql/TheSpaceDevs/config.yaml:/config.yaml \
+      -v $PWD/graphql/TheSpaceDevs:/data \
```
mabuyo, line 115: Is `$PWD` an environment variable we set up somewhere? What are the `:` for?

docs/source/quickstart.mdx
```diff
@@ -74,6 +78,47 @@ The example files located in `graphql/TheSpaceDevs/` include:
...
+      ghcr.io/apollographql/apollo-mcp-server:latest /config.yaml
+    ```
+
+    This command:
+    - Starts an MCP Server in a Docker container
+    - Maps configuration files into the proper place for the Apollo MCP Server container
+    - Forwards port 5000 for accessing the MCP Server
```
mabuyo, line 122: Does the graph also run with this command?

docs/source/quickstart.mdx
jcaudle: I like it. Thanks for the drafted language.

docs/source/quickstart.mdx
jcaudle: Good call. We refer to deployment elsewhere, but since it's just local, running makes a lot more sense.

docs/source/quickstart.mdx
jcaudle: No. The configuration file points the MCP server to a graph deployed in Railway. This is only the MCP server docker container.

docs/source/quickstart.mdx
jcaudle: `$PWD` is a `bash` native environment variable that refers to the present working directory's full path. the `:` creates a map between the file on the host machine to the location specified after the `:` in the Docker container. I'm not sure if this is helpful to point out for someone who is used to working with Docker as it's a pretty common practice to do this with the Docker CLI, but I also could see the value of including what exactly is going on here.

docs/source/quickstart.mdx
jcaudle: This will probably cause trouble further down the article and I didn't know the best way to deal with it, since there's only one step for running with Rover, but two steps for Docker. What would you recommend for differentiating? Maybe we add another heading for checking the MCP server using MCP Inspector?

docs/source/quickstart.mdx
```diff
@@ -74,6 +80,47 @@ The example files located in `graphql/TheSpaceDevs/` include:
 
     See the [command reference](/apollo-mcp-server/command-reference#config-options) for a list of available configuration options. 
 
+### Run with Docker
+
+1. Modify the MCP server configuration in `graphql/TheSpaceDevs/config.yaml` to the following so it points to the `/data` path you'll map things to as expected by the Dockerfile:
```
mabuyo:
```suggestion
1. Modify the MCP server configuration in `graphql/TheSpaceDevs/config.yaml`. Update the paths to point to the `/data` directory, as the Docker container expects:
```

docs/source/quickstart.mdx
mabuyo: Then we should update the previous step since it currently says "run this Docker command to start a local graph with an MCP Server:"

docs/source/quickstart.mdx
mabuyo: As we discussed last week, let's add a new step to handle the MCP Inspector step!

mabuyo (changes requested):

docs/source/quickstart.mdx
jcaudle: Great catch. I completely missed it since that was what I was trying to do when I first wrote the step for the command, but then realized the command I was trying wasn't as demonstrative of what the Docker container for the MCP server could do and then didn't reread the English after my docker command worked. 😖

Anyway, I think it's all good now.

mabuyo (approved):

mabuyo:
@alocay this docs-only change is ready to go but I see you have a PR up for `develop` and `main`. Are we OK to merge this directly to `main` now? We're also getting a failure on "verify changeset" but I don't think we should be adding a changeset file for docs changes right?

alocay:
> @alocay this docs-only change is ready to go but I see you have a PR up for `develop` and `main`. Are we OK to merge this directly to `main` now? We're also getting a failure on "verify changeset" but I don't think we should be adding a changeset file for docs changes right?

If this is a docs change that needs to go out immediately to production, then yeah it can point to main. If it can wait until next week then it can go to develop. As to the verify changeset issue, it should not be running on PRs that only change docs... not sure why thats not working but we can add the `skip-changeset` label to have it skipped.

Update: it should be fixed now

alocay (approved):

PR #270, opened 2025-08-20
Ignore certain branches for verify changeset

alocay:
Ignoring the changeset file verification workflow if the merge destination is either main or release.


.github/workflows/verify-changeset.yml
```diff
@@ -1,6 +1,9 @@
 name: Verify Changeset
 on:
   pull_request:
+    branches-ignore:
+      - main
+      - release/**
```
DaleSeo, line 6: Just to confirm, do the branches listed here refer to the base branches of the pull requests, or the source branches being merged?

DaleSeo (approved):
Thanks!

.github/workflows/verify-changeset.yml
alocay: The branches listed here are the destination branches. So any PR that is merging INTO main or release/* will be ignored.

PR #272, opened 2025-08-20
fix: generate openAI-compatible json schemas for list types

DaleSeo:
The MCP server is generating JSON schemas that don't match OpenAI's function calling specification. It puts `oneOf` at the array level instead of using `items` to define the JSON schemas for GraphQL list types. While some other LLMs are more flexible about this, it technically violates the [JSON Schema specification](https://json-schema.org/understanding-json-schema/reference/array) that OpenAI strictly follows.

This PR updates the list type handling logic to move `oneOf` inside `items` for GraphQL list types.

#### Before

Valid input fails validation

2025-08-22 at 09 48 55

https://jsonschema.dev/s/kqwqR

#### After

Valid input passes validation

2025-08-22 at 09 50 05

https://jsonschema.dev/s/3jAXW

### DIFF
--- a/crates/apollo-mcp-server/src/custom_scalar_map.rs
+++ b/crates/apollo-mcp-server/src/custom_scalar_map.rs
@@ -94,33 +94,33 @@ mod tests {
     fn empty_file() {
         let result = CustomScalarMap::from_str("").err().unwrap();
 
-        insta::assert_debug_snapshot!(result, @r#"
-        CustomScalarConfig(
-            Error("EOF while parsing a value", line: 1, column: 0),
-        )
-        "#)
+        insta::assert_debug_snapshot!(result, @r###"
+            CustomScalarConfig(
+                Error("EOF while parsing a value", line: 1, column: 0),
+            )
+        "###)
     }
 
     #[test]
     fn only_spaces() {
         let result = CustomScalarMap::from_str("    ").err().unwrap();
 
-        insta::assert_debug_snapshot!(result, @r#"
-        CustomScalarConfig(
-            Error("EOF while parsing a value", line: 1, column: 4),
-        )
-        "#)
+        insta::assert_debug_snapshot!(result, @r###"
+            CustomScalarConfig(
+                Error("EOF while parsing a value", line: 1, column: 4),
+            )
+        "###)
     }
 
     #[test]
     fn invalid_json() {
         let result = CustomScalarMap::from_str("Hello: }").err().unwrap();
 
-        insta::assert_debug_snapshot!(result, @r#"
-        CustomScalarConfig(
-            Error("expected value", line: 1, column: 1),
-        )
-        "#)
+        insta::assert_debug_snapshot!(result, @r###"
+            CustomScalarConfig(
+                Error("expected value", line: 1, column: 1),
+            )
+        "###)
     }
 
     #[test]
@@ -135,13 +135,13 @@ mod tests {
         .err()
         .unwrap();
 
-        insta::assert_debug_snapshot!(result, @r#"
-        CustomScalarJsonSchema(
-            Object {
-                "test": Bool(true),
-            },
-        )
-        "#)
+        insta::assert_debug_snapshot!(result, @r###"
+            CustomScalarJsonSchema(
+                Object {
+                    "test": Bool(true),
+                },
+            )
+        "###)
     }
 
     #[test]
@@ -161,7 +161,7 @@ mod tests {
         .err()
         .unwrap();
 
-        insta::assert_debug_snapshot!(result, @r#"
+        insta::assert_debug_snapshot!(result, @r###"
         CustomScalarJsonSchema(
             Object {
                 "type": String("object"),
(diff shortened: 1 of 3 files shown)


swcollard:
Is there a utility we can include in our unit tests that does the same validation as https://jsonschema.dev/ ?
That way we have more than just the snapshot to verify its correctness

DaleSeo:
Thanks for your input, @swcollard. With a customer waiting to evaluate the MCP server, merging this quick fix should be the immediate priority. I've created [a backlog ticket](https://apollographql.atlassian.net/browse/GT-363) for adding such unit tests, so we can address this improvement in a follow-up.

swcollard (approved):

crates/apollo-mcp-server/src/operations.rs
```diff
@@ -1044,25 +1044,36 @@ fn type_to_schema(
...
+                                instance_type: Some(SingleOrVec::Single(Box::new(
+                                    InstanceType::Null,
+                                ))),
+                                ..Default::default()
+                            }),
+                        ]),
+                        ..Default::default()
+                    })),
+                    ..Default::default()
+                })
```
Jephuff: This was the only place the subschema arg for schemaFactory was used. So I'd say lets either use the schema factory here as well, or delete the subschema argument from schema factory
```suggestion
                schema_factory(
                    None,
                    None,
                    None,
                    None,
                    Some(SubschemaValidation {
                        one_of: Some(vec![
                            inner_type_schema,
                            Schema::Object(SchemaObject {
                                instance_type: Some(SingleOrVec::Single(Box::new(
                                    InstanceType::Null,
                                ))),
                                ..Default::default()
                            }),
                        ]),
                        ..Default::default()
                    }),
                    None,
                )
```

DaleSeo: @Jephuff Thanks for catching that! I've addressed it!

crates/apollo-mcp-server/src/custom_scalar_map.rs
```diff
@@ -94,33 +94,33 @@ mod tests {
     fn empty_file() {
         let result = CustomScalarMap::from_str("").err().unwrap();
 
-        insta::assert_debug_snapshot!(result, @r###"
```
Jephuff: This looks unrelated to the PR change? I'd say if we want to change the style here we should make a separate PR? though maybe Im misremembering what these # do :D

crates/apollo-mcp-server/src/operations.rs
```diff
@@ -1570,14 +1585,16 @@ mod tests {
...
-                            Object {
-                                "type": String("string"),
-                            },
-                            Object {
-                                "type": String("null"),
-                            },
-                        ],
+                        "items": Object {
+                            "oneOf": Array [
+                                Object {
```
Jephuff, line 1590: Looks like all our array test currently only have scalars, could we have one that s a list of input objects? (one for nullable and one for non nullable)

crates/apollo-mcp-server/src/custom_scalar_map.rs
DaleSeo: I didn't really mean to change this. 😅 I just accepted the snapshot changes that Insta generated. This might be cause by the recent dependency updates.

Jephuff: If these are just always updating, imo best to do a dedicated PR for updating the format, especially in unrelated files. For the ones in separate files it muddies the git history where this PR will show up in that files history, for the ones in the same file it makes reviewing more difficult.  

I think you've got approval for the change so maybe not worth removing the ones in operations.rs, but I'd definitely encourage reverting the ones in this file to not muddy the history :)

DaleSeo: I tracked down the relevant Insta changeset, but you're right. I agree it can be confusing to include unrelated changes in this PR. I'll revert the changes in this file.

Image

https://insta.rs/changelog/#1-40-0

crates/apollo-mcp-server/src/operations.rs
DaleSeo: I've added tests for input object lists.

Jephuff (approved):

crates/apollo-mcp-server/src/operations.rs
```diff
@@ -1603,18 +1620,20 @@ mod tests {
...
+                  {
+                    "type": "null"
+                  }
+                ]
+              }
             }
           }
         }
-        "###);
+        "#);
```
swcollard, line 1636: Not blocking, but what caused the change in all these snapshots to use the single `#`?

DaleSeo: This test code was written before Insta v1.40.0. It used to use `###` for inline snapshots but now it uses `#` by default. See https://github.com/apollographql/apollo-mcp-server/pull/272#discussion_r2294333085

PR #274, opened 2025-08-20
Prototype OpenTelemetry Traces in MCP Server

swcollard:
This change pulls in new crates and SDKs for instrumenting the Apollo MCP Server with Open Telemetry Traces. There is still plenty to work on so this PR is set up to merge into a long standing feature branch `feature/telemetry`.

## Changes

* Adds new rust crates to support OTel
* Annotates excecute and call_tool functions with trace macro
* Adds Axum and Tower middleware's for OTel tracing
* Refactors Logging so that all the tracing_subscribers are set together in a single module.

## Testing

Currently no unit testing since most tracing is happening through macros and there is not yet any configuration beyond the environment variables used within the SDKs.

I ran the all-in-one Jaegar container and started the mcp server with the following env vars

```sh
export OTEL_EXPORTER_OTLP_TRACES_ENDPOINT="http://127.0.0.1:4318/v1/traces"
export OTEL_EXPORTER_OTLP_TRACES_PROTOCOL="http/protobuf"
export OTEL_TRACES_SAMPLER="always_on"
```

This was able to produce traces such as...
Screenshot 2025-08-20 at 1 16 06 PM

### DIFF
--- a/crates/apollo-mcp-server/Cargo.toml
+++ b/crates/apollo-mcp-server/Cargo.toml
@@ -55,7 +55,6 @@ opentelemetry-semantic-conventions = "0.30.0"
 tracing-core.workspace = true
 tracing-opentelemetry = "0.31.0"
 opentelemetry-stdout = "0.30.0"
-init-tracing-opentelemetry = "0.30.0"
 
 [dev-dependencies]
 chrono = { version = "0.4.41", default-features = false, features = ["now"] }
--- a/crates/apollo-mcp-server/Cargo.toml
+++ b/crates/apollo-mcp-server/Cargo.toml
@@ -14,6 +14,7 @@ apollo-mcp-registry = { path = "../apollo-mcp-registry" }
 apollo-schema-index = { path = "../apollo-schema-index" }
 axum = "0.8.4"
 axum-extra = { version = "0.10.1", features = ["typed-header"] }
+axum-tracing-opentelemetry = "0.29.0"
 bon = "3.6.3"
 clap = { version = "4.5.36", features = ["derive", "env"] }
 figment = { version = "0.10.19", features = ["env", "yaml"] }
@@ -24,7 +25,16 @@ humantime-serde = "1.1.1"
 jsonwebtoken = "9"
 jwks = "0.4.0"
 lz-str = "0.2.1"
+opentelemetry = "0.30.0"
+opentelemetry-appender-log = "0.30.0"
+opentelemetry-otlp = "0.30.0"
+opentelemetry-resource-detectors = "0.9.0"
+opentelemetry-semantic-conventions = "0.30.0"
+opentelemetry-stdout = "0.30.0"
+opentelemetry_sdk = "0.30.0"
 regex = "1.11.1"
+reqwest-middleware = "0.4.2"
+reqwest-tracing = "0.5.8"
 reqwest.workspace = true
 rmcp = { version = "0.2", features = [
   "server",
@@ -37,24 +47,14 @@ serde.workspace = true
 serde_json.workspace = true
 thiserror.workspace = true
 tokio.workspace = true
-tracing.workspace = true
-tracing-appender = "0.2.3"
-tracing-subscriber = { version = "0.3.19", features = ["env-filter"] }
 tokio-util = "0.7.15"
 tower-http = { version = "0.6.6", features = ["cors", "trace"] }
-url.workspace = true
-opentelemetry = "0.30.0"
-opentelemetry_sdk = "0.30.0"
-opentelemetry-otlp = "0.30.0"
-opentelemetry-resource-detectors = "0.9.0"
-opentelemetry-appender-log = "0.30.0"
-reqwest-tracing = "0.5.8"
-reqwest-middleware = "0.4.2"
-axum-tracing-opentelemetry = "0.29.0"
-opentelemetry-semantic-conventions = "0.30.0"
+tracing-appender = "0.2.3"
 tracing-core.workspace = true
 tracing-opentelemetry = "0.31.0"
-opentelemetry-stdout = "0.30.0"
+tracing-subscriber = { version = "0.3.19", features = ["env-filter"] }
+tracing.workspace = true
+url.workspace = true
 
 [dev-dependencies]
 chrono = { version = "0.4.41", default-features = false, features = ["now"] }
--- a/crates/apollo-mcp-server/Cargo.toml
+++ b/crates/apollo-mcp-server/Cargo.toml
@@ -41,8 +41,21 @@ tracing.workspace = true
 tracing-appender = "0.2.3"
 tracing-subscriber = { version = "0.3.19", features = ["env-filter"] }
 tokio-util = "0.7.15"
-tower-http = { version = "0.6.6", features = ["cors"] }
+tower-http = { version = "0.6.6", features = ["cors", "trace"] }
 url.workspace = true
+opentelemetry = "0.30.0"
+opentelemetry_sdk = "0.30.0"
+opentelemetry-otlp = "0.30.0"
+opentelemetry-resource-detectors = "0.9.0"
+opentelemetry-appender-log = "0.30.0"
+reqwest-tracing = "0.5.8"
+reqwest-middleware = "0.4.2"
+axum-tracing-opentelemetry = "0.29.0"
+opentelemetry-semantic-conventions = "0.30.0"
+tracing-core.workspace = true
+tracing-opentelemetry = "0.31.0"
+opentelemetry-stdout = "0.30.0"
+init-tracing-opentelemetry = "0.30.0"
 
 [dev-dependencies]
 chrono = { version = "0.4.41", default-features = false, features = ["now"] }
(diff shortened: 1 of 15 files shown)


crates/apollo-mcp-server/Cargo.toml
```diff
@@ -41,8 +41,20 @@ tracing.workspace = true
...
+opentelemetry-otlp = "0.30.0"
+opentelemetry-resource-detectors = "0.9.0"
+opentelemetry-appender-log = "0.30.0"
+reqwest-tracing = "0.5.8"
+reqwest-middleware = "0.4.2"
+axum-tracing-opentelemetry = "0.29.0"
+opentelemetry-semantic-conventions = "0.30.0"
+tracing-core.workspace = true
+tracing-opentelemetry = "0.31.0"
+opentelemetry-stdout = "0.30.0"
```
DaleSeo: Why don't we keep the dependencies sorted alphabetically?

crates/apollo-mcp-server/src/runtime/trace.rs
```diff
@@ -0,0 +1,111 @@
...
+    trace::{RandomIdGenerator, SdkTracerProvider},
+};
+use opentelemetry_semantic_conventions::{
+    SCHEMA_URL,
+    attribute::{DEPLOYMENT_ENVIRONMENT_NAME, SERVICE_VERSION},
+};
+use tracing_opentelemetry::{MetricsLayer, OpenTelemetryLayer};
+use tracing_subscriber::{layer::SubscriberExt, util::SubscriberInitExt};
+
+use crate::runtime::{Config, logging::Logging};
```
DaleSeo: nit: Would be helpful to use ‎`super` instead to communicate that they come from the parent module?

```suggestion
use super::{Config, logging::Logging};
```

crates/apollo-mcp-server/src/runtime/trace.rs
```diff
@@ -0,0 +1,111 @@
...
+        .with_resource(resource())
+        .with_batch_exporter(exporter)
+        .build();
+
+    global::set_tracer_provider(trace_provider.clone());
+
+    Ok(trace_provider)
+}
+
+// Initialize tracing-subscriber and return OtelGuard for opentelemetry-related termination processing
```
DaleSeo: nit: I'd like to see this comment when hovering on this function call in `main.rs`. 😄

```suggestion
/// Initialize tracing-subscriber and return OtelGuard for opentelemetry-related termination processing
```

crates/apollo-mcp-server/src/runtime/trace.rs
```diff
@@ -0,0 +1,111 @@
...
+// Create a Resource that captures information about the entity for which telemetry is recorded.
+fn resource() -> Resource {
+    Resource::builder()
+        .with_service_name(env!("CARGO_PKG_NAME"))
+        .with_schema_url(
+            [
+                KeyValue::new(SERVICE_VERSION, env!("CARGO_PKG_VERSION")),
+                KeyValue::new(
+                    DEPLOYMENT_ENVIRONMENT_NAME,
+                    std::env::var("ENVIRONMENT").unwrap_or_else(|_| "develop".to_string()),
```
DaleSeo: This should more conventional in most ecosystems like Docker, Node.js, Rails, etc. (Not confident about Rust though 😅)

```suggestion
                    std::env::var("ENVIRONMENT").unwrap_or_else(|_| "development".to_string()),
```

crates/apollo-mcp-server/src/runtime/trace.rs
```diff
@@ -0,0 +1,111 @@
...
+pub struct TelemetryGuard {
+    tracer_provider: SdkTracerProvider,
+    meter_provider: SdkMeterProvider,
+    logging_guard: Option<tracing_appender::non_blocking::WorkerGuard>,
+}
+
+impl Drop for TelemetryGuard {
+    fn drop(&mut self) {
+        if let Err(err) = self.tracer_provider.shutdown() {
+            eprintln!("{err:?}");
```
DaleSeo: Is it okay to use `eprintln!` here? I've noticed ‎`tracing::error!` being used elsewhere in the codebase, but I’m not sure which is more appropriate in this context.

crates/apollo-mcp-server/src/runtime/trace.rs
DaleSeo, line 1: What is the rationale for placing this module under `runtime`? Sorry, I'm still a bit unclear about the intended purpose of the `runtime` folder. 😭
The existing modules mostly seem to define configuration structs without much logic, so I assumed this was the right place for code like this, but I might be wrong.

crates/apollo-mcp-server/src/introspection/tools/introspect.rs
```diff
@@ -27,7 +27,7 @@ pub struct Introspect {
 }
 
 /// Input for the introspect tool.
-#[derive(JsonSchema, Deserialize)]
+#[derive(JsonSchema, Deserialize, Debug)]
```
DaleSeo, line 30: So, I understand instrumenting a function requires that all its inputs implement `Debug`? Right?

crates/apollo-mcp-server/src/introspection/tools/introspect.rs
```diff
@@ -55,6 +55,7 @@ impl Introspect {
         }
     }
 
+    #[tracing::instrument(skip(self))]
```
DaleSeo, line 58: What does `skip(self)` mean?

DaleSeo: Oh, I guess it means ignoring `&self` because it's huge object.

crates/apollo-mcp-server/src/runtime/logging.rs
```diff
@@ -52,8 +50,18 @@ impl Default for Logging {
...
+        tracing_subscriber::fmt::format::DefaultFields,
+        tracing_subscriber::fmt::format::Format,
+        BoxMakeWriter,
+    >,
+    Option<tracing_appender::non_blocking::WorkerGuard>,
+);
+
 impl Logging {
-    pub fn setup(config: &Config) -> Result<Option<WorkerGuard>, anyhow::Error> {
+    pub fn env_filter(config: &Config) -> Result<EnvFilter, anyhow::Error> {
```
DaleSeo: Would it make sense to pass only `config.logging` to this function, since that's the only part of the config it actually uses?

crates/apollo-mcp-server/src/runtime/logging.rs
```diff
@@ -62,7 +70,10 @@ impl Logging {
                 .add_directive("rmcp=warn".parse()?)
                 .add_directive("tantivy=warn".parse()?);
         }
+        Ok(env_filter)
+    }
 
+    pub fn logging_layer(config: &Config) -> Result<LoggingLayerResult, anyhow::Error> {
```
DaleSeo: Same feedback as above just to reduce the API surface.

crates/apollo-mcp-server/src/graphql.rs
```diff
@@ -74,11 +79,21 @@ pub trait Executable {
             }
         }
 
-        reqwest::Client::new()
+        let client = ClientBuilder::new(reqwest::Client::new())
+            .with_init(Extension(OtelName("graphql-client".into())))
```
DaleSeo: Should we include 'mcp' somewhere in the name for clarity, or would it already be distinguishable enough based on other attributes?

crates/apollo-mcp-server/src/graphql.rs
```diff
@@ -74,11 +79,21 @@ pub trait Executable {
...
+
+        client
             .post(request.endpoint.as_str())
+            .with_extension(OtelName(
+                format!("POST {}", request.endpoint.as_str()).into(),
+            ))
             .headers(self.headers(&request.headers))
             .body(Value::Object(request_body).to_string())
             .send()
+            .with_current_context()
```
DaleSeo: I'm curious about the order here. Is this called after `send()` so that both the request and response can be captured?

crates/apollo-mcp-server/Cargo.toml
swcollard: Will fix, thats rough that cargo add doesn't do that for you :(

crates/apollo-mcp-server/src/introspection/tools/introspect.rs
swcollard: Correct, if we want to include the request object, they need to implement the Debug trait. Alternatively, we could `skip` and not log the inputs

crates/apollo-mcp-server/src/runtime/trace.rs
swcollard: My rationale was co-locating the Tracing  setup code with the Logging code since they both need to register components to the `tracing_subscriber`

crates/apollo-mcp-server/src/runtime/trace.rs
swcollard: TIL about `///`

crates/apollo-mcp-server/src/runtime/trace.rs
swcollard: Good catch, changed to `tracing::error!`

crates/apollo-mcp-server/src/graphql.rs
swcollard: I don't think this is working correctly. I put it after the send() because it operates on a Future so it needed to be right before the `await` call. But I am not seeing the trace headers in the graphql request

DaleSeo (approved):
Thanks for addressing my feedback! 🙇

PR #282, opened 2025-08-25
Gt 367 GitHub workflow sync main pr

alocay:
Adding a github workflow to auto-generate a sync PR to merge main into develop.

- Triggers on PR being merged into main.
- Creates a sync branch to merge changes from main into develop (not a rebase).
- Adds a comment to the original merged PR with a link to this sync PR.
- Added a workflow dispatch trigger for manual testing on branches other than main and develop.

### DIFF
--- a/.github/workflows/sync-develop.yml
+++ b/.github/workflows/sync-develop.yml
@@ -4,6 +4,23 @@ on:
   pull_request:
     types: [closed]
     branches: [main]
+  workflow_dispatch:
+    inputs:
+      head_branch:
+        description: "Branch to merge FROM (default: main)."
+        required: false
+        default: "main"
+      base_branch:
+        description: "Branch to merge INTO (default: develop)."
+        required: false
+        default: "develop"
+      source_pr_number:
+        description: "If testing, the PR number to comment on (optional)."
+        required: false
+      test_mode:
+        description: "Bypass PR/push guards for manual testing"
+        required: false
+        default: "true"
 
 permissions:
   contents: write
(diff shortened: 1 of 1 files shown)


.github/workflows/sync-develop.yml
```diff
@@ -0,0 +1,116 @@
+name: Sync main → develop
+
+on:
+  pull_request:
+    types: [closed]
+    branches: [main]
```
swcollard, line 6: A pull request to main happens when we have a release, correct? 

Can we make sure that the settings for this package is set up to where you can only merge to main from pull requests? So folks can't push to main directly

.github/workflows/sync-develop.yml
```diff
@@ -0,0 +1,116 @@
...
+          git push origin HEAD
+          echo "merge_status=$rc" >> "$GITHUB_OUTPUT"
+
+      # Open the PR targeting develop
+      - name: Open PR to develop
+        id: syncpr
+        uses: peter-evans/create-pull-request@v6
+        with:
+          branch: ${{ steps.meta.outputs.branch }}
+          base: ${{ env.BASE_BRANCH }}
```
swcollard, line 94: I'm a little confused why in this case it uses `env.BASE_BRANCH` when above it just references it as `BASE_BRANCH`?

I am guessing it's because above that is in a bash script and not just yml templating so the syntax is different.

.github/workflows/sync-develop.yml
```diff
@@ -0,0 +1,116 @@
...
+          title: ${{ steps.meta.outputs.title }}
+          body:  |
+            ${{ steps.meta.outputs.body }}
+
+            Merge status: ${{ steps.prep.outputs.merge_status == '0' && 'clean ✅' || 'conflicts ❗' }}
+          labels: ${{ steps.prep.outputs.merge_status != '0' && 'back-merge,automation,conflicts' || 'back-merge,automation' }}
```
swcollard: nit, seems weird to use == in the first but then != in the second row, swapping the order of the ternary flow

swcollard (approved):
This all looks reasonable to me! Inputs line up with the create pr action, https://github.com/marketplace/actions/create-pull-request

.github/workflows/sync-develop.yml
alocay: I believe main already has this branch protection rule in place by default but we can double check.

.github/workflows/sync-develop.yml
alocay: So the `run` portion above runs as a bash script so we just reference it as `${BASE_BRANCH}` but this portion doesn't run as a bash script and instead needs it as a GitHub action expression. Hence the two different formats.

.github/workflows/sync-develop.yml
alocay: Good point, I'll keep it consistent.

PR #286, opened 2025-08-25
ci: setting git author in sync workflow to github bot

alocay:


.github/workflows/sync-develop.yml
```diff
@@ -54,6 +54,11 @@ jobs:
         with:
           fetch-depth: 0
 
+      - name: Configure git author
+        run: |
+          git config --local user.name "github-actions[bot]"
+          git config --local user.email "41898282+github-actions[bot]@users.noreply.github.com"
```
swcollard: minor question, where does this email come from and can we assume it will be static?

I would have thought this would be part of a context object we can inject in, but that is unclear from the docs
https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#github-context

alocay: From what I've gathered this is the default user and no-reply email for the github bot. Using this will show that the git commands run later are done by the github bot. This isn't ideal so I'm working with some folks to see if there's an Apollo automation account we can use instead.

swcollard (approved):

PR #290, opened 2025-08-26
ci: using gh cli for PR creation and adding a fallback in case of merge conflicts

alocay:
1. Using GH CLI for PR creation
2. Adding fallback in case of merge conflicts
3. Updating comment to include merge conflict status

### DIFF
--- a/.github/workflows/sync-develop.yml
+++ b/.github/workflows/sync-develop.yml
@@ -48,6 +48,7 @@ jobs:
       HEAD_BRANCH: ${{ (github.event_name == 'workflow_dispatch' && inputs.head_branch) || 'main' }}
       BASE_BRANCH: ${{ (github.event_name == 'workflow_dispatch' && inputs.base_branch) || 'develop' }}
       SOURCE_PR: ${{ (github.event_name == 'pull_request' && github.event.pull_request.number) || inputs.source_pr_number || '' }}
+      GH_TOKEN: ${{ secrets.GH_PAT }}
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/sync-develop.yml
```diff
@@ -91,34 +92,94 @@ jobs:
...
-          branch: ${{ steps.meta.outputs.branch }}
-          base: ${{ env.BASE_BRANCH }}
-          title: ${{ steps.meta.outputs.title }}
-          body:  |
-            ${{ steps.meta.outputs.body }}
+        id: sync_pr
+        if: ${{ steps.prep.outputs.merge_status == '0' }}
+        run: |
+          # Does sync branch actually differ from base?
+          read left right < <(git rev-list --left-right --count "origin/${BASE_BRANCH}...HEAD")
```
Hunter-apollo: IIRC you can simplify this a bit. Something like.
```
if [ x`git rev-list --count origin/main...` != "0" ];
```

.github/workflows/sync-develop.yml
```diff
@@ -91,34 +92,94 @@ jobs:
...
+          if [ "$right" -eq 0 ]; then
+            echo "No diff; skipping gh pr create"
+            exit 0
+          fi
+
+          # Avoid duplicate PRs
+          existing=$(gh pr list --base "${BASE_BRANCH}" --head "${{ steps.meta.outputs.branch }}" --state open --json number --jq '.[0].number' || true)
```
Hunter-apollo, line 108: Im pretty sure the gh cli already errors out on duplicate PRs. As long as the branch names are the same we should just be able to fail there.

.github/workflows/sync-develop.yml
```diff
@@ -91,34 +92,94 @@ jobs:
...
+          gh pr view --base "${BASE_BRANCH}" --head "${{ steps.meta.outputs.branch }}" \
+            --json number,url | jq -r '"pr_number=\(.number)\npr_url=\(.url)"' >> "$GITHUB_OUTPUT"
+
+      # If the merge hit conflicts, open a DIRECT PR: HEAD_BRANCH -> BASE_BRANCH so conflicts can be resolved prior to merge
+      - name: Open direct PR via gh (conflicts fallback)
+        id: pr_fallback
+        if: ${{ steps.prep.outputs.merge_status != '0' }}
+        run: |
+          # Skip if no diff between head and base
+          read left right < <(git rev-list --left-right --count "origin/${BASE_BRANCH}...origin/${HEAD_BRANCH}")
```
Hunter-apollo: see above.

.github/workflows/sync-develop.yml
```diff
@@ -91,34 +92,94 @@ jobs:
...
+            echo "No diff between ${HEAD_BRANCH} and ${BASE_BRANCH}; nothing to open."
+            exit 0
+          fi
+          
+          # Reuse existing open PR if present
+          existing=$(gh pr list --base "${BASE_BRANCH}" --head "${HEAD_BRANCH}" --state open --json number --jq '.[0].number' || true)
+          if [ -n "$existing" ] && [ "$existing" != "null" ]; then
+            echo "pr_number=$existing" >> "$GITHUB_OUTPUT"
+            url=$(gh pr view "$existing" --json url --jq .url)
+            echo "pr_url=$url" >> "$GITHUB_OUTPUT"
```
Hunter-apollo, line 144: Have we tested updating an existing branch?

.github/workflows/sync-develop.yml
Hunter-apollo: Reading further I think what I suggested is not the right approach. Closing comment

.github/workflows/sync-develop.yml
alocay: My concern here is this will add up the value of right and left but I think we only care about commits that are on the head (main) and not the base (develop) which would be right. Left would anything develop has that main doesn't which isn't relevant for this (and likely to often be different).

.github/workflows/sync-develop.yml
alocay: Not yet, I'll test this via the dispatch trigger.

.github/workflows/sync-develop.yml
alocay: Updated

.github/workflows/sync-develop.yml
alocay: Updated

Hunter-apollo (approved):

PR #305, opened 2025-08-28
test: hardcoded version strings in tests

DaleSeo:
The GraphQL tests have hardcoded version strings that we need to update manually each time we release a new version. Since this isn't included in the release checklist, it's easy to miss it and only notice the test failures later.

### DIFF
--- a/crates/apollo-mcp-server/src/graphql.rs
+++ b/crates/apollo-mcp-server/src/graphql.rs
@@ -187,7 +187,7 @@ mod test {
             "extensions": {
                 "clientLibrary": {
                     "name":"mcp",
-                    "version":"0.7.4"
+                    "version": std::env!("CARGO_PKG_VERSION")
                 }
             },
             "operationName":"mock_operation"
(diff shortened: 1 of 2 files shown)


crates/apollo-mcp-server/src/graphql.rs
```diff
@@ -233,7 +233,7 @@ mod test {
                 },
                 "clientLibrary": {
                     "name":"mcp",
-                    "version":"0.7.4"
+                    "version": std::env!("CARGO_PKG_VERSION")
```
swcollard, line 236: Thanks for this change. The release checklist makes no mention of this needing to be updated each time. https://github.com/apollographql/apollo-mcp-server/blob/main/RELEASE_CHECKLIST.md

DaleSeo: @swcollard That's exactly what I was saying in the PR description! 
I just wanted to eliminate the need for the manual updates here, rather than add another step to the release checklist.

alocay: 👏🏼

PR #308, opened 2025-08-28
[Docs] Re-organizing docs pages

mabuyo:
## Proposal to re-organize docs:
- Split up **User Guide** page into multiple pages, one for each section. This change surfaces the developer workflow higher up to the sidebar navigation.
- Replaces **Command Reference** with **YAML Config Reference**. This change provides a clearer title to what the page is referring to. 
- Add new page for **Run the MCP Server**
- Remove **Install** in favour of co-locating install commands with the **Run** commands depending on what they're using

With this PR, I aimed to keep the content as-is and focus on re-organizing and placement only. In a follow-up PR I'd like to do a full content review for each page and tighten things up! This is just the bare minimum PR.

## **Before**

Screenshot 2025-08-28 at 3 27 02 PM

## **After**

Screenshot 2025-09-02 at 11 42 58 AM

## TODO
- [x] Set up redirects to ensure no broken links


docs/source/run.mdx
```diff
@@ -0,0 +1,69 @@
...
+
+```sh
+rover dev --mcp <PATH/TO/CONFIG> 
+```
+
+## Run the MCP server with Docker
+
+1. Modify the MCP server configuration in `graphql/TheSpaceDevs/config.yaml`. Update the paths to point to the `/data` directory, as the Docker container expects:
+    ```yaml title="graphql/TheSpaceDevs/config.yaml" {7, 10}
+    endpoint: https://thespacedevs-production.up.railway.app/
```
DaleSeo: I recommend updating this setting as well to point to the user's GraphQL API URL or Apollo Router URL. Since instructions involve modifying the schema file, this endpoint won't work, so adjusting the URL will help prevent issues.

docs/source/command-reference.mdx
jcaudle, line 1: Is this being retained just for the redirect?

docs/source/guides/auth.mdx
```diff
@@ -16,262 +16,4 @@ To implement authorization, you need an [OAuth 2.1-compliant](https://oauth.net/
 
 Then, you [configure the MCP server with `auth` settings](/apollo-mcp-server/command-reference#auth) and the [GraphOS Router for JWT authentication](/graphos/routing/security/jwt) using those IdP values.
```
jcaudle:
```suggestion
Then, you [configure the MCP server with `auth` settings](/apollo-mcp-server/config-file#auth) and the [GraphOS Router for JWT authentication](/graphos/routing/security/jwt) using those IdP values.
```

docs/source/_sidebar.yaml
```diff
@@ -4,22 +4,39 @@ switcher:
...
+  - label: "Debugging"
+    href: "./debugging"
+  - label: "Deployment"
+    href: "./deploy"
   - label: "Authorization"
     href: "./guides/auth"
-  - label: "Command Reference"
-    href: "./command-reference"
+  - label: "Health Checks"
+    href: "./health-checks"
```
jcaudle: Part of me thinks it would make sense to put Health Checks under the Configuration parent element, but I could go either way. Since the container is opened by default, it doesn't make it any less discoverable, unless the sidebar behavior changes.

docs/source/_sidebar.yaml
```diff
@@ -4,22 +4,39 @@ switcher:
...
+      - label: "YAML Config Reference"
+        href: "./config-file"
+      - label: "Custom Scalars"
+        href: "./custom-scalars"
+  - label: "Debugging"
+    href: "./debugging"
+  - label: "Deployment"
+    href: "./deploy"
   - label: "Authorization"
     href: "./guides/auth"
```
jcaudle: Should this file be moved to a base path since it doesn't show up in the Guides parent?

jcaudle (commented):
This is great stuff. I love how you're making this a lot more approachable.

docs/source/command-reference.mdx
mabuyo: This file will be removed and redirected!

docs/source/_sidebar.yaml
mabuyo: I think Health checks might be more related to Deployment since it's only really necessary when the server is in production. I'll try an iteration with this and see how it looks.

docs/source/_sidebar.yaml
mabuyo: Yes good call!

mabuyo:
cc @DaleSeo (and anyone from @apollographql/graph-tooling team!) -- this PR is ready for official review! I pushed a second iteration based on feedback from y'all and Amanda.

As mentioned in the PR description, I'd like to do a full content review for each page and tighten things up in a separate PR. This just sets the stage for the content organization so that we can focus in on each page for content review afterwards.

docs/source/_sidebar.yaml
```diff
@@ -4,22 +4,39 @@ switcher:
     - label: v0
       latest: true
       href: ./
+defaultOpenDepth: 2
 items:
-  - label: "Welcome"
+  - label: "Overview"
     href: "."
   - label: "Quickstart"
```
esilverm, line 11: Should we rename this label "Get Started"? The title for the page is that and it seems to be the pattern for other quickstart guides in our docs

docs/source/config-file.mdx
```diff
@@ -180,25 +157,25 @@ The available fields depend on the value of the nested `type` key:
 
 ##### stdio (default)
 
-| Option    | Value               | Default Value | Description                                                                                                              |
-| :-------- | :------------------ | :------------ | :----------------------------------------------------------------------------------------------------------------------- |
-| `type`    | `"stdio"`           | \*             | Use standard IO for communication between the server and client                                                          |
+| Option | Value     | Default Value | Description                                                     |
```
DaleSeo, line 160: The tables in this section use slightly different formats, which could be confusing for users.

Image

DaleSeo (approved):
This looks great, thanks for the re-organization! I have just one minor comment.

esilverm (approved):
Structure-wise this is looking a lot better! I think this may need another pass when we add more content but as of right now it looks a lot better and I can find what I need more easily!

docs/source/config-file.mdx
mabuyo: I'll log this in a separate ticket, good catch!

docs/source/_sidebar.yaml
mabuyo: Yes you're right! I'll log this with an existing ticket where we tackle the contents of the page 👍🏽

mabuyo:
TY for the reviews! 🙌🏽 I'll merge this tomorrow since it's near the end of my day now and I wanna make sure I'm available for a big docs change like this 👀

docs/source/auth.mdx
```diff
@@ -0,0 +1,21 @@
+---
+title: Authorization with Apollo MCP Server
+redirectFrom:
+    - /apollo-mcp-server/guides/auth
+---
+
+The Apollo MCP server supports authorizing clients (e.g., LLMs) in accordance with [the MCP specification](https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization).
+
+The current implementation passes through OAuth tokens from MCP clients directly to upstream GraphQL APIs. You can read more about [security considerations](/apollo-mcp-server/limitations#oauth-token-passthrough) when using this feature.
```
swcollard, line 9:
```suggestion
The current implementation passes through validated OAuth tokens from MCP clients directly to upstream GraphQL APIs. You can read more about [security considerations](/apollo-mcp-server/limitations#oauth-token-passthrough) when using this feature.
```
We've had some confusion around the token passthrough, so I think it would be good to call out that the tokens we do pass to the graphql API have been validated by the MCP server using their configured oauth provider.

README.md
jcaudle, line 1: Nice catches on the readme!

jcaudle (approved):
This looks like it'll be much easier to navigate. I especially love that you caught the URLs in the `README`. Nice work!

docs/source/auth.mdx
mabuyo: Thank you! I've added this note to a separate ticket where we tackle updating the Auth pages

PR #309, opened 2025-08-29
ci: adding a release prep github workflow

alocay:
Creating github workflow triggered by a manual dispatch to prep the release branch. Need this workflow in main for further testing. The workflow looks like this following:

1. Provide version bump type
2. Enforce custom version provided if custom is given
3. Pulls current version from Cargo.toml
4. Bumps the version
5. Creates the release branch off develop in the format `release/x.y.x` where `x.y.z` is the new version
6. Updates the Cargo.toml using `cargo set-version`
7. Updates the various scripts using python regex
8. Push the changes to the release branch

### DIFF
--- a/.github/workflows/prep-release.yml
+++ b/.github/workflows/prep-release.yml
@@ -1,0 +1,164 @@
+name: Prep release
+
+on:
+  workflow_dispatch:
+    inputs:
+      version_bump:
+        type: choice
+        description: "Type of version bump"
+        default: patch
+        required: true
+        options:
+          - major
+          - minor
+          - patch
+          - custom
+      custom_version:
+        type: string
+        required: false
+        description: "Custom version (ignore for other bump types)"
+      generate_pre_release:
+        type: boolean
+        default: true
+        required: true
+        description: "Generate a RC build"
+
+permissions:
+  contents: write
+  pull-requests: write
+
+concurrency:
+  group: pre-release
+  cancel-in-progress: false
+
+jobs:
+  validate:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Enforce custom_version when bump=custom
+        run: |
+          if [[ "${{ inputs.version_bump }}" == "custom" ]]; then
+            if [[ -z "${{ inputs.custom_version }}" ]]; then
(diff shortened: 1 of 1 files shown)


alocay:
**Note:** The custom bump type I expect to rarely be used but added it as a just in case

esilverm (approved):

.github/workflows/prep-release.yml
```diff
@@ -0,0 +1,164 @@
...
+        env:
+          CURR: ${{ steps.meta.outputs.current_version }}
+          CUSTOM: ${{ inputs.custom_version }}
+          BUMP: ${{ inputs.bump }}
+        run: |
+          set -euo pipefail
+          
+          if [[ -n "${CUSTOM:-}" ]]; then then
+            # strip any pre-release / build metadata for arithmetic (e.g., -rc.1, +build.5)
+            BASE="${CURR%%[-+]*}"
```
swcollard, line 84: This is some advanced bash. I guess its using some kind of regex to strip things after a `-` or `+`?

.github/workflows/prep-release.yml
```diff
@@ -0,0 +1,164 @@
...
+            
+            IFS=. read -r MA MI PA <<< "$BASE"
+            
+            case "$BUMP" in
+              major) MA=$((MA+1)); MI=0; PA=0 ;;
+              minor) MI=$((MI+1)); PA=0 ;;
+              patch) PA=$((PA+1)) ;;
+              *) echo "::error::Unknown bump '$BUMP'"; exit 1 ;;
```
swcollard, line 92: Are there missing open parenthesis here?

.github/workflows/prep-release.yml
alocay: Sorry, i merged this before answering. Yeah this just looks for either `+` or `-`, strips everything after it.

.github/workflows/prep-release.yml
alocay: I don't think so, case statements in bash start each pattern with `pattern1)` and the default being `*)`

PR #315, opened 2025-09-02
ci: updating release prep and adding canary release workflow

alocay:
Updating release prep workflow and adding a canary workflow to generate a container on `develop` push.


nicholascioli (approved):
Looks good to me!

PR #319, opened 2025-09-02
feat: Open Telemetry Integration

swcollard:
This change implements Open Telemetry instrumentation to the Apollo MCP Server. With this feature, metrics and traces can be exported to an OpenTelemetry Protocol (OTLP) compatible collector for use in any observability vendor that supports OTel such as grafana, datadog, and jaegar.

All documentation to configure the feature is included in this PR, as well as the list of metrics being reported.

For reviewing, the main sections are
* Configuration for setting up metric and trace exporting
* Instrumentation of important sections of the code. MCP functions like call tool and initialization, and the processes of generating MCP Tools from GraphQL Operations.
* A new prebuild step that generates an Enum of metrics and attributes from a telemetry.toml file
* Propagation of OTel trace headers to downstream API for distributed tracing of MCP Tool calls.
* Configuration to omit potentially high-cardinality attributes from the metrics and traces being exported.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,6 +54,9 @@ url = { version = "2.4", features = ["serde"] }
 [workspace.metadata]
 crane.name = "apollo-mcp"
 
+# This allows usage of coverage(off) attribute without causing a linting error.
+# This attribute doesn't work in stable Rust yet and can be removed whenever it does.
+# See https://github.com/apollographql/apollo-mcp-server/pull/372
 [workspace.lints.rust]
 unexpected_cfgs = { level = "warn", check-cfg = ['cfg(coverage_nightly)'] }
 
--- a/crates/apollo-mcp-server/build.rs
+++ b/crates/apollo-mcp-server/build.rs
@@ -100,10 +100,6 @@ fn main() {
     // Generate the keys
     let telemetry_attribute_data = flatten(telemetry.attributes);
     let telemetry_metrics_data = flatten(telemetry.metrics);
-    println!(
-        "a {:?} | m {:?}",
-        telemetry_attribute_data, telemetry_metrics_data
-    );
 
     // Write out the generated keys
     let out_dir = std::env::var_os("OUT_DIR").expect("could not retrieve output directory");
--- a/crates/apollo-mcp-server/src/graphql.rs
+++ b/crates/apollo-mcp-server/src/graphql.rs
@@ -1,7 +1,8 @@
 //! Execute GraphQL operations from an MCP tool
 
+use crate::errors::McpError;
 use crate::generated::telemetry::{TelemetryAttribute, TelemetryMetric};
-use crate::{errors::McpError, meter::get_meter};
+use crate::meter;
 use opentelemetry::KeyValue;
 use reqwest::header::{HeaderMap, HeaderValue};
 use reqwest_middleware::{ClientBuilder, Extension};
@@ -40,7 +41,7 @@ pub trait Executable {
     /// Execute as a GraphQL operation using the endpoint and headers
     #[tracing::instrument(skip(self, request))]
     async fn execute(&self, request: Request<'_>) -> Result<CallToolResult, McpError> {
-        let meter = get_meter();
+        let meter = &meter::METER;
         let start = std::time::Instant::now();
         let mut op_id: Option<String> = None;
         let client_metadata = serde_json::json!({
@@ -135,14 +136,13 @@ pub trait Executable {
             ),
             KeyValue::new(
                 TelemetryAttribute::OperationId.to_key(),
-                op_id.unwrap_or("unknown".to_string()),
+                op_id.unwrap_or("".to_string()),
             ),
             KeyValue::new(
                 TelemetryAttribute::OperationSource.to_key(),
-                if self.persisted_query_id().is_some() {
-                    "persisted_query"
-                } else {
-                    "operation"
+                match self.persisted_query_id() {
+                    Some(_) => "persisted_query",
+                    None => "operation",
                 },
             ),
         ];
@@ -162,6 +162,7 @@ pub trait Executable {
 #[cfg(test)]
 mod test {
     use crate::errors::McpError;
+    use crate::generated::telemetry::TelemetryMetric;
     use crate::graphql::{Executable, OperationDetails, Request};
     use http::{HeaderMap, HeaderValue};
     use opentelemetry::global;
@@ -453,7 +454,7 @@ mod test {
                 .find(|scope_metrics| scope_metrics.scope().name() == "apollo.mcp")
             {
                 for metric in scope_metrics.metrics() {
-                    if metric.name() == "apollo.mcp.operation.count"
+                    if metric.name() == TelemetryMetric::OperationCount.as_str()
                         && let AggregatedMetrics::U64(MetricData::Sum(data)) = metric.data()
                     {
                         for point in data.data_points() {
--- a/crates/apollo-mcp-server/src/lib.rs
+++ b/crates/apollo-mcp-server/src/lib.rs
@@ -17,6 +17,7 @@ pub(crate) mod schema_tree_shake;
 pub mod server;
 pub mod telemetry_attributes;
 
+/// These values are generated at build time by build.rs using telemetry.toml as input.
 pub mod generated {
     pub mod telemetry {
         include!(concat!(env!("OUT_DIR"), "/telemetry_attributes.rs"));
--- a/crates/apollo-mcp-server/src/meter.rs
+++ b/crates/apollo-mcp-server/src/meter.rs
@@ -1,8 +1,4 @@
 use opentelemetry::{global, metrics::Meter};
-use std::sync::OnceLock;
+use std::sync::LazyLock;
 
-static METER: OnceLock<Meter> = OnceLock::new();
-
-pub fn get_meter() -> &'static Meter {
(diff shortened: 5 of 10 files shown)


docs/source/telemetry.mdx
```diff
@@ -0,0 +1,168 @@
...
+        protocol: "http/protobuf"       # or "grpc"
+    tracing:
+      otlp:
+        endpoint: "https://your-traces-endpoint" 
+        protocol: "http/protobuf"
+```
+
+#### Observability platform integration
+
+The MCP server works with any OTLP-compatible backend. Consult your provider's documentation for specific endpoint URLs and authentication:
```
ronan-apollo, line 67: Love the _joining-the-dots_ here!

Two links are throwing 404s though - DataDog & New Relic.

Cargo.toml
```diff
@@ -54,6 +54,9 @@ url = { version = "2.4", features = ["serde"] }
 [workspace.metadata]
 crane.name = "apollo-mcp"
 
+[workspace.lints.rust]
+unexpected_cfgs = { level = "warn", check-cfg = ['cfg(coverage_nightly)'] }
```
nicholascioli, line 61: Could we add a note as to why this is needed? I saw from [here](https://blog.rust-lang.org/2024/05/06/check-cfg/#expecting-custom-cfgs) that this is configurable but I'm not sure why there is something with nightly coverage in our stack if we are pinned to a stable version of rust.

crates/apollo-mcp-server/build.rs
```diff
@@ -0,0 +1,173 @@
...
+        toml::from_str(&raw).expect("could not parse telemetry file")
+    };
+
+    // Generate the keys
+    let telemetry_attribute_data = flatten(telemetry.attributes);
+    let telemetry_metrics_data = flatten(telemetry.metrics);
+    println!(
+        "a {:?} | m {:?}",
+        telemetry_attribute_data, telemetry_metrics_data
+    );
```
nicholascioli: These seems like debug prints. Did we still want them?

crates/apollo-mcp-server/src/graphql.rs
```diff
@@ -109,7 +125,37 @@ pub trait Executable {
...
+
+        // Record response metrics
+        let attributes = vec![
+            KeyValue::new(
+                TelemetryAttribute::Success.to_key(),
+                result.as_ref().is_ok_and(|r| r.is_error != Some(true)),
+            ),
+            KeyValue::new(
+                TelemetryAttribute::OperationId.to_key(),
+                op_id.unwrap_or("unknown".to_string()),
```
nicholascioli: Nit: Do we want the unknown case to be written in a way that would be impossible for it to be confused with a legitimate operation named "unknown"? Maybe something like `<UNKNOWN>`.

crates/apollo-mcp-server/src/graphql.rs
```diff
@@ -109,7 +125,37 @@ pub trait Executable {
...
+            ),
+            KeyValue::new(
+                TelemetryAttribute::OperationSource.to_key(),
+                if self.persisted_query_id().is_some() {
+                    "persisted_query"
+                } else {
+                    "operation"
+                },
```
nicholascioli: Nit: Might be nice to match here so that if we eventually want to do something with the PQ ID then we have it at least.
```suggestion
                match self.persisted_query_id() {
                    Some(_) => "persisted_query",
                    None => "operation",
                }
```

crates/apollo-mcp-server/src/graphql.rs
```diff
@@ -357,4 +408,76 @@ mod test {
...
+        let finished_metrics = exporter.get_finished_metrics().unwrap();
+
+        // validate the attributes of the apollo.mcp.operation.count counter
+        for resource_metrics in finished_metrics {
+            if let Some(scope_metrics) = resource_metrics
+                .scope_metrics()
+                .find(|scope_metrics| scope_metrics.scope().name() == "apollo.mcp")
+            {
+                for metric in scope_metrics.metrics() {
+                    if metric.name() == "apollo.mcp.operation.count"
```
nicholascioli: Nit: Is this one of the constants generated by the build script?

crates/apollo-mcp-server/src/lib.rs
```diff
@@ -7,7 +9,15 @@ mod graphql;
...
 pub mod sanitize;
 pub(crate) mod schema_tree_shake;
 pub mod server;
+pub mod telemetry_attributes;
+
+pub mod generated {
+    pub mod telemetry {
+        include!(concat!(env!("OUT_DIR"), "/telemetry_attributes.rs"));
+    }
+}
```
nicholascioli, line 25: Nit: Maybe add a module doc here explaining that these values are generated by the build script from the telemetry attributes file?

crates/apollo-mcp-server/src/meter.rs
```diff
@@ -0,0 +1,8 @@
+use opentelemetry::{global, metrics::Meter};
+use std::sync::OnceLock;
+
+static METER: OnceLock<Meter> = OnceLock::new();
+
+pub fn get_meter() -> &'static Meter {
+    METER.get_or_init(|| global::meter(env!("CARGO_PKG_NAME")))
+}
```
nicholascioli: Any reason this isn't using [`LazyLock`](https://doc.rust-lang.org/stable/std/sync/struct.LazyLock.html#method.new) instead? I think if you switch to using `LazyLock` then you won't need to have a function to access the Meter, as it can be directly referenced by the static instead.

crates/apollo-mcp-server/src/runtime/filtering_exporter.rs
```diff
@@ -0,0 +1,238 @@
...
+    fn export(&self, mut batch: Vec<SpanData>) -> impl Future<Output = OTelSdkResult> + Send {
+        for span in &mut batch {
+            span.attributes
+                .retain(|kv| filter_omitted_apollo_attributes(kv, &self.omitted));
+
+            // TODO: while not strictly necessary for dealing with high-cardinality, do we want to
+            //  filter out from span.events.events as well?
+            // for ev in &mut span.events.events {
+            //     ev.attributes.retain(|kv| filter_omitted_apollo_attributes(kv, &self.allow));
+            // }
```
nicholascioli: Nit: Was this meant to be kept?

crates/apollo-mcp-server/src/runtime/filtering_exporter.rs
```diff
@@ -0,0 +1,238 @@
...
+use opentelemetry_sdk::error::OTelSdkResult;
+use opentelemetry_sdk::trace::{SpanData, SpanExporter};
+use std::collections::HashSet;
+use std::fmt::Debug;
+
+#[derive(Debug)]
+pub struct FilteringExporter<E> {
+    inner: E,
+    omitted: HashSet<Key>,
+}
```
nicholascioli, line 12: Nit: If most of the methods are going to be pass-through, you might want to just implement `Deref` for this.

docs/source/telemetry.mdx
swcollard: Good catch, just updated those with new links

crates/apollo-mcp-server/src/server/states/running.rs
```diff
@@ -170,41 +176,51 @@ impl Running {
...
+            .u64_counter(TelemetryMetric::InitializeCount.as_str())
+            .build()
+            .add(1, &[]);
         // TODO: how to remove these?
         let mut peers = self.peers.write().await;
         peers.push(context.peer);
         Ok(self.get_info())
     }
 
+    #[tracing::instrument(skip(self, context, request), fields(apollo.mcp.tool_name = request.name.as_ref(), apollo.mcp.request_id = %context.id.clone()))]
```
nicholascioli, line 196: Nit: Using the raw field name here instead of the constants generated by the build script might cause some confusion later if they ever diverge. While the macro might not let you use the constant, you should be able to run the code generated by this macro manually.

crates/apollo-mcp-server/src/telemetry_attributes.rs
nicholascioli, line 1: Nit: This entire file can probably be generated with the generated code.

Cargo.toml
swcollard: Agreed, will add a comment and a link to the PR https://github.com/apollographql/apollo-mcp-server/pull/372

crates/apollo-mcp-server/src/graphql.rs
swcollard: Or I think an empty string might be more appropriate in this case.

crates/apollo-mcp-server/src/meter.rs
swcollard: Good catch, converted to LazyLock

crates/apollo-mcp-server/src/runtime/filtering_exporter.rs
swcollard: Probably not. Removed

crates/apollo-mcp-server/src/runtime/filtering_exporter.rs
swcollard: For the remaining ones, since they are more related to code cleanup and not functional changes, I am going to create an internal follow up ticket to address them so we can get this merged and the release candidate build created for QA. Appreciate all the feedback!

PR #320, opened 2025-09-02
test: adding a basic manual e2e test for mcp server

alocay:
Adding some basic e2e tests using [mcp-server-tester](https://github.com/steviec/mcp-server-tester). Currently the tool does not always exit (ctrl+c is sometimes needed) so this should be run manually.

### How to run tests?
Added a script `run_tests.sh` (may need to run `chmod +x` to run it) to run tests. Basic usage found via `./run_tests.sh -h`. The script does the following:

1. Builds test/config yaml paths and verifies the files exist.
2. Checks if release `apollo-mcp-server` binary exists. If not, it builds the binary via `cargo build --release`.
3. Reads in the template file (used by `mcp-server-tester`) and replaces all `<test-dir>` placeholders with the test directory value. Generates this test server config file and places it in a temp location.
4. Invokes the `mcp-server-tester` via `npx`.
5. On script exit the generated config is cleaned up.

### Example run: 
To run the tests for `local-operations` simply run `./run_tests.sh local-operations`

### DIFF
--- a/.changesets/feat_alocay_add_basic_e2e_test.md
+++ b/.changesets/feat_alocay_add_basic_e2e_test.md
@@ -1,0 +1,15 @@
+### test: adding a basic manual e2e test for mcp server - @alocay PR #320
+
+Adding some basic e2e tests using [mcp-server-tester](https://github.com/steviec/mcp-server-tester). Currently, the tool does not always exit (ctrl+c is sometimes needed) so this should be run manually.
+
+### How to run tests?
+Added a script `run_tests.sh` (may need to run `chmod +x` to run it) to run tests. Basic usage found via `./run_tests.sh -h`. The script does the following:
+
+1. Builds test/config yaml paths and verifies the files exist.
+2. Checks if release `apollo-mcp-server` binary exists. If not, it builds the binary via `cargo build --release`.
+3. Reads in the template file (used by `mcp-server-tester`) and replaces all `<test-dir>` placeholders with the test directory value. Generates this test server config file and places it in a temp location.
+4. Invokes the `mcp-server-tester` via `npx`.
+5. On script exit the generated config is cleaned up.
+
+### Example run: 
+To run the tests for `local-operations` simply run `./run_tests.sh local-operations`
--- a/e2e/mcp-server-tester/local-operations/config.yaml
+++ b/e2e/mcp-server-tester/local-operations/config.yaml
@@ -1,0 +1,21 @@
+endpoint: https://thespacedevs-production.up.railway.app/
+transport:
+  type: stdio
+operations:
+  source: local
+  paths:
+    - ./local-operations/operations
+schema:
+  source: local
+  path: ./local-operations/api.graphql
+overrides:
+  mutation_mode: all
+introspection:
+  execute:
+    enabled: true
+  introspect:
+    enabled: true
+  search:
+    enabled: true
+  validate:
+    enabled: true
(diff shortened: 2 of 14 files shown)


.changesets/feat_alocay_add_basic_e2e_test.md
```diff
@@ -0,0 +1,15 @@
...
+Added a script `run_tests.sh` (may need to run `chmod +x` to run it) to run tests. Basic usage found via `./run_tests.sh -h`. The script does the following:
+
+1. Builds test/config yaml paths and verifies the files exist.
+2. Checks if release `apollo-mcp-server` binary exists. If not, it builds the binary via `cargo build --release`.
+3. Reads in the template file (used by `mcp-server-tester`) and replaces all `<test-dir>` placeholders with the test directory value. Generates this test server config file and places it in a temp location.
+4. Invokes the `mcp-server-tester` via `npx`.
+5. On script exit the generated config is cleaned up.
+
+### Example run: 
+To run the tests for `local-operations` simply run `./run_tests.sh local-operations`
```
swcollard, line 15: slight improvement suggestion. It might be nice if `./run_tests.sh` by default runs all the tests?

e2e/mcp-server-tester/local-operations/config.yaml
```diff
@@ -0,0 +1,21 @@
+endpoint: https://thespacedevs-production.up.railway.app/
+transport:
+  type: stdio
```
swcollard, line 3: Will this test suite eventually support other transports?

e2e/mcp-server-tester/pq-manifest/tool-tests.yaml
```diff
@@ -0,0 +1,65 @@
+tools:
+  expected_tool_list: ['introspect', 'execute', 'search', 'validate', 'SearchUpcomingLaunches', 'ExploreCelestialBodies', 'GetAstronautDetails', 'GetAstronautsCurrentlyInSpace']
```
swcollard, line 2: Would it be feasible to actually call one of the graphql operation tools in this first round of tests?

.changesets/feat_alocay_add_basic_e2e_test.md
alocay: I could but the problem is the mcp-server-tester sometimes doesn't exit. This would hang the script.

e2e/mcp-server-tester/local-operations/config.yaml
alocay: I'm not sure, currently it does not work with streamable-http.

e2e/mcp-server-tester/pq-manifest/tool-tests.yaml
alocay: Yeah we can add a test for that. It would just be like the tests below, but calling something like `SearchUpcomingLaunches` instead.

alocay: I'll see if I can add one in a follow up PR

e2e/mcp-server-tester/run_tests.sh
```diff
@@ -0,0 +1,74 @@
...
+# Safe replacement for <test-dir> with absolute path (handles /, &, and |)
+safe_dir="${TEST_DIR//\\/\\\\}"
+safe_dir="${safe_dir//&/\\&}"
+safe_dir="${safe_dir//|/\\|}"
+
+# Replace the literal token "<test-dir>" everywhere
+sed "s|<test-dir>|$safe_dir|g" "$TEMPLATE_PATH" > "$GEN_CONFIG"
+
+# Run the command
+exec npx -y mcp-server-tester tools "$TESTS" --server-config "$GEN_CONFIG"
```
swcollard: last question, what does `exec` provide here when you can also run npx directly in the bash script?

alocay: Actually you'r eright, this should not use exec. the `trap` will not work if this is exec.

.changesets/feat_alocay_add_basic_e2e_test.md
swcollard: Linking the issue here for future reference. https://github.com/steviec/mcp-server-tester/issues/30

PR #326, opened 2025-09-03
Gt 337 GitHub workflow release finalize

alocay:
Finalizing the prep release workflow. Adding the following steps:

1. Updating changelog via cargo xtask
2. Extracting the changelog portion for the new version (using python)
3. Pushing changes to branch
4. Opening a draft PR with these changes and the extracted changelog section as the description

### DIFF
--- a/.github/workflows/prep-release.yml
+++ b/.github/workflows/prep-release.yml
@@ -238,14 +238,14 @@ jobs:
         run: |
           set -euo pipefail
           # Try to create; if it already exists, update it
-          gh pr create \
+          if ! gh pr create \
             --base main \
             --head "$HEAD" \
             --title "$TITLE" \
             --draft \
             --body-file CHANGELOG_SECTION.md \
             --label release
-          || {
+          then
             num=$(gh pr list --head "$HEAD" --base main --state open --json number -q '.[0].number' || true)
             if [[ -n "$num" ]]; then
               gh pr edit "$num" --title "$TITLE" --body-file CHANGELOG_SECTION.md --add-label release
@@ -253,4 +253,4 @@ jobs:
               echo "::error::Failed to create or find PR from $HEAD to main"
               exit 1
             fi
-          }
+          fi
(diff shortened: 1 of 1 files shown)


.github/workflows/canary-release.yml
```diff
@@ -5,6 +5,15 @@ on:
...
       - "**"
+    paths-ignore:
+      - '.github/**'
+      - '.cargo/**'
+      - '.direnv/**'
+      - '.vscode/**'
+      - 'docs/**'
+      - 'Cargo.*'
+      - 'crates/**/Cargo.*'
+      - '*.md'
```
nicholascioli, line 16: Nit: Do we want this to just be `paths` instead set to rust source files?

.github/workflows/prep-release.yml
```diff
@@ -161,10 +161,96 @@ jobs:
...
+              if nxt:
+                # adjust to absolute end
+                segment = segment[: (len('# [' + new + ']') + nxt.start())]
+      
+            segment = segment.rstrip() + "\n"
+            print(segment)
+          except Exception:
+            import traceback
+            traceback.print_exc()
+            sys.exit(1)
```
nicholascioli, line 216: Nit: We might want this to just be a file that it includes and runs vs writing inline. Your choice

.github/workflows/prep-release.yml
```diff
@@ -161,10 +161,96 @@ jobs:
...
+            --label release
+          then
+            num=$(gh pr list --head "$HEAD" --base main --state open --json number -q '.[0].number' || true)
+            if [[ -n "$num" ]]; then
+              gh pr edit "$num" --title "$TITLE" --body-file CHANGELOG_SECTION.md --add-label release
+            else
+              echo "::error::Failed to create or find PR from $HEAD to main"
+              exit 1
+            fi
+          fi
```
nicholascioli, line 256: Nit: This works for now, but it might be more stable as a cargo xtask in the future since bash can be very finicky.

PR #328, opened 2025-09-03
chore: update RMCP version to latest

nicholascioli:
This PR updates the RMCP version to latest.

Sadly, this also means that re-exported dependencies have changed, mainly schemars being removed and also stabilized since. Since we use schemars to generate JSON schemas for operations and custom scalar maps, this meant changing a lot of unrelated filed to conform to the new API.

I've also gone ahead and split up the `operations.rs` file to make it more legible and easier to maintain in the future.


crates/apollo-mcp-server/Cargo.toml
```diff
@@ -21,12 +21,13 @@ futures.workspace = true
...
 http = "1.3.1"
 humantime-serde = "1.1.1"
+jsonschema = "0.33.0"
 jsonwebtoken = "9"
 jwks = "0.4.0"
 lz-str = "0.2.1"
 regex = "1.11.1"
 reqwest.workspace = true
-rmcp = { version = "0.2", features = [
+rmcp = { version = "0.6", features = [
```
DaleSeo, line 33: How about we pin it to an exact version like 0.6.3 to ensure reproducible builds, especially since this update required significant refactoring?

crates/apollo-mcp-server/src/operations.rs
```diff
@@ -1,3541 +1,15 @@
...
+//! Operations
+//!
+//! This module includes transformation utilities that convert GraphQL operations
+//! into MCP tools.
+
+mod mutation_mode;
+mod operation;
+mod operation_source;
+mod raw_operation;
+mod schema_walker;
```
DaleSeo, line 10: Oh, I love this refactoring! ❤️
Working with a single module that had over 3000 lines was not fun. Thanks for breaking it up!

crates/apollo-mcp-server/src/errors.rs
```diff
@@ -93,7 +93,7 @@ pub enum ServerError {
     StartupError(#[from] JoinError),
 
     #[error("Failed to initialize MCP server")]
-    McpInitializeError(#[from] rmcp::service::ServerInitializeError<std::io::Error>),
+    McpInitializeError(#[from] rmcp::service::ServerInitializeError),
```
DaleSeo: It's great to see this updated with rmcp. 😁

crates/apollo-mcp-server/src/operations/operation.rs
```diff
@@ -0,0 +1,3021 @@
...
+        schema
+            .ensure_object()
+            .insert("definitions".to_string(), definitions.into());
+    }
+
+    schema
+}
+
+#[cfg(test)]
+mod tests {
```
DaleSeo, line 584: Have you considered also breaking this massive test suite into the submodules? I guess that would be too ambitious? 😂

crates/apollo-mcp-server/src/operations/schema_walker/type.rs
```diff
@@ -0,0 +1,74 @@
+use apollo_compiler::{Schema as GraphQLSchema, ast::Type as GraphQLType};
+use schemars::{Schema as JSONSchema, json_schema};
+use serde_json::{Map, Value};
+
+use crate::custom_scalar_map::CustomScalarMap;
+
+use super::name::Name;
+
+pub(super) struct Type<'a> {
```
DaleSeo, line 9: https://github.com/apollographql/apollo-mcp-server/blob/ea32f7da9871633512af3323068f3286aa216ec6/crates/apollo-mcp-server/src/operations/schema_walker/name.rs#L10-L11

Do you want to add a similar comment here?

```suggestion
/// A GraphQL Type Walker
pub(super) struct Type<'a> {
```

crates/apollo-mcp-server/src/operations/schema_walker.rs
```diff
@@ -0,0 +1,43 @@
...
+use schemars::Schema;
+use serde_json::{Map, Value};
+
+use crate::custom_scalar_map::CustomScalarMap;
+
+pub(super) mod name;
+pub(super) mod r#type;
+
+/// Walk a top-level type and generate a JSON Schema
+pub fn walk(
```
DaleSeo: Does this function name clearly convey its purpose? 🤔 It might be confusing for someone new to the codebase. I feel like the previous name, `type_to_schema` was more intuitive. What do you think?

crates/apollo-mcp-server/src/operations/schema_walker.rs
```diff
@@ -0,0 +1,43 @@
...
+//! the types recursively.
+
+use apollo_compiler::{Schema as GraphQLSchema, ast::Type};
+use schemars::Schema;
+use serde_json::{Map, Value};
+
+use crate::custom_scalar_map::CustomScalarMap;
+
+pub(super) mod name;
+pub(super) mod r#type;
```
DaleSeo: These could just be private since they're not used outside this module.

```suggestion
mod name;
mod r#type;
```

crates/apollo-mcp-server/src/operations/operation.rs
```diff
@@ -0,0 +1,3021 @@
...
+                }
+            "#,
+            "operation.graphql",
+        )
+        .expect("schema should parse")
+        .validate()
+        .expect("schema should be valid")
+    });
+
+    /// JSON-ifies the input, sorting the object keys
```
DaleSeo: You meant `JSON-files` or perhaps `serializes to JSON`?

crates/apollo-mcp-server/src/operations/operation.rs
nicholascioli: Sadly they all use the same `Operation::from_document` test instead of testing the specific flow, which makes breaking it up harder :(

We should definitely revisit this later to break it up.

crates/apollo-mcp-server/Cargo.toml
nicholascioli: The cargo lock has this as `0.6.1`, but I can make it more explicit here if you prefer!

crates/apollo-mcp-server/src/operations.rs
nicholascioli: Yeah, this file used to make my eyes bleed 😢

crates/apollo-mcp-server/src/errors.rs
nicholascioli: Yeah I'm not sure why it templatized the error before

crates/apollo-mcp-server/src/operations/schema_walker/type.rs
nicholascioli: Nice catch!

crates/apollo-mcp-server/src/operations/schema_walker.rs
nicholascioli: Hmm, yeah I can use that name again. I tried to convey that this process is recursive by using `walk`, but maybe that doesn't need to be conveyed in the name. I'll just add to the docs for this method that it might overflow the stack if the type is sufficiently nested.

As an aside, I mostly kept the existing implementation, but we should probably change this to not be recursive since we do not control the input here.

crates/apollo-mcp-server/src/operations/schema_walker.rs
nicholascioli: Nice catch!

crates/apollo-mcp-server/src/operations/operation.rs
nicholascioli: Oh, I meant to become JSON by using the -ify (but plural) suffix. I'll update it to be more explicit.

swcollard (approved):

PR #330, opened 2025-09-03
Add basic config file options to otel telemetry

swcollard:
Adds new Configuration options for setting up configuration beyond the standard OTEL environment variables needed before.

* Renames trace->telemetry
* Adds OTLP options for metrics and tracing to choose grpc or http upload protocols and setting the endpoints
* This configuration is all optional, so by default nothing will be logged


crates/apollo-mcp-server/src/runtime/telemetry.rs
```diff
@@ -0,0 +1,314 @@
...
+                exporters: Some(Exporters { metrics, tracing }),
+                service_name: service_name.map(|s| s.to_string()),
+                version: version.map(|v| v.to_string()),
+            },
+            ..Default::default()
+        }
+    }
+
+    #[tokio::test]
+    async fn test_full_config() {
```
alocay: nit: I tend to push for unit test names that indicate what is expected, using full sentences if needed... e.g. `guard_is_provided_when_tracing_configued()`, `unknown_provider_raises_an_error()`, etc.

I personally like a convention like this one https://enterprisecraftsmanship.com/posts/you-naming-tests-wrong/

swcollard: Good call, updated in latest revision

Issue #327, opened 2025-09-03
Validate variables with query


m4recek:
Can the validate tool validate variables along with the query? I see a lot of variable errors while trying to generate fairly simple queries. It is bad on GPT5 and a lot worse on other models. 

This is with these in the model context:
- UUID instructions
- query format instructions 
- query examples
- schema description
- product description

Examples: 
GPT5: 
<img width="864" height="720" alt="Image" src="https://github.com/user-attachments/assets/db6e3a18-5b1b-42c0-b268-bba70d08776c" />

Sonet 3.5
<img width="869" height="968" alt="Image" src="https://github.com/user-attachments/assets/90e2d070-7c54-48c8-8e26-d509259d500f" />

esilverm:
Thank you for reaching out. It looks like our `validate` tool isn't properly checking the operation against the schema and we have a fix in line to be released.

As for the variable issues, those appear to be results of LLM hallucination as we do provide input validation which can be seen by the `{"error": "MCP error -32602: Invalid input"}` messages. We have some changes to improve the LLMs understanding of the the input arguments so it is provided as a string containing JSON so there is less hallucination.

Both of these issues should be resolved in `v0.7.5`.

esilverm:
The latest release contains the changes that resolve this issue!

PR #331, opened 2025-09-03
release: v0.7.5

alocay:
# [0.7.5] - 2025-09-03

### fix: Validate ExecutableDocument in validate tool - @swcollard PR #329

Contains fixes for https://github.com/apollographql/apollo-mcp-server/issues/327

The validate tool was parsing the operation passed in to it against the schema but it wasn't performing the validate function on the ExecutableDocument returned by the Parser. This led to cases where missing required arguments were not caught by the Tool.

This change also updates the input schema to the execute tool to make it more clear to the LLM that it needs to provide a valid JSON object

### test: adding a basic manual e2e test for mcp server - @alocay PR #320

Adding some basic e2e tests using [mcp-server-tester](https://github.com/steviec/mcp-server-tester). Currently, the tool does not always exit (ctrl+c is sometimes needed) so this should be run manually.

### How to run tests?
Added a script `run_tests.sh` (may need to run `chmod +x` to run it) to run tests. Basic usage found via `./run_tests.sh -h`. The script does the following:

1. Builds test/config yaml paths and verifies the files exist.
2. Checks if release `apollo-mcp-server` binary exists. If not, it builds the binary via `cargo build --release`.
3. Reads in the template file (used by `mcp-server-tester`) and replaces all `<test-dir>` placeholders with the test directory value. Generates this test server config file and places it in a temp location.
4. Invokes the `mcp-server-tester` via `npx`.
5. On script exit the generated config is cleaned up.

### Example run:
To run the tests for `local-operations` simply run `./run_tests.sh local-operations`

### Update snapshot format - @DaleSeo PR #313

Updates all inline snapshots in the codebase to ensure they are consistent with the latest insta format.

### Hardcoded version strings in tests - @DaleSeo PR #305

The GraphQL tests have hardcoded version strings that we need to update manually each time we release a new version. Since this isn't included in the release checklist, it's easy to miss it and only notice the test failures later.


swcollard (approved):
🚀

esilverm (approved):

Issue #288, opened 2025-08-26
Using ${env.APOLLO_COLLECTION_ID} for operations.id throws UUID validation


riginoommen:
## Description  
When configuring **`operations.id`** in `mcp.yaml` with an environment variable, such as:  

```yaml
operations:
  source: collection
  id: ${env.APOLLO_COLLECTION_ID}
```

The server fails to start and throws the following error:  

```
Error loading collection: Error in response: ${env.APOLLO_COLLECTION_ID} is not a valid UUID
```

This happens even though the environment variable is properly set in `.env` file and exported in the shell before running the command:  

```bash
set -a; source .env; set +a; ./apollo-mcp-server mcp.yaml
```

---

### Expected Behavior  
`operations.id` should accept a valid UUID when injected via environment variables (similar to other fields like `apollo_graph_ref`).  

---

### Actual Behavior  
Fails with error:  

```
Error in response: ${env.APOLLO_COLLECTION_ID} is not a valid UUID
```

---

### Steps to Reproduce
1. Add `APOLLO_COLLECTION_ID` to `.env` file:
    ```env
    APOLLO_COLLECTION_ID=<valid-uuid>
    ```
2. Update `mcp.yaml`:
    ```yaml
    operations:
      source: collection
      id: ${env.APOLLO_COLLECTION_ID}
    ```
3. Run:
    ```bash
    set -a; source .env; set +a; ./apollo-mcp-server mcp.yaml
    ```
4. Observe the error.

**Screenshot** 

<img width="951" height="862" alt="Image" src="https://github.com/user-attachments/assets/de7918cb-a597-4158-b7b9-49a7f8fc3b57" />

nicholascioli:
Hello, thanks for reaching out.

The error stems from the fact that the MCP server does not yet support Router-style variable expansion in the config file itself. It does support setting values in the config using environment variables using a well-structured format, however.

In your case, remove the `operations.id` entry in your config and set the environment variable `APOLLO_MCP_OPERATIONS__ID` to the value that you want to use.

As a side note, I've noticed that our documentation does not address this, so I'll make sure that it gets updated to include this warning in the command reference for the Apollo MCP server. Meanwhile, feel free to refer to the [release notes](https://github.com/apollographql/apollo-mcp-server/releases/tag/v0.6.0) for the configuration changes.

riginoommen:
also on the same grounds endpoint variable cannot be customised

```
endpoint: https://graphql.com
    graphos:

[...]

PR #332, opened 2025-09-04
Docs: add section on overriding config options with environment variables

mabuyo:
Adding missing documentation for overriding config options with environment variables, as flagged in https://github.com/apollographql/apollo-mcp-server/issues/288#issuecomment-3224823429

Also adds a few more helpful links to relevant pages for config options.

### DIFF
--- a/docs/source/config-file.mdx
+++ b/docs/source/config-file.mdx
@@ -234,12 +234,12 @@ You can override configuration options using environment variables. The environm
 For example, to override the `introspection.execute.enabled` option, you can set the `APOLLO_MCP_INTROSPECTION__EXECUTE__ENABLED` environment variable.
 
 ```sh
-APOLLO_MCP_INTROSPECTION__EXECUTE__ENABLED=true
+APOLLO_MCP_INTROSPECTION__EXECUTE__ENABLED="true"
 ```
 
 For list values, you can set the environment variable to a comma-separated list.
 
-For example, to override the `transport.auth.servers` option, you can set the ` APOLLO_MCP_TRANSPORT__AUTH__SERVERS` environment variable to a comma-separated list.
+For example, to override the `transport.auth.servers` option, you can set the `APOLLO_MCP_TRANSPORT__AUTH__SERVERS` environment variable to a comma-separated list.
 
 ```sh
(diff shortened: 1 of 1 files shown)


docs/source/config-file.mdx
```diff
@@ -220,3 +204,43 @@ transport:
...
+```
+
+## Override configuration options using environment variables
+
+You can override configuration options using environment variables. The environment variable name is the same as the option name, but with `APOLLO_MCP_` prefixed. You can use `__` to mark nested options.
+
+For example, to override the `introspection.execute.enabled` option, you can set the `APOLLO_MCP_INTROSPECTION__EXECUTE__ENABLED` environment variable.
+
+```sh
+APOLLO_MCP_INTROSPECTION__EXECUTE__ENABLED=true
```
DaleSeo: Did it work? 😮 I thought bools would throw an error.

```suggestion
APOLLO_MCP_INTROSPECTION__EXECUTE__ENABLED="true"
```

docs/source/config-file.mdx
```diff
@@ -220,3 +204,43 @@ transport:
...
+
+The following example file sets your endpoint to `localhost:4001`, configures transport over Streamable HTTP, provides a GraphOS key and graph reference,
+enables introspection, and provides two local MCP operations for the server to expose.
+
+```yaml config.yaml
+endpoint: http://localhost:4001/
+transport:
+  type: streamable_http
+graphos:
+  apollo_key: <YOUR_APOLLO_KEY>
```
DaleSeo: Not sure who would want to set the API key directly in a config file. 🤔 I feel most folks would prefer using an environment variable instead.

docs/source/config-file.mdx
mabuyo: It didn't give me an error in the command line... but I'll have to verify if this _worked_!

docs/source/config-file.mdx
mabuyo: That's a good point, will remove!

docs/source/config-file.mdx
mabuyo: Can confirm that double quotes are needed, thanks for the catch!

DaleSeo (approved):
LGTM!

Issue #280, opened 2025-08-24
GraphQL API responding with 403 Forbidden for unauthenticated operations


kedgm:
Hi, I'm trying to use the Apollo MCP Server with OAuth to connect to a GraphQL API. I'm getting 403 Forbidden errors when the MCP server tries to perform unauthenticated operations. It looks like it's automatically adding a bearer token to all outgoing requests, which causes the GraphQL API to reject operations that shouldn't need authentication. Is there a way to set up the MCP Server so that bearer tokens aren't included in requests for unauthenticated operations?

nicholascioli:
Thanks for reaching out. Could you share some more details about your setup? Do you have auth enabled and if so, what does your config look like? The MCP server only forwards tokens in one of two cases:

- Your [configuration](https://www.apollographql.com/docs/apollo-mcp-server/command-reference#configuration-options) has some headers statically configured using the `headers` key
- You have auth configured and it passes along the validated token once you finish the authorization flow.

Let us know which case reflects your environment and we can go from there.

kedgm:
Hi @nicholascioli, thanks for your reply. Yes, we have authentication enabled. Here's our current config:

endpoint: <REDACTED: GraphQL API URL>
transport:
  type: streamable_http
  address: 0.0.0.0
  port: 8080
  auth:
    servers:
      - <REDACTED: Authorization Server URL>
    audiences:
      - mcp-server
    resource: <REDACTED: MCP Server URL>
    scopes:
      - read:mcp
schema:
  source: local
  path: /schema.graphql
operations:
  source: local
  paths:
    - /operations
introspection:
  execute:
    enabled: true
  introspect:
    enabled: true
  search:
    enabled: true
logging:
  level: debug

Based on your explanation, our setup matches the second scenario: the MCP server forwards the validated token after completing the authorization flow. The challenge is that our GraphQL API initiates its own authorization checks whenever the Authorization header is present, even for operations that should be accessible without authentication.

kedgm:
@nicholascioli Any updates on this issue?

alocay:
I may not fully understand the scenario but I think it isn't possible to have the bearer token to only be sent for only some operations. @nicholascioli @swcollard do you know if this is possible? I thought it was an all or nothing situation but please correct me if I'm wrong.

swcollard:
@alocay you are correct. As implemented, if an authenticated request is sent to the MCP server with a valid token, it will be propagated to all requests to the GraphQL API.

[...]

PR #336, opened 2025-09-08
feat: Configuration for disabling authorization token passthrough

swcollard:
Auth in the MCP server works by having the user go through the OAuth flow to obtain an auth token. This auth token is then validated and passed to the upstream API in Tool calls for accessing the resources in that API. There has been a request to not perform this passthrough of validated tokens in the event that the upstream API does not share the same auth methods as the mcp server.

This PR adds an optional new parameter to the config.transport.auth config object, `disable_auth_token_passthrough`, which is `false` by default, that when true, will no longer pass through validated Auth tokens to the GraphQL API.

Contains fix for https://github.com/apollographql/apollo-mcp-server/issues/280


crates/apollo-mcp-server/src/main.rs
```diff
@@ -125,6 +127,23 @@ async fn main() -> anyhow::Result<()> {
         .mutation_mode(config.overrides.mutation_mode)
         .disable_type_description(config.overrides.disable_type_description)
         .disable_schema_description(config.overrides.disable_schema_description)
+        .disable_auth_token_passthrough(match transport {
```
alocay, line 130: Can we shorten this a bit via:
```
match transport {
    apollo_mcp_server::server::Transport::STDIO => false,
    _ { auth, .. } => auth.map(...).unwrap_or(false)
}
```

crates/apollo-mcp-server/src/main.rs
```diff
@@ -125,6 +127,23 @@ async fn main() -> anyhow::Result<()> {
...
+            apollo_mcp_server::server::Transport::StreamableHttp {
+                auth,
+                address: _,
+                port: _,
+            } => auth
+                .map(|a| a.disable_auth_token_passthrough)
+                .unwrap_or(true),
```
alocay: Should this be false?

swcollard: Yes, good catch.

crates/apollo-mcp-server/src/main.rs
swcollard: I can't quite get this syntax working? I think the match needs the enum name so it can pull the auth values from the struct. And the compiler isn't smart enough to pull it as an 'else' case?

PR #345, opened 2025-09-10
docs: remove placeholder for rover dev flags

DaleSeo:
`rover dev --mcp` takes only one argument now.

2025-09-10 at 10 59 20

### DIFF
--- a/docs/source/run.mdx
+++ b/docs/source/run.mdx
@@ -135,7 +135,7 @@ The Rover CLI is a tool for working with GraphQL APIs locally.
 You can use the [`rover dev`](/rover/commands/dev) command of Rover CLI `v0.35` or later to run an Apollo MCP Server instance alongside your local graph. Use the `--mcp` flag to start an MCP server and provide an optional configuration file.
 
 ```sh
-rover dev --mcp <PATH/TO/CONFIG> [...other rover dev flags]
+rover dev --mcp <PATH/TO/CONFIG>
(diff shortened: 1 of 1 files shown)


docs/source/run.mdx
```diff
@@ -135,7 +135,7 @@ The Rover CLI is a tool for working with GraphQL APIs locally.
 You can use the [`rover dev`](/rover/commands/dev) command of Rover CLI `v0.35` or later to run an Apollo MCP Server instance alongside your local graph. Use the `--mcp` flag to start an MCP server and provide an optional configuration file.
 
 ```sh
-rover dev --mcp <PATH/TO/CONFIG> [...other rover dev flags]
```
swcollard, line 138: This is just to remove the placeholder? There are other flags that are valid in this use case such as when a user wants to run both router and mcp using rover dev.

Such as,
```sh
rover dev --supergraph-config ./supergraph.yaml --mcp ./config.yaml
```

DaleSeo: @swcollard Thanks for pointing that out. I can see how others might interpret it differently now. I was mainly considering it from the `--mcp` usage angle, so I may have missed some broader context. Will close the PR.

swcollard: I think it is a good assumption that people reading this document would be focused on running the MCP server. But it might also imply that there is only one way to use rover dev?

Issue #344, opened 2025-09-09
[docs] Re-org of pages causing 404s - directs needed


ronan-apollo:
Noticing some of the prior urls no longer working, following #308 - can we add some redirects to bring these to the new location (or root, if there is no new location)

`https://www.apollographql.com/docs/apollo-mcp-server/guides`

DaleSeo:
Hey @ronan-apollo, confirming that PR #348 has resolved the issue. The ‎`/docs/apollo-mcp-server/guides` URL now properly redirects, and I've verified the `301 Moved Permanently` response with curl. 

```sh
$ curl -i https://www.apollographql.com/docs/apollo-mcp-server/guides
HTTP/2 301
age: 0
cache-control: no-cache
cache-status: "Netlify Durable"; fwd=bypass
cache-status: "Netlify Edge"; fwd=miss
date: Tue, 23 Sep 2025 20:23:45 GMT
location: https://www.apollographql.com/docs/apollo-mcp-server
netlify-vary: query
server: Netlify
strict-transport-security: max-age=31536000
x-content-type-options: nosniff
x-nf-request-id: 01K5W3Y652WBF6Z44J6MHBM5QG
content-length: 0
```

PR #348, opened 2025-09-10
Redirect `/docs/apollo-mcp-server/guides` to fix 404

swcollard:
This page is currently 404'ing https://www.apollographql.com/docs/apollo-mcp-server/guides

Attempting to add a redirect to the mcp server docs homepage.

https://github.com/apollographql/apollo-mcp-server/issues/344

### DIFF
--- a/docs/source/index.mdx
+++ b/docs/source/index.mdx
@@ -3,6 +3,7 @@ title: Apollo MCP Server
 subtitle: Enable graph-based API orchestration with AI
 redirectFrom:
     - /apollo-mcp-server/user-guide
+    - /apollo-mcp-server/guides
 ---
 
 <PreviewFeature>


swcollard:
With this change, the deploy preview, https://www.apollographql.com/docs/deploy-preview/05fbab664fb9eb98dfaad29c/apollo-mcp-server/guides

Will redirect to drop the /guides

mabuyo:
Was this a page that existed previously?

esilverm:
> Was this a page that existed previously?

This was the former "User Guide" page that was split up into multiple pages

mabuyo:
~Ahh I thought that had been under `/user-guide`, which we already redirect from. Was there a link you clicked from somewhere that led to `/guides`? cc @ronan-apollo re: original issue.~

~I ask because down the road we'll probably have content under `/guides`.~

Edit: Never mind, just verified it was under `/guides` index file 🤦🏽‍♀️

PR #350, opened 2025-09-11
feat: Enhance tool descriptions

DaleSeo:
We received feedback from DevRel that the descriptions of the introspect and search MCP tools are too simple for AI models to identify the intended tools and optimize the context window. This PR enhances the descriptions of these tools to offer clearer guidance for AI models on efficient GraphQL schema exploration patterns.

Before|After
-|-
2025-09-11 at 10 45 52|2025-09-11 at 11 00 09

### DIFF
--- a/crates/apollo-mcp-server/src/introspection/tools/introspect.rs
+++ b/crates/apollo-mcp-server/src/introspection/tools/introspect.rs
@@ -158,7 +158,7 @@ mod tests {
 
     #[rstest]
     #[tokio::test]
-    async fn test_tool_description_non_minified(schema: Valid<Schema>) {
+    async fn test_introspect_tool_description_is_not_minified(schema: Valid<Schema>) {
         let introspect = Introspect::new(Arc::new(Mutex::new(schema)), None, None, false);
 
         let description = introspect.tool.description.unwrap();
@@ -176,7 +176,9 @@ mod tests {
 
     #[rstest]
     #[tokio::test]
-    async fn test_tool_description_minified(schema: Valid<Schema>) {
+    async fn test_introspect_tool_description_is_minified_with_an_appropriate_legend(
+        schema: Valid<Schema>,
+    ) {
         let introspect = Introspect::new(Arc::new(Mutex::new(schema)), None, None, true);
 
         let description = introspect.tool.description.unwrap();
--- a/crates/apollo-mcp-server/src/introspection/tools/search.rs
+++ b/crates/apollo-mcp-server/src/introspection/tools/search.rs
@@ -249,7 +249,7 @@ mod tests {
 
     #[rstest]
     #[tokio::test]
-    async fn test_tool_description_non_minified(schema: Valid<Schema>) {
+    async fn test_search_tool_description_is_not_minified(schema: Valid<Schema>) {
         let schema = Arc::new(Mutex::new(schema));
         let search = Search::new(schema.clone(), false, 1, 15_000_000, false)
             .expect("Failed to create search tool");


crates/apollo-mcp-server/src/introspection/tools/introspect.rs
```diff
@@ -140,3 +136,53 @@ fn tool_description(
...
+    fn schema() -> Valid<Schema> {
+        Schema::parse(TEST_SCHEMA, "schema.graphql")
+            .expect("Failed to parse test schema")
+            .validate()
+            .expect("Failed to validate test schema")
+    }
+
+    #[rstest]
+    #[tokio::test]
+    async fn test_tool_description_non_minified(schema: Valid<Schema>) {
```
alocay: Can we use better test names are that a bit more descriptive? I like following this guide for test name conventions https://enterprisecraftsmanship.com/posts/you-naming-tests-wrong/

DaleSeo: Isn't the current test name already close to the conventions you linked?  If you have any specific suggestions for renaming, just let me know and I'll update it.

alocay: So the test is verifying that the description is not minified and doesn't contain the minified legend. I would use a test name that indicates this at first glance, maybe something like `introspect_tool_description_is_not_minified`. Not need to add "test" in the name IMO since thats a bit superfluous.

crates/apollo-mcp-server/src/introspection/tools/introspect.rs
```diff
@@ -140,3 +136,53 @@ fn tool_description(
...
+
+        assert!(
+            description
+                .contains("Get information about a given GraphQL type defined in the schema")
+        );
+        assert!(description.contains("Instructions: Use this tool to explore the schema"));
+        // Should not contain minification legend
+        assert!(!description.contains("T=type,I=input"));
+        // Should mention conditional search tool usage
+        assert!(description.contains("If the search tool is also available"));
```
alocay, line 174: nit: you could argue that this last condition is testing something else entirely and not related to the minification stuff. You could technically split this in a separate test called like `instrospect_description_mentions_search_tool_usage`. I find it helpful to have small very directed unit tests. It helps quickly determining what is broken in tests in the future.

crates/apollo-mcp-server/src/introspection/tools/introspect.rs
```diff
@@ -140,3 +136,53 @@ fn tool_description(
...
+    }
+
+    #[rstest]
+    #[tokio::test]
+    async fn test_tool_description_minified(schema: Valid<Schema>) {
```
alocay: this name could be `introspect_tool_description_is_minified_with_an_appropriate_legend` or something

crates/apollo-mcp-server/src/introspection/tools/search.rs
```diff
@@ -246,4 +246,37 @@ mod tests {
             "Expected to find the createUser mutation in search results"
         );
     }
+
+    #[rstest]
+    #[tokio::test]
+    async fn test_tool_description_non_minified(schema: Valid<Schema>) {
```
alocay: nit: ditto here and below, can use something like `search_tool_description_is_not_minified`.

crates/apollo-mcp-server/src/introspection/tools/search.rs
```diff
@@ -246,4 +246,37 @@ mod tests {
...
+        let search = Search::new(schema.clone(), false, 1, 15_000_000, false)
+            .expect("Failed to create search tool");
+
+        let description = search.tool.description.unwrap();
+
+        assert!(
+            description
+                .contains("Search a GraphQL schema for types matching the provided search terms")
+        );
+        assert!(description.contains("Instructions: If the introspect tool is also available"));
```
alocay, line 263: nit: these two asserts aren't related to minified stuff. You could split these two into a separate test like `search _tool_description_contains_expected_instructions` or something of that nature. That way if it breaks in the future we don't have debug to see if it's breaking due to the minification related asserts or these.

crates/apollo-mcp-server/src/introspection/tools/introspect.rs
DaleSeo: @alocay I get your naming preference, but adding `test_` as a prefix for test methods is a common practice in our codebase and might even be a company standard, like with Router. I'm concerned that removing it here could introduce inconsistency so I'll update it to ‎`test_introspect_tool_description_is_not_minified` for now.

If you’d like to revisit this convention, I recommend proposing the change to the team and, if there's consensus, we can bulk update test names across the codebase to keep things consistent.

swcollard (approved):

PR #338, opened 2025-09-08
docs: Draft documentation and user guide for mcp otel telemetry

swcollard:
Starting a first draft of documentation for the Telemetry feature. This includes two pages.

`telemetry.mdx`: Top level documentation for the telemetry feature. Includes a description of what the feature is for and how to use it. Contains full reference to all metrics being emitted.

`config-file.mdx`: The config file reference has been updated to include all options added as part of telemetry

PR #352, opened 2025-09-11
Telemetry Spec

kbychu:
Add telemetry spec for documentation and requirement discussion. Don't need to merge this but wanted to start writing things down.


DaleSeo:
@kbychu This PR might be a duplicate of #338

specs/telemetry.md
```diff
@@ -0,0 +1,81 @@
...
+|                           | `apollo_mcp.schema.load_success` / `apollo_mcp.schema.load_failure`     | Counter         | schema_source                                                      | Schema load status                                                      | Must Have     |
+|                           | `apollo_mcp.schema.size`                                                | Gauge           | —                                                                  | # of types/fields in schema                                             | Should Have   |
+|                           | `apollo_mcp.version.info`                                               | Attribute/Event | server_version, schema_hash, manifest_version, manifest_source        | Server binary version, GraphQL schema hash, manifest version, manifest type (persisted_query/operation_collection) | Must Have     |
+| **Usage**                 | `apollo.mcp.calls`                                                      | Counter         | tool_name, success, error_code, client_type                       | Total tool invocations                                                  | Must Have     |
```
swcollard: What are your thoughts for `client_type`? AFAIK, That data is unfortunately only available to us during the initialization call.

specs/telemetry.md
lrlna, line 1: Any particular reason we are using `apollo_mcp` instead of `apollo.mcp` like we do with `apollo.router` ([docs](https://www.apollographql.com/docs/graphos/routing/observability/telemetry/instrumentation/standard-instruments#standard-metric-instruments))?

specs/telemetry.md
```diff
@@ -0,0 +1,81 @@
+# Apollo MCP Server Telemetry Spec
+
+| Category                  | Metric / Trace / Event                                                   | Type            | Attributes                                                         | Notes                                                                   | Priority      |
+|---------------------------|--------------------------------------------------------------------------|-----------------|--------------------------------------------------------------------|-------------------------------------------------------------------------|---------------|
+| **Configuration**         | `apollo_mcp.config.load_success`                                        | Counter         | error_type                                                         | Successful config / startup loads                                       | Must Have     |
+|                           | `apollo_mcp.config.load_failure`                                        | Counter         | error_type                                                         | Failed startup (bad schema, manifest, endpoint)                         | Must Have     |
```
lrlna: It's a bit strange to have two separate metrics for `success` and `failure`. You'd usually have a `status` attribute on a metric, like a `apollo.mcp.config.load.count`, similar to `system.process.count` described in [otel docs](https://opentelemetry.io/docs/specs/semconv/system/system-metrics/#metric-systemprocesscount). You can have an additional attribute of `error` for the error message or `code` for the error code.

specs/telemetry.md
```diff
@@ -0,0 +1,81 @@
+# Apollo MCP Server Telemetry Spec
+
+| Category                  | Metric / Trace / Event                                                   | Type            | Attributes                                                         | Notes                                                                   | Priority      |
+|---------------------------|--------------------------------------------------------------------------|-----------------|--------------------------------------------------------------------|-------------------------------------------------------------------------|---------------|
+| **Configuration**         | `apollo_mcp.config.load_success`                                        | Counter         | error_type                                                         | Successful config / startup loads                                       | Must Have     |
```
swcollard: We have mostly been using `apollo.mcp` instead of `apollo_mcp` in our implementation so far as the namespacing.

specs/telemetry.md
```diff
@@ -0,0 +1,81 @@
...
+|                           | `apollo_mcp.schema.size`                                                | Gauge           | —                                                                  | # of types/fields in schema                                             | Should Have   |
+|                           | `apollo_mcp.version.info`                                               | Attribute/Event | server_version, schema_hash, manifest_version, manifest_source        | Server binary version, GraphQL schema hash, manifest version, manifest type (persisted_query/operation_collection) | Must Have     |
+| **Usage**                 | `apollo.mcp.calls`                                                      | Counter         | tool_name, success, error_code, client_type                       | Total tool invocations                                                  | Must Have     |
+|                           | `apollo.mcp.calls.latency`                                              | Histogram       | tool_name, success, error_code, client_type                       | End-to-end request latency                                              | Must Have     |
```
lrlna: are you going to return the latency for all the calls, or a single call is going to be recorded as a single call in the histogram? If the latter, the semantic convention advises to use singular noun ([docs](https://www.apollographql.com/docs/graphos/routing/configuration/cli#-graph-artifact-reference)). 

With `apollo.mcp.calls.latency`, you should probably use `apollo.mcp.call.duration`, much like `http.server.request.duration` does ([docs](https://opentelemetry.io/docs/specs/semconv/http/http-metrics/#metric-httpserverrequestduration)).

With `apollo.mcp.calls` counter, I am not sure if you need a counter, since you would have the `apollo.mcp.call.duration` histogram, which you could just do a `count` on, for example:

Screenshot 2025-09-15 at 19 09 49

specs/telemetry.md
```diff
@@ -0,0 +1,81 @@
...
+|                           | `apollo.mcp.operation.calls`                                            | Counter         | tool_name, success, error_code, client_type, operation_name       | # of backend GraphQL operations executed                                | Must Have     |
+|                           | `apollo.mcp.operation.latency`                                          | Histogram       | tool_name, success, error_code, client_type, operation_name       | Latency of GraphQL backend call (excludes tool overhead)               | Must Have     |
```
lrlna: Similar to the comment about `apollo.mcp.calls.latency`, this should be also be `apollo.mcp.operation.duration`, and you don't need a separate counter.

specs/telemetry.md
```diff
@@ -0,0 +1,81 @@
...
+|                           | `apollo_mcp.operation.type.mix`                                         | Counter         | query, mutation, subscription                                      | Breakdown of operation types                                            | Should Have   |
```
lrlna: The operation type should live as an attribute on the `apollo.mcp.operation.duration`.

specs/telemetry.md
```diff
@@ -0,0 +1,81 @@
...
+|                           | `apollo_mcp.responses.size`                                             | Histogram       | tool_name, client_type                                             | Size of responses (bytes)                                               | Should Have   |
+|                           | `apollo_mcp.responses.characters`                                        | Histogram       | tool_name, client_type                                             | Character count of response payloads (proxy for token estimation)      | Nice to Have  |
```
lrlna: is this server's response or client's response? That should be specified. If it's the _actual_ `http.server.response.body.size` or `http.client.response.body.size`, you should be using the existing standard instrument ([server docs](https://opentelemetry.io/docs/specs/semconv/http/http-metrics/#metric-httpserverresponsebodysize), [client docs](https://opentelemetry.io/docs/specs/semconv/http/http-metrics/#metric-httpclientresponsebodysize)). 

If it's the response size of something else, you should use the singular `response`.

specs/telemetry.md
```diff
@@ -0,0 +1,81 @@
...
+|                           | `apollo_mcp.graphql.backend.errors`                                     | Counter         | status_code, operation_name                                        | Errors from upstream GraphQL API                                       | Must Have     |
+|                           | `apollo_mcp.transport.errors`                                           | Counter         | error_type                                                         | Invalid JSON-RPC, dropped connections                                  | Should Have   |
+|                           | `apollo_mcp.auth.failures`                                              | Counter         | reason, client_type                                                | Authentication failures                                                 | Must Have     |
+|                           | `apollo_mcp.timeouts`                                                   | Counter         | tool_name, client_type                                             | Tool or backend operation timed out                                     | Must Have     |
```
lrlna: These are all just error kinds that should live on the standard instrument implementation for probably `http.server.request.duration` under its `error.type` attribute ([docs](https://opentelemetry.io/docs/specs/semconv/http/http-metrics/#metric-httpserverrequestduration)) with additional instrumentation for things like `tool_name`.

**As a heads up**, there is a **huge** cardinality associated with `operation_name`, which is why the router doesn't ship with that instrumentation by default. It might cost users **a lot** of money to have this reported, and it will also cost us a **a lot** of money if we are additionally ingesting that data.

specs/telemetry.md
```diff
@@ -0,0 +1,81 @@
...
+|                           | Span: `graphql.operation`                                               | Trace           | operation_name, latency, success, error_code                      | Child span for backend GraphQL operation                                | Must Have     |
+|                           | Span: `serialization`                                                   | Trace           | size_bytes, latency                                               | Encoding/decoding JSON-RPC overhead                                     | Nice to Have  |
+| **Events**                | `apollo_mcp.client.connected`                                           | Event           | client_type                                                        | Client connection established                                           | Should Have   |
+|                           | `apollo_mcp.client.disconnected`                                        | Event           | client_type                                                        | Client disconnected                                                     | Should Have   |
```
lrlna: This could _also_ be a gauge, so you can monitor the number of client connections at any given time. You can probably mirror `http.client.open_connections` here ([docs](https://opentelemetry.io/docs/specs/semconv/http/http-metrics/#metric-httpclientopen_connections)). This can be called something like `apollo.mcp.client.open_connections`. I am not sure if you would want both a gauge and an event though.

specs/telemetry.md
```diff
@@ -0,0 +1,81 @@
...
+|                           | `apollo_mcp.config.reload`                                              | Event           | schema_source, version_hash                                        | Config/schema/manifest/collection reload                                | Nice to Have  |
```
lrlna: How is this different from the first item on the list `apollo.mcp.config.load`? As in, if you wanted to track reload specifically, I'd add that as an attribute to to `apollo.mcp.config.load` and extract it out of that.

specs/telemetry.md
lrlna: Additionally, I'd take a look at [this doc](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-events/), while it's for gen AI, I think it'll have nice standards to follow for things like tokens.

specs/telemetry.md
```diff
@@ -0,0 +1,81 @@
...
+|                           | `apollo_mcp.auth.failed`                                                | Event           | client_type, reason                                                | Auth failure                                                            | Must Have     |
+| **Resource Usage**        | `process.cpu.usage`                                                     | Gauge           | —                                                                  | CPU usage of MCP server process                                         | Nice to Have  |
+|                           | `process.memory.usage`                                                  | Gauge           | —                                                                  | Memory usage                                                            | Nice to Have  |
```
lrlna: are you able to get this information directly in the mcp server implementation? that's pretty cool!

specs/telemetry.md
```diff
@@ -0,0 +1,81 @@
...
+| **Events**                | `apollo_mcp.client.connected`                                           | Event           | client_type                                                        | Client connection established                                           | Should Have   |
+|                           | `apollo_mcp.client.disconnected`                                        | Event           | client_type                                                        | Client disconnected                                                     | Should Have   |
+|                           | `apollo_mcp.config.reload`                                              | Event           | schema_source, version_hash                                        | Config/schema/manifest/collection reload                                | Nice to Have  |
+|                           | `apollo_mcp.auth.failed`                                                | Event           | client_type, reason                                                | Auth failure                                                            | Must Have     |
```
lrlna: I would track this as an error type from `http.server.request.duration` I mentioned a few comments ago. But if you do want a specific event, I'd count both success and failures under a `status` attribute.

specs/telemetry.md
```diff
@@ -0,0 +1,81 @@
...
+| **Resource Usage**        | `process.cpu.usage`                                                     | Gauge           | —                                                                  | CPU usage of MCP server process                                         | Nice to Have  |
+|                           | `process.memory.usage`                                                  | Gauge           | —                                                                  | Memory usage                                                            | Nice to Have  |
+|                           | `network.bytes.sent` / `network.bytes.received`                         | Counter         | —                                                                  | Network traffic                                                         | Nice to Have  |
```
lrlna: This is usually tracked under `http.client|server.request|response.body.size`. Or are you proposing a different sort of aggregation of body size?

lrlna (commented):
I think you all should have one more look over a few of the OTEL docs before you officially decide on how you're going to be aggregating this data. I think people will find a few things quite surprising with what's currently in this PR, from both OTEL point of view, as well as what we have in the router. 

Docs that you might find helpful to read:
* [Naming conventions](https://opentelemetry.io/docs/specs/semconv/general/naming/)
* [HTTP metrics](https://opentelemetry.io/docs/specs/semconv/http/http-metrics/)
* [System metrics](https://opentelemetry.io/docs/specs/semconv/system/system-metrics/)
* [Gen AI metrics](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-metrics/), while not specific to MCP, still provides good inspiration
    * [Gen AI events](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-events/)

kbychu:
> @kbychu This PR might be a duplicate of #338

@DaleSeo absolutely, this is not meant to be a replicate of docs, but I wanted to share thoughts on all the potential telemetry we might want to have and have a place for discussion and things to reason about. I'll close this PR at some point in the near future.

kbychu:
Thanks @lrlna!

specs/telemetry.md
swcollard: > a single call is going to be recorded as a single call in the histogram

I'll update this to use the singular `call`

Agree that the count can be derived via the histogram. Does it follow convention to not have a standalone counter in that case?

specs/telemetry.md
swcollard: This I'm not sure of? We are using many of the same SDKs as the router and the process information might be better sourced from something like Kubernetes or wherever the system is hosted

specs/telemetry.md
swcollard: No, we should be using the semantic conventions for http metrics and are already logging `http.server.request.size`, etc.

I'll update these to reflect that.

specs/telemetry.md
```diff
@@ -0,0 +1,80 @@
...
+|                           | `apollo.mcp.responses.characters`                                        | Histogram       | tool_name, client_type                                             | Character count of response payloads (proxy for token estimation)      | Nice to Have  |
+|                           | `apollo.mcp.clients.active`                                             | Gauge           | —                                                                  | # of active MCP clients                                                 | Must Have     |
+|                           | `apollo.mcp.concurrency.current_requests`                               | Gauge           | —                                                                  | # of concurrent tool executions                                         | Should Have   |
+| **Errors / Reliability**  | `apollo.mcp.requests.errors`                                            | Counter         | error_type, tool_name, client_type                                | Failed tool calls (generic catch-all)                                  | Must Have     |
```
swcollard: I think these errors should live within the counters/durations on the toll calls rather than be a separate metric

specs/telemetry.md
lrlna: No, in fact we used to have a counter, [and had to remove it in order to follow convention in router 2.0](https://www.apollographql.com/docs/graphos/routing/upgrade/from-router-v1#removed-metrics).

As I was double checking this, I noticed that [otel spec has `gen_ai.client.operation.duration`](https://www.apollographql.com/docs/graphos/routing/upgrade/from-router-v1#removed-metrics) (and the server equivalent). Is `call` equivalent to `client.operation`?

specs/telemetry.md
lrlna: ah, yea in that case k8s or the container will have this info instead. the router cannot access it's own memory and cpu utilisation.

lrlna:
@swcollard i also just remembered we have [a dev doc on metrics in the router](https://github.com/apollographql/router/blob/dev/dev-docs/metrics.md), in case you haven't seen it (it's a bit hidden away). might be useful for some of our own conventions.

specs/telemetry.md
kbychu: Just my vibe coded oversight!

kbychu:
Closing PR, it's served its purpose.

PR #355, opened 2025-09-12
Add load balancer configuration details for Apollo MCP

lennyburdette:
Added instructions for configuring load balancers with Apollo MCP Server to ensure session affinity.

### DIFF
--- a/docs/source/deploy.mdx
+++ b/docs/source/deploy.mdx
@@ -56,3 +56,9 @@ docker run \
 ```
 
 To learn more, review the [Apollo Runtime container documentation](/graphos/routing/self-hosted/containerization/docker).
+
+### Using a load balancer
+
+Because [MCP is a stateful protocol](https://modelcontextprotocol.io/docs/learn/architecture#lifecycle-management), the default configuration of most load balancers is incorrect. When the MCP client initializes a session with Apollo MCP Server, it receives a session identifier unique to that server instance. If the load balancer routes subsequent requests to a different instance, Apollo MCP Server will reject the request as unauthorized. 
+
+To run Apollo MCP Server in a high-availability, multi-instance setup, you must configure your load balancer to use session affinity ("sticky sessions") based on the `mcp-session-id` header. Load balancers offered by major cloud vendors do not support header-based affinity, so you will need software like Nginx, HAProxy, or Istio/Envoy Proxy in front of your Apollo MCP instances.
--- a/docs/source/deploy.mdx
+++ b/docs/source/deploy.mdx
@@ -63,6 +63,6 @@ Because [MCP is a stateful protocol](https://modelcontextprotocol.io/docs/learn/
 
 When the MCP client initializes a session with Apollo MCP Server, it receives a session identifier unique to that server instance through the `mcp-session-id` header. You must enable session affinity ("sticky sessions") in your load balancer so that all requests that share the same `mcp-session-id` are routed to the same backend instance. 
 
-If the load balancer routes subsequent requests to a different instance, Apollo MCP Server will reject the request as unauthorized.
+If the load balancer routes subsequent requests to a different instance, Apollo MCP Server rejects the request because it doesn't recognize the session id.
 
-Many load balancers offered by major cloud vendors don't support header-based session affinity. If yours does not, use software such as Nginx, HAProxy, or Envoy/Istio in front of your Apollo MCP Server instances.
+Many load balancers offered by major cloud vendors don't support header-based session affinity. If yours does not, use software such as Nginx, HAProxy, or Envoy/Istio in front of the Apollo MCP Server instances.
--- a/docs/source/deploy.mdx
+++ b/docs/source/deploy.mdx
@@ -59,6 +59,10 @@ To learn more, review the [Apollo Runtime container documentation](/graphos/rout
 
 ### Using a load balancer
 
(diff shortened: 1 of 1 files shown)


docs/source/deploy.mdx
```diff
@@ -56,3 +56,9 @@ docker run \
 ```
 
 To learn more, review the [Apollo Runtime container documentation](/graphos/routing/self-hosted/containerization/docker).
+
+### Using a load balancer
+
+Because [MCP is a stateful protocol](https://modelcontextprotocol.io/docs/learn/architecture#lifecycle-management), the default configuration of most load balancers is incorrect. When the MCP client initializes a session with Apollo MCP Server, it receives a session identifier unique to that server instance. If the load balancer routes subsequent requests to a different instance, Apollo MCP Server will reject the request as unauthorized.
```
swcollard: >Apollo MCP Server will reject the request as unauthorized. 

Is that what we're seeing even if Auth is disabled on the server?

lennyburdette: yeah, i believe this is an "i don't know this mcp-session-id" authorization error.

DaleSeo: @lennyburdette To be exact, was it 406 Not Acceptable instead of 401 Unauthorized?

- https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/406
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/401

docs/source/deploy.mdx
```diff
@@ -56,3 +56,9 @@ docker run \
 ```
 
 To learn more, review the [Apollo Runtime container documentation](/graphos/routing/self-hosted/containerization/docker).
+
+### Using a load balancer
+
+Because [MCP is a stateful protocol](https://modelcontextprotocol.io/docs/learn/architecture#lifecycle-management), the default configuration of most load balancers is incorrect. When the MCP client initializes a session with Apollo MCP Server, it receives a session identifier unique to that server instance. If the load balancer routes subsequent requests to a different instance, Apollo MCP Server will reject the request as unauthorized. 
+
+To run Apollo MCP Server in a high-availability, multi-instance setup, you must configure your load balancer to use session affinity ("sticky sessions") based on the `mcp-session-id` header. Load balancers offered by major cloud vendors do not support header-based affinity, so you will need software like Nginx, HAProxy, or Istio/Envoy Proxy in front of your Apollo MCP instances.
```
mabuyo:
```suggestion
Because [MCP is a stateful protocol](https://modelcontextprotocol.io/docs/learn/architecture#lifecycle-management), you need to configure your load balancer to keep each session on the _same server instance_.

When the MCP client initializes a session with Apollo MCP Server, it receives a session identifier unique to that server instance through the `mcp-session-id` header. You must enable session affinity ("sticky sessions") in your load balancer so that all requests that share the same `mcp-session-id` are routed to the same backend instance. 

If the load balancer routes subsequent requests to a different instance, Apollo MCP Server will reject the request as unauthorized.

Many load balancers offered by major cloud vendors don't support header-based session affinity. If yours does not, use software such as Nginx, HAProxy, or Envoy/Istio in front of your Apollo MCP Server instances.
```

A couple rewrites to help readability:
- direct call out at the start on what they need to do
- remove “incorrect by default” framing
- split up into multiple paragraphs

docs/source/deploy.mdx
lennyburdette: i have no idea, the cursor logs don't have a status code

DaleSeo: @lennyburdette In my logs, I saw errors like `Not Acceptable: Client must accept both application/json and text/event-stream`, which aligns with HTTP 406. We don't have to mention a specific status code. How about we update this to something like this to avoid confusion?

```diff
- Apollo MCP Server will reject the request as unauthorized. 
+ Apollo MCP Server will reject the request as not acceptable. 
```

lennyburdette: yeah i think that's a symptom, not a cause. i need to confirm this, but i think the flow is:

1. client send initialization request
2. server instance A mints an mcp-session-id and responds
3. client sends a tools/list request with mcp-session-id, load balancer sends it to server instance B
4. server instance B says "nope" and responds with something we need to confirm (401?)
5. client tries to reinitialize with a new request. *this one* doesn't have the correct `accept` headers, so it's rejected by any instance

so the problem is actually step 3, but there might be a client bug on the reinitialization request too.

mabuyo (approved):
LGTM on docs-side unless you'd like to further clarify that 401/406 error!

docs/source/deploy.mdx
DaleSeo: @lennyburdette I double-checked the Slack thread and confirmed that you did encounter 406:
https://apollograph.slack.com/archives/C08LW3Z8PJ4/p1757615864947459?thread_ts=1757615514.990019&cid=C08LW3Z8PJ4

Looking at the rmcp code, it appears the server checks for a 406 Not Acceptable before checking for a 401 Unauthorized:
https://github.com/modelcontextprotocol/rust-sdk/blob/f8a747eaca997c8382e7bfbc7de3a859700c51f5/crates/rmcp/src/transport/streamable_http_server/tower.rs#L156-L185

lennyburdette: yeah we definitely saw that, but i still think it's a symptom. why is the client making an invalid request? it must be getting into a weird state after the server rejects a request because of the unknown session id. 

regardless, i don't think we need to get too specific. i can drop "unauthorized" and say something more generic like "... will reject the request because it doesn't recognize the session id."

PR #358, opened 2025-09-12
feat: adding ability to omit attributes for traces and metrics

alocay:
Adding ability to configure which attributes are omitted from telemetry traces and metrics.

1. Using a Rust build script (`build.rs`) to auto-generate telemetry attribute code based on the data found in `telemetry.toml`.
2. Utilizing an enum for attributes so typos in the config file raise an error.
3. Omitting trace attributes by filtering it out in a custom exporter.
4. Omitting metric attributes by indicating which attributes are allowed via a view.
5. Created `telemetry_attributes.rs` to map `TelemetryAttribute` enum to a OTEL `Key`.

The `telemetry.toml` file includes attributes (both for metrics and traces) as well as list of metrics gathered. An example would look like the following:
```
[attributes.apollo.mcp]
my_attribute = "Some attribute info"

[metrics.apollo.mcp]
some.count = "Some metric count info"
```
This would generate a file that looks like the following:
```
/// All TelemetryAttribute values
pub const ALL_ATTRS: &[TelemetryAttribute; 1usize] = &[
    TelemetryAttribute::MyAttribute
];
#[derive(Debug, ::serde::Deserialize, ::schemars::JsonSchema,, Clone, Eq, PartialEq, Hash, Copy)]
pub enum TelemetryAttribute {
    ///Some attribute info
    #[serde(alias = "my_attribute")]
    MyAttribute,
}
impl TelemetryAttribute {
    /// Supported telemetry attribute (tags) values
    pub const fn as_str(&self) -> &'static str {
        match self {
            TelemetryAttribute::MyAttribute => "apollo.mcp.my_attribute",
        }
    }
}
#[derive(Debug, ::serde::Deserialize, ::schemars::JsonSchema,, Clone, Eq, PartialEq, Hash, Copy)]
pub enum TelemetryMetric {
    ///Some metric count info
    #[serde(alias = "some.count")]
    SomeCount,
}
impl TelemetryMetric {
    /// Converts TelemetryMetric to &str
    pub const fn as_str(&self) -> &'static str {
        match self {
            TelemetryMetric::SomeCount => "apollo.mcp.some.count",
        }
    }
}
```
An example configuration that omits `tool_name` attribute for metrics and `request_id` for tracing would look like the following:
```
telemetry:
  exporters:
    metrics:
      otlp:
        endpoint: "http://localhost:4317"
        protocol: "grpc"
      omitted_attributes:
        - tool_name
    tracing:
      otlp:
        endpoint: "http://localhost:4317"
        protocol: "grpc"
      omitted_attributes:
        - request_id
```


crates/apollo-mcp-server/src/server/states/running.rs
```diff
@@ -189,7 +194,7 @@ impl ServerHandler for Running {
         Ok(self.get_info())
     }
 
-    #[tracing::instrument(skip(self, context), fields(tool_name = request.name.as_ref(), request_id = %context.id.clone()))]
+    #[tracing::instrument(skip(self, context), fields(apollo.mcp.attribute.tool_name = request.name.as_ref(), apollo.mcp.attribute.request_id = %context.id.clone()))]
```
alocay: Unfortunately I have not figured out a way to use the const variables from the generated code in this `fields()` section.

crates/apollo-mcp-server/src/runtime/filtering_exporter.rs
```diff
@@ -0,0 +1,236 @@
...
+
+            fn set_resource(&mut self, _resource: &Resource) {
+                assert!(true);
+            }
+        }
+
+        let mock_exporter = TestExporter {};
+
+        let mut filtering_exporter = FilteringExporter::new(mock_exporter, HashSet::new());
+        filtering_exporter.set_resource(&Resource::builder_empty().build());
```
swcollard, line 231: Is this test missing an assert? Or does not throwing an error show that it's working?

crates/apollo-mcp-server/telemetry.toml
```diff
@@ -0,0 +1,15 @@
+[apollo.mcp.attribute]
+tool_name = "The tool name"
+operation_id = "The operation id"
+operation_type = "The operation type"
```
swcollard: This probably reflects more on my previous implementation. But we might want to add the detail that operation_id is either the persisted query ID, the graphql operations name, or unknown for an unnamed operation.

The type is either persisted_query or operation

crates/apollo-mcp-server/src/runtime/filtering_exporter.rs
alocay: The assert is within the innner `set_resource` call which has:
```
assert!(true);
```
It's not an amazing test but since `set_resource` doesn't return anything I just assert within the set_resource function and call it day. Although this isn't great since calling nothing will also pass but I could not think of another way to ensure this. If any other inner function is called, it'll fail.

crates/apollo-mcp-server/telemetry.toml
alocay: Sure I'll update it. This is mainly used for in code doc string atm.

crates/apollo-mcp-server/src/server/states/running.rs
```diff
@@ -189,7 +190,7 @@ impl ServerHandler for Running {
         Ok(self.get_info())
     }
 
-    #[tracing::instrument(skip(self, context), fields(tool_name = request.name.as_ref(), request_id = %context.id.clone()))]
+    #[tracing::instrument(skip(self, context), fields(apollo.mcp.tool_name = request.name.as_ref(), apollo.mcp.request_id = %context.id.clone()))]
```
swcollard: I think we might want to skip the `request` here which could potentially have PII or other sensitive data? Thought it has been useful for debugging while developing.

swcollard (approved):

crates/apollo-mcp-server/src/server/states/running.rs
alocay: Done

PR #365, opened 2025-09-16
Update SDL handling in sdl_to_api_schema function

lennyburdette:
`::new` uses "default" supergraph specs, which does not include Connectors, so all connector supergraph result in an error and are parsed as standard GraphQL. This ends up exposing join directives, which is a huge waste of tokens.


lennyburdette:
@swcollard done!

esilverm (approved):

Issue #346, opened 2025-09-10
Providing custom instructions/hints for GraphQL query generation


michalpie:
When working with large and complex GraphQL schemas, the model can sometimes struggle to generate accurate queries based solely on introspection and search tools. Our schema, for instance, has several non-obvious patterns and relationships that are difficult for the model to infer on its own.

Based on our experiments, providing the model with specific hints or instructions about the schema's structure and common query patterns dramatically improves the accuracy and reliability of the generated GraphQL queries.

I'd like to use a mechanism to inject this custom guidance into the model's context, if such mechanism exists. Otherwise to propose it:

Allow users to override the default descriptions for the built-in tools, especially the execute tool. By modifying its description, we could embed our hints directly into the context the model uses when it considers executing a GraphQL query.

For example, we could change the execute tool's description from its default to something like:
"Executes a GraphQL query. IMPORTANT HINT: To get a user's active cart, always query the 'carts' connection with a filter, like 'carts(where: { status: ACTIVE })'. The 'shoppingCart' field is deprecated."

Example Use Case
Imagine a schema where fetching a user's active shopping cart requires a specific filter:

Naive/Incorrect Query: query { user(id: "123") { shoppingCart { id } } }

Correct Query: query { user(id: "123") { carts(where: { status: ACTIVE }) { id } } }

Without guidance, the model might default to the simpler but incorrect field. A custom hint could explicitly guide it to use the correct pattern, preventing errors and improving the user experience.

swcollard:
Thanks for this feedback and suggestion! The team is currently looking at different ways we can help the LLM craft better operations. We'll include this as an idea and let you know what our plans are. cc @esilverm

michalpie:
Great, this library is already fantastic.

Maybe you could add an option to pass user customizable hints to the tool descriptions (instead of overriding the whole description), which could be appended to the descriptions. 

Also, MCP level "instructions" field could be configured (but this is not enough as some clients like Copilot do not use it).

michalpie:
I have created this draft PR (https://github.com/apollographql/apollo-mcp-server/pull/359) with an initial implementation of hints for the execute tool only. If you are accept this approach, I can extend it to other tools.

kbychu:
@michalpie 

Thanks for this proposal! The shopping cart example really illustrates the challenge with deprecated fields and non-obvious patterns.
Have you considered putting these hints at the client level instead? For example:

Cursor: Add to project-level system prompts
[...]

PR #367, opened 2025-09-17
Minify: Add support for deprecated directive

esilverm:
In #346, it is mentioned that our tools aren't properly picking up `deprecated` directives when building and executing operations. This led me to dig deeper and find that our `minify` handling for the builtin tools does not include any directives in the minified output. This can potentially be harmful since it means the LLM doesn't know to exclude certain fields when building operations with `introspect`. 

In this PR, I've added the ability to minify deprecated directives with the inclusion of their `reason` argument to give better direction to the LLM when it works within this problem space.
Additionally, I've added some snapshot tests to the `minify` file so we can assert that the minified schema is built properly and includes the things we expect it to include.

I've also made sure to add proper coverage for the directive since it has recently been updated to be allowed to be placed on `INPUT_FIELD_DEFINITION` and `ARGUMENT_DEF` (though these technically fail schema validation with our current graphql-rs parser.

### DIFF
--- a/docs/source/define-tools.mdx
+++ b/docs/source/define-tools.mdx
@@ -156,6 +156,7 @@ Both the `introspect` and `search` tools support minification of their results t
 
 - **Type prefixes**: `T=type`, `I=input`, `E=enum`, `U=union`, `F=interface`
 - **Scalar abbreviations**: `s=String`, `i=Int`, `f=Float`, `b=Boolean`, `d=ID`
+- **Directive abbreviations**: `@D=deprecated`
 - **Type modifiers**: `!=required`, `[]=list`, `<>=implements`
 
 Example comparison:


crates/apollo-mcp-server/src/introspection/minify.rs
```diff
@@ -72,6 +72,39 @@ fn minify_input_object(input_object_type: &apollo_compiler::schema::InputObjectT
...
+        if let Some(minified_name) = directives_to_minify.get(directive.name.as_str()) {
+            if !directive.arguments.is_empty() {
+                // Since we're only handling @deprecated right now we can just add the reason and minify it.
+                // We should handle this more generically in the future.
+                let reason = directive
+                    .arguments
+                    .iter()
+                    .find(|a| a.name == "reason")
+                    .and_then(|a| a.value.as_str())
+                    .unwrap_or("No longer supported")
```
DaleSeo: GraphQL doesn't require a reason argument for the ‎`@deprecated` directive.

- Spec: https://spec.graphql.org/September2025/#sec--deprecated
- PR: https://github.com/graphql/graphql-spec/pull/1040

To be consistent with the spec, how about we allow for just ‎`@D`, instead of always adding a default reason? This could also help reduce the minified output size slightly.

crates/apollo-mcp-server/src/introspection/minify.rs
```diff
@@ -211,3 +253,27 @@ fn normalize_description(desc: &str) -> String {
...
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    const TEST_SCHEMA: &str = include_str!("tools/testdata/schema.graphql");
+
+    #[test]
+    fn test_minify_schema() {
```
DaleSeo, line 263: Can we add a test case with a deprecated field that has no reason specified?

crates/apollo-mcp-server/src/introspection/tools/introspect.rs
```diff
@@ -122,7 +122,7 @@ fn tool_description(
     minify: bool,
 ) -> String {
     if minify {
-        "Get GraphQL type information - T=type,I=input,E=enum,U=union,F=interface;s=String,i=Int,f=Float,b=Boolean,d=ID;!=required,[]=list,<>=implements;".to_string()
+        "Get GraphQL type information - T=type,I=input,E=enum,U=union,F=interface;s=String,i=Int,f=Float,b=Boolean,d=ID;@D=deprecated;!=required,[]=list,<>=implements;".to_string()
```
DaleSeo, line 125: Just out of curiosity, why `@D` instead of `@d`? I think it would be useful to establish consistent rules for abbreviation since we'll likely need to add more items in the future.

DaleSeo (commented):
We should update the relevant section in our docs:

image

https://www.apollographql.com/docs/apollo-mcp-server/define-tools#minification

crates/apollo-mcp-server/src/introspection/tools/introspect.rs
esilverm: Great question! I was concerned about collisions in understanding because if `d=ID` than would the LLM see `@d` as `@ID` in certain unexpected cases? I could do more testing to see if this does happen but with nondeterministic understanding I dont know if is worth taking that risk.

crates/apollo-mcp-server/src/introspection/minify.rs
esilverm: Thanks for bringing this up, since I had been questioning the need for this handling myself after putting this together. Currently, a `@deprecated` directive with no arguments would be treated as `@D` and this code is likely unreachable unless the `@deprecated` directive spec changed and another argument was added. I'll tweak things to make it clear that we should only include reason if we have the reason argument.

PR #375, opened 2025-09-22
Telemetry: Trace operations and auth

swcollard:
We were missing traces for the MCP server generating Tools from Operations and performing authorization. This PR also add the HTTP status code to the top level HTTP trace

### DIFF
--- a/crates/apollo-mcp-server/src/auth.rs
+++ b/crates/apollo-mcp-server/src/auth.rs
@@ -84,6 +84,7 @@ impl Config {
 }
 
 /// Validate that requests made have a corresponding bearer JWT token
+#[tracing::instrument(skip_all, fields(status_code, reason))]
 async fn oauth_validate(
     State(auth_config): State<Config>,
     token: Option<TypedHeader<Authorization<Bearer>>>,
@@ -104,17 +105,23 @@ async fn oauth_validate(
     };
 
     let validator = NetworkedTokenValidator::new(&auth_config.audiences, &auth_config.servers);
-    let token = token.ok_or_else(unauthorized_error)?;
+    let token = token.ok_or_else(|| {
+        tracing::Span::current().record("reason", "missing_token");
+        tracing::Span::current().record("status_code", StatusCode::UNAUTHORIZED.as_u16());
+        unauthorized_error()
+    })?;
 
-    let valid_token = validator
-        .validate(token.0)
-        .await
-        .ok_or_else(unauthorized_error)?;
+    let valid_token = validator.validate(token.0).await.ok_or_else(|| {
+        tracing::Span::current().record("reason", "invalid_token");
+        tracing::Span::current().record("status_code", StatusCode::UNAUTHORIZED.as_u16());
+        unauthorized_error()
+    })?;
 
     // Insert new context to ensure that handlers only use our enforced token verification
     // for propagation
     request.extensions_mut().insert(valid_token);
 
     let response = next.run(request).await;
+    tracing::Span::current().record("status_code", response.status().as_u16());
     Ok(response)
 }
--- a/crates/apollo-mcp-server/src/operations/operation.rs
+++ b/crates/apollo-mcp-server/src/operations/operation.rs
@@ -48,6 +48,7 @@ impl Operation {
         self.inner
     }
 
+    #[tracing::instrument(skip(graphql_schema, custom_scalar_map))]
     pub fn from_document(
         raw_operation: RawOperation,
         graphql_schema: &GraphqlSchema,
@@ -138,6 +139,7 @@ impl Operation {
     }
 
     /// Generate a description for an operation based on documentation in the schema
+    #[tracing::instrument(skip(comments, tree_shaker, graphql_schema))]
     fn tool_description(
         comments: Option<String>,
         tree_shaker: &mut SchemaTreeShaker,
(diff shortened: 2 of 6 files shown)


crates/apollo-mcp-server/src/auth.rs
```diff
@@ -104,17 +105,85 @@ async fn oauth_validate(
...
+    use super::*;
+    use axum::middleware::from_fn_with_state;
+    use axum::routing::get;
+    use axum::{
+        Router,
+        body::Body,
+        http::{Request, StatusCode},
+    };
+    use http::header::{AUTHORIZATION, WWW_AUTHENTICATE};
+    use tower::ServiceExt; // for .oneshot()
```
DaleSeo, line 140: Cool! 👍

crates/apollo-mcp-server/src/auth.rs
```diff
@@ -104,17 +105,85 @@ async fn oauth_validate(
...
+            .header(AUTHORIZATION, "Bearer invalidtoken")
+            .body(Body::empty())
+            .unwrap();
+        let res = app.oneshot(req).await.unwrap();
+        assert_eq!(res.status(), StatusCode::UNAUTHORIZED);
+        let headers = res.headers();
+        let www_auth = headers.get(WWW_AUTHENTICATE).unwrap().to_str().unwrap();
+        assert!(www_auth.contains("Bearer"));
+        assert!(www_auth.contains("resource_metadata"));
+    }
```
DaleSeo, line 188: Thanks for adding these tests! Would you be able to add a test for the successful case as well?

crates/apollo-mcp-server/src/server/states/starting.rs
```diff
@@ -193,15 +193,27 @@ impl Starting {
...
-                            },
-                        ));
+                        .layer(
+                            TraceLayer::new_for_http()
+                                .make_span_with(|request: &axum::http::Request<_>| {
+                                    tracing::info_span!(
+                                        "mcp_server",
+                                        method = %request.method(),
+                                        uri = %request.uri(),
+                                        status = tracing::field::Empty,
```
DaleSeo, line 203: I guess this is needed as a placeholder until the response is available?

crates/apollo-mcp-server/src/server/states/starting.rs
```diff
@@ -297,3 +309,49 @@ async fn health_endpoint(
...
+mod tests {
+    use http::HeaderMap;
+    use url::Url;
+
+    use crate::health::HealthCheckConfig;
+
+    use super::*;
+
+    #[tokio::test]
+    async fn start_basic_server() {
```
DaleSeo, line 323: 👍

crates/apollo-mcp-server/src/operations/operation.rs
```diff
@@ -48,6 +48,7 @@ impl Operation {
         self.inner
     }
 
+    #[tracing::instrument(skip(graphql_schema, custom_scalar_map))]
```
DaleSeo, line 51: Do we skip them because they can be too large, or is it to avoid infinite recursion?

crates/apollo-mcp-server/src/operations/operation_source.rs
```diff
@@ -38,6 +38,7 @@ pub enum OperationSource {
 }
 
 impl OperationSource {
+    #[tracing::instrument]
```
DaleSeo: Will this include `self`?

crates/apollo-mcp-server/src/server/states/starting.rs
swcollard: Correct. All the fields need to be defined when the Span is created, even if the value is recorded later.

https://docs.rs/tracing/latest/tracing/attr.instrument.html#adding-fields

crates/apollo-mcp-server/src/auth.rs
swcollard: I would need to refactor the class some to do that since this line that creates the validator assumes we have a remote validation server accessible over the network

```rust
let validator = NetworkedTokenValidator::new(&auth_config.audiences, &auth_config.servers);
```

crates/apollo-mcp-server/src/operations/operation.rs
swcollard: In this case, because they are too large. We wouldn't want to log the entire GraphQL schema on each trace of generating an operation. To me, the more interesting data is the contents of the operation document.

crates/apollo-mcp-server/src/server/states/starting.rs
DaleSeo: Thanks for the link, @swcollard! That makes sense.

crates/apollo-mcp-server/src/auth.rs
DaleSeo: If refactoring for the test is too much work right now, I'm fine with postponing it. We can come back to it and add the test when it’s more convenient. 🙂

crates/apollo-mcp-server/src/operations/operation_source.rs
swcollard: Yes, so it will log either the files like in this example.

Screenshot 2025-09-23 at 9 05 07 AM

Or it will log whether the operation is generated from persisted queries or operation collections.

`self` is not the best name though.. I am going to rename it

PR #376, opened 2025-09-23
Add section on SDL docstrings for tool definitions

jcaudle:
Updated best practices and tool definition docs to reference and explain the use of SDL docstrings for improving tool descriptions. Added guidance and an example to help users document their GraphQL operations for better LLM integration.

### DIFF
--- a/docs/source/define-tools.mdx
+++ b/docs/source/define-tools.mdx
@@ -234,3 +234,18 @@ introspection:
 ```sh title="Example command using introspection"
 apollo-mcp-server <path to the preceding config>
 ```
+
+## Adding SDL Docstrings
+
+Apollo MCP Server also supports adding [SDL docstrings](apollo-server/schema/schema#descriptions-docstrings) to tool defintions. This form of documentation can make it easier for LLMs to determine which tool is best for any given prompt. Apollo recommends adding docstrings to any tool you define so you can be sure your MCP tools are used appropriately.
+
+Docstrings are a great way to provide more specific information about what your operations do than the default description provided by your GraphQL operation. Use this feature to describe the options available from the tool or explain why the operation is useful and your MCP client will be able to make better choices of which tool to use.
+
+```graphql
+# Get the detailed weather forecast for a given location
+query GetForecast($coordinate: InputCoordinate!) {
+  forecast(coordinate: $coordinate) {
+    detailed
+  }
+}
+```
--- a/docs/source/best-practices.mdx
+++ b/docs/source/best-practices.mdx
@@ -13,7 +13,7 @@ This feature is in [preview](/graphos/resources/feature-launch-stages#preview). 
 
 The schema is required for:
 
-- **Tool Descriptions**: The schema provides type information used to generate tool descriptions. You can override these descriptions by adding comments to your operation files.
+- **Tool Descriptions**: The schema provides type information used to generate tool descriptions. You can override these descriptions by [adding comments](/apollo-mcp-server/define-tools#adding-sdl-docstrings) to your operation files.
 - **Input Validation**: The schema is used to translate GraphQL input types into JSON Schema, ensuring that AI models provide correctly formatted inputs.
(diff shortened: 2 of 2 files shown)


docs/source/define-tools.mdx
```diff
@@ -233,4 +233,19 @@ introspection:
 
 ```sh title="Example command using introspection"
 apollo-mcp-server <path to the preceding config>
-```
\ No newline at end of file
+```
+
+## Adding SDL Docstrings
+
+Apollo MCP Server also supports adding [SDL docstrings](apollo-server/schema/schema#descriptions-docstrings) to tool defintions. This form of documentation can make it easier for LLMs to determine which tool is best for any given prompt. Apollo recommends adding docstrings to any tool you define so you can be sure your MCP tools are used appropriately.
```
DaleSeo, line 240: > Apollo recommends adding docstrings to any tool you define so you can be sure your MCP tools are used appropriately.

I'm not sure this should be our primary guidance. In my view, it's more effective to encourage deverlops to write clear, descriptive schema documentation from the start. If the schema itself is well-documented, those descriptions can be leveraged across all MCP server instances, making additional docstrings less necessary. This approach also reinforces best practices in GraphQL, where thorough schema descriptions have always been the standard. I'd prefer we nudge users toward improving their schema documentation rather than relying on docstrings as a workaround.

docs/source/define-tools.mdx
```diff
@@ -233,4 +233,19 @@ introspection:
...
+
+Docstrings are a great way to provide more specific information about what your operations do than the default description provided by your GraphQL operation. Use this feature to describe the options available from the tool or explain why the operation is useful and your MCP client will be able to make better choices of which tool to use.
+
+```graphql
+# Get the detailed weather forecast for a given location
+query GetForecast($coordinate: InputCoordinate!) {
+  forecast(coordinate: $coordinate) {
+    detailed
+  }
+}
```
DaleSeo, line 250: It would be great if the example operation shows that docstrings can also be added for input arguments.

```suggestion
# Get the detailed weather forecast for a given location
query GetForecast(
  # latitude and longitude
  $coordinate: InputCoordinate!
) {
  forecast(coordinate: $coordinate) {
    detailed
  }
}
```

mabuyo: Watson shared this code snippet as an example.

```
# Get the schema change proposals related to a given graph id
# If the count of proposals returned is less than the totalCount, use offset to paginate through the rest of proposals
query GetSchemaChangeProposals(
  $graphId: ID!
  # Valid values: "APPROVED", "CLOSED", "DRAFT", "IMPLEMENTED", "OPEN".
  # Default to ["OPEN"] if not provided.
  $filterByStatus: [ProposalStatus!]
  # Filter proposals to only affected subgraph names
  $filterBySubgraph: [String!]
  # Filter proposals to specific source variants
  $filterBySourceVariant: [String!]
  # Use for pagination
  $offset: Int
  # Defaults to 5, don't exceed 25
  $limit: Int = 5
) {
  graph(id: $graphId) {
    id
    name
    proposals(
      filterBy: { 
        status: $filterByStatus
        subgraphs: $filterBySubgraph
        sourceVariants: $filterBySourceVariant 
      }
      offset: $offset
      limit: $limit
    ) {
      totalCount
      proposals {
        id
        createdAt
        activities {
          nodes {
            activity
            createdAt
          }
        }
        status
        displayName
      }
    }
  }
}
```

mabuyo: When we add these to an _operation_ / tool though, should it be referred to as docstrings? Technically they are comments

alocay: I agree with @mabuyo , these are operation comments. We should also show an example of the tool descriptions being added as schema file docstrings:
```
"""
my description
"""
type Foo { ... }
```
To me, this is the preferred method and operation comments are mainly for specific overrides.

alocay:
Should we add a note that mentioning that overrides via comments in the operation is not really supported when using manifest files? It can technically be done (I think) but it would be tedious and be high maintenance.

PR #387, opened 2025-09-26
deps: update rust to v1.90.0

DaleSeo:
Just updating the Rust version to match Rover.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,7 @@ authors = ["Apollo <opensource@apollographql.com>"]
 edition = "2024"
 license-file = "LICENSE"
 repository = "https://github.com/apollographql/apollo-mcp-server"
-rust-version = "1.90.0"
+rust-version = "1.89.0"
 version = "0.9.0"
 
 [workspace.dependencies]


Cargo.toml
```diff
@@ -11,7 +11,7 @@ authors = ["Apollo <opensource@apollographql.com>"]
 edition = "2024"
 license-file = "LICENSE"
 repository = "https://github.com/apollographql/apollo-mcp-server"
-rust-version = "1.89.0"
+rust-version = "1.90.0"
```
nicholascioli: Nit: So this is technically the minimally supported rust version (MSRV) [more details [here](https://doc.rust-lang.org/cargo/reference/rust-version.html)]. We had previously pinned this to `1.89.0` since we were using some of its newer features (let chaining in conditionals).

We can bump this as we bump our in-use version of rust, but let me know if you prefer otherwise.

nicholascioli (approved):
LGTM! I left a small nit.

Cargo.toml
DaleSeo: I did not know this field indicates the minimum supported Rust version. I'll update it back to 1.89 since our code doesn't depend on any 1.90-only features.

PR #390, opened 2025-09-29
chore: forwarding headers

sasansamit:


crates/apollo-mcp-server/src/server/states/running.rs
orgito, line 1: Will this forward the authorization header from the mcp client to the graphql server?

swcollard:
Closing this pr. We prefer to add configuration to allow certain headers to be passed through rather than hardcoding them. 
see: https://github.com/apollographql/apollo-mcp-server/pull/428

sasansamit:
@swcollard the PR was more for showcasing the changes i had made for a POC i was doing and sharing it with the apollo mcp server team ... yes the ideal way would be to have a configuration defined set of headers to forward thru, thats the way i envision it but my understanding is that the team is already looking in this direction so didn't clean it up.

PR #405, opened 2025-10-01
Add rc version to install script before tag

alocay:
Adding a step replace install script versions with RC build (if set), commit, push, and then tag. Afterwards it will update the install scripts again with the new release version.

### DIFF
--- a/.github/workflows/prep-release.yml
+++ b/.github/workflows/prep-release.yml
@@ -91,7 +91,9 @@ jobs:
             esac
             
             NEW_VERSION="$MA.$MI.$PA"
+            RC_VERSION="$NEW_VERSION-rc.1"
             echo "new_version=$NEW_VERSION" >> "$GITHUB_OUTPUT"
+            echo "rc_version=$RC_VERSION"
             echo "Bumped: $CURR -> $NEW_VERSION"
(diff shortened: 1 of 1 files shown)


.github/workflows/prep-release.yml
```diff
@@ -161,12 +164,81 @@ jobs:
...
+          git tag -a "v$tag" -m "$tag"
+          git push origin tag "v$tag"
+
+      - name: Replace versions in scripts and docs with release version
+        env:
+          # If rc_version is non-empty, use it; otherwise use current_version
+          CURR_VERSION: ${{ steps.meta.outputs.rc_version != '' && steps.meta.outputs.rc_version || steps.meta.outputs.current_version }}
+          NEW_VERSION: ${{ steps.bump.outputs.new_version }}
+        run: |
+          python3 - <<'PY'
```
swcollard, line 195: This seems copy/pasted from the other step where the RC is generated. It'd be nice if it could be reused and we just pass either the RC version or the new version into it but its probably fine if we want to refactor that later

alocay: Yeah I was considering this as well. Let me see if there a way to do this and update.

Issue #411, opened 2025-10-02
Support for sending GRPC headers to authenticate MCP server with remote OTEL server


swcollard:
> does the OTEL PR support sending GRPC headers to authenticate MCP server with the OTEL server.  This is a requirement for NewRelic OTEL servers.
> 
> Here is an example from the Apollo Router config.yaml for OTEL headers plus tagging to help with the context of my question.  This would is fairly generic for a NewRelic Operator running on Kubernetes with Apollo MCP deployment.
> 
> ```
> telemetry:
>   instrumentation:
>     spans:
>       mode: spec_compliant
>   exporters:
>     tracing:
>       common:
>         parent_based_sampler: true
>         # (Optional) Set the service name to easily find metrics related to the apollo-router in your metrics dashboards
>         service_name: "${env.NEW_RELIC_APP_NAME}"
>         resource:
>           environment.cluster_name: "${env.NEW_RELIC_METADATA_KUBERNETES_CLUSTER_NAME}"
>           environment.container_image_name: "${env.NEW_RELIC_METADATA_KUBERNETES_CONTAINER_IMAGE_NAME}"
>           environment.container_name: "${env.NEW_RELIC_METADATA_KUBERNETES_CONTAINER_NAME}"
>           environment.namespace: "${env.NEW_RELIC_METADATA_KUBERNETES_NAMESPACE_NAME}"
>           environment.node_name: "${env.NEW_RELIC_METADATA_KUBERNETES_NODE_NAME}"
>           environment.pod_name: "${env.NEW_RELIC_METADATA_KUBERNETES_POD_NAME}"
>           # (Optional) Set the service name to easily find metrics related to the apollo-router in your metrics dashboards
>           service.name: "${env.NEW_RELIC_APP_NAME}"
>       otlp:
>         enabled: true
>         # Temporality MUST be set to delta. Failure to do this will result in incorrect metrics.
>         temporality: delta
>         # Endpoint for your region.
>         endpoint: https://otlp.nr-data.net
>         protocol: grpc
>         grpc:
>           metadata:
>             "api-key":
>               - "${env.NEW_RELIC_API_KEY}"
>       propagation:
>         trace_context: true
>     metrics:
>       common:
>         # (Optional) Set the service name to easily find metrics related to the apollo-router in your metrics dashboards
>         service_name: "${env.NEW_RELIC_APP_NAME}"
>         resource:
[...]

PR #413, opened 2025-10-02
feat: Add configuration option for metric temporality

swcollard:
re: https://github.com/apollographql/apollo-mcp-server/issues/411

One of the other features mentioned in this issue is setting the metric Temporality configuration option on the SDK. This can have implications as different vendors requires the use of Delta over the default of Cumulative. This change creates a new enum to expose that option in the configuration file and sets the Temporality on the otel sdk.

Spec: https://opentelemetry.io/docs/specs/otel/metrics/data-model/#temporality
Blogs explaining differences: https://grafana.com/blog/2023/09/26/opentelemetry-metrics-a-guide-to-delta-vs.-cumulative-temporality-trade-offs/ & https://docs.datadoghq.com/opentelemetry/guide/otlp_delta_temporality/

Ports similar change made in Apollo Router https://github.com/apollographql/router/pull/3412

### DIFF
--- a/docs/source/config-file.mdx
+++ b/docs/source/config-file.mdx
@@ -254,7 +254,7 @@ transport:
 
 | Option        | Type                                     | Default                  | Description                                                                                                                                                                                             |
 | :---------    | :--------------------------------------- | :----------------------- | :------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
-| `endpoint`    | `URL`                                    | `http://localhost:4137`  | URL to export data to. Requires full path.                                                                                                                                                              |
+| `endpoint`    | `URL`                                    | `http://localhost:4317`  | URL to export data to. Requires full path.                                                                                                                                                              |
 | `protocol`    | `string`                                 | `grpc`                   | Protocol for export. Only `grpc` and `http/protobuf` are supported.                                                                                                                                     |
 | `temporality` | `MetricTemporality` | `Cumulative`             | Optional OTel property that refers to the way additive quantities are expressed, in relation to time, indicating whether reported values incorporate previous measurements (Cumulative) or not (Delta). |
 


crates/apollo-mcp-server/src/runtime/telemetry.rs
```diff
@@ -44,17 +44,34 @@ pub struct MetricsExporters {
...
+            MetricTemporality::Delta => Temporality::Delta,
+        }
+    }
+}
+
 #[derive(Debug, Deserialize, JsonSchema)]
 pub struct OTLPMetricExporter {
     endpoint: String,
     protocol: String,
+    temporality: MetricTemporality,
```
DaleSeo: Shouldn't we make it optional for backward compatibility?

```suggestion
    #[serde(default)]
    temporality: Option<MetricTemporality>,
```

docs/source/config-file.mdx
```diff
@@ -244,20 +244,37 @@ transport:
...
+| `protocol`    | `string`      | `grpc`                      | Protocol for export. Only `grpc` and `http/protobuf` are supported.                                                                                                       |
+| `temporality` | `Temporality` | `Cumulative`                | OTel property that refers to the way additive quantities are expressed, in relation to time, indicating whether reported values incorporate previous measurements or not. |
+
+##### Temporality
+
+| Option       | Type    | Description                                                                                       |
+| :----------- | :------ | :------------------------------------------------------------------------------------------------ |
+| `Cumulative` | `enum`  | The metric value will be the overall value since the start of the measurement. (default behavior) |
+| `Delta`      | `enum`  | The metric will be the difference in the measurement since the last time it was reported.         |
```
DaleSeo: Could we embed this into the table above? Navigating multiple nested config references will be hard for readers. 😭

crates/apollo-mcp-server/src/runtime/telemetry.rs
alocay: We should definitely have a default

crates/apollo-mcp-server/src/runtime/telemetry.rs
```diff
@@ -44,17 +44,34 @@ pub struct MetricsExporters {
...
 }
 
+#[derive(Debug, Deserialize, JsonSchema)]
+pub enum MetricTemporality {
+    Cumulative,
+    Delta,
+}
+
+impl MetricTemporality {
+    pub fn to_sdk(&self) -> Temporality {
```
alocay: nit: is `to_sdk` a convention? If not, what about `to_temporality` or something to be more clear?

swcollard: Will update

crates/apollo-mcp-server/src/runtime/telemetry.rs
swcollard: Will make optional

docs/source/config-file.mdx
swcollard: Will try, I don't think I can nest the table in markdown but I can include the two options.

docs/source/config-file.mdx
```diff
@@ -244,20 +244,29 @@ transport:
...
+| :---------    | :--------------------------------------- | :----------------------- | :------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| `endpoint`    | `URL`                                    | `http://localhost:4137`  | URL to export data to. Requires full path.                                                                                                                                                              |
+| `protocol`    | `string`                                 | `grpc`                   | Protocol for export. Only `grpc` and `http/protobuf` are supported.                                                                                                                                     |
+| `temporality` | `Optional oneOf ["Cumulative", "Delta"]` | `Cumulative`             | Optional OTel property that refers to the way additive quantities are expressed, in relation to time, indicating whether reported values incorporate previous measurements (Cumulative) or not (Delta). |
```
DaleSeo: nit: Looking at how existing enums such as ‎`OperationSource` are documented, we should just mention the enum name. The possible values are already clarified in the description, so that'll be sufficient.

```suggestion
| `temporality` | `MetricTemporality` | `Cumulative`             | Optional OTel property that refers to the way additive quantities are expressed, in relation to time, indicating whether reported values incorporate previous measurements (Cumulative) or not (Delta). |
```

PR #417, opened 2025-10-03
Air 42

DaleSeo:


.changesets/breaking_defalut_port_8000.md
```diff
@@ -0,0 +1,20 @@
+### Change default port from 5000 to 8000 - @DaleSeo PR #417
```
swcollard, line 1: nit, from the README the filename requires a prefix. 

> 3. Files must start with a prefix that indicates the classification of the changeset.

https://github.com/apollographql/apollo-mcp-server/blob/main/.changesets/README.md

Also do we consider this a breaking change?

docs/source/guides/auth-auth0.mdx
```diff
@@ -255,8 +255,8 @@ Then, continue with the following steps:
 
 #### MCP Server Won't Start
 
-- **Error**: "Port 5000 is already in use"
-  - Solution: Kill any existing processes using port 5000 or specify a different port with the `transport.port` option or `APOLLO_MCP_TRANSPORT__PORT` env variable
+- **Error**: "Port 8000 is already in use"
```
swcollard, line 258: Could we just remove this section? I assume this troubleshooting step was only added because of the port use region that incentivized us to change the default.

graphql/TheSpaceDevs/README.md
```diff
@@ -69,16 +69,16 @@ There are operations located at `./operations/*.graphql` for you to use in your
...
   --name apollo-mcp-server \
-  -p 5000:5000 \
+  -p 8000:8000 \
   -v $PWD/graphql/TheSpaceDevs:/data \
   ghcr.io/apollographql/apollo-mcp-server:latest \
-  --http-port 5000 \
+  --http-port 8000 \
   --schema api.graphql \
   --operations operations \
   --endpoint https://thespacedevs-production.up.railway.app/
```
swcollard: This docker run command is using CLI arguments instead of the config file which no longer work in the MCP server

graphql/TheSpaceDevs/README.md
```diff
@@ -69,16 +69,16 @@ There are operations located at `./operations/*.graphql` for you to use in your
...
 ```
 
-2. Add the MCP SSE port to your MCP Server configuration for the client appliction you are running. If you are running locally, the server link will be `http://127.0.0.1:5000/sse`.
+2. Add the MCP SSE port to your MCP Server configuration for the client appliction you are running. If you are running locally, the server link will be `http://127.0.0.1:8000/sse`.
```
swcollard: SSE is also deprecated :(

.changesets/breaking_defalut_port_8000.md
DaleSeo: Good catch, @swcollard! Yeah, this should be a breaking change. Will update the file name.

graphql/TheSpaceDevs/README.md
DaleSeo: OMG, this README.md is really outdated! 🙈

docs/source/guides/auth-auth0.mdx
DaleSeo: This is probably still relevant. I often forgot to kill the server before starting a new one. 😂

graphql/TheSpaceDevs/README.md
```diff
@@ -70,24 +64,24 @@ docker run \
   -it --rm \
   --name apollo-mcp-server \
   -p 8000:8000 \
+  -v $PWD/graphql/TheSpaceDevs/config.yaml \
```
swcollard:
```suggestion
  -v $PWD/graphql/TheSpaceDevs/config.yaml:/config.yaml \
```

I forget, does this need the path on the other side to tell docker where to mount the file?

graphql/TheSpaceDevs/README.md
```diff
@@ -38,28 +38,22 @@ Here is an example configuration you can use _(Note: you must provide your fill
     "thespacedevs": {
       "command": "/Users/michaelwatson/Documents/GitHub/apollographql/apollo-mcp-server/target/debug/apollo-mcp-server",
```
swcollard, line 39: I also didn't realize this path was hardcoded to `/Users/michaelwatson`

graphql/TheSpaceDevs/README.md
DaleSeo: You're right and I copied that from our deploy docs! 😂 Will fix it in both places.

https://github.com/apollographql/apollo-mcp-server/blob/93e607b4abc7f1c2e6f453e934df7946d8fd92ff/docs/source/deploy.mdx?plain=1#L75

graphql/TheSpaceDevs/README.md
DaleSeo: It really shows how long ago this README.md was written. We should probably delete it and just use the official docs instead.

PR #421, opened 2025-10-06
feat: Add mcp-session-id header to HTTP request trace attributes

swcollard:


.changesets/feat_trace_session_id.md
```diff
@@ -0,0 +1,3 @@
+### feat: Add mcp-session-id header to HTTP request trace attributes
```
DaleSeo: The PR number and author is missing at the end.

crates/apollo-mcp-server/src/server/states/starting.rs
```diff
@@ -222,6 +222,7 @@ impl Starting {
                                 "mcp_server",
                                 method = %request.method(),
                                 uri = %request.uri(),
+                                session_id = %request.headers().get("mcp-session-id").and_then(|v| v.to_str().ok()).unwrap_or("none"),
```
DaleSeo: Have you considered setting it conditionally? Omitting it would be clearer than setting it to `none`.

DaleSeo (commented):
Thanks, @swcollard, for taking care of this nice-to-have that we discussed!

crates/apollo-mcp-server/telemetry.toml
```diff
@@ -7,6 +7,7 @@ success = "Sucess flag indicator"
 raw_operation = "Graphql operation text and metadata used for Tool generation"
 client_name = "The client name that initializes with the MCP Server"
 client_version = "The client version that initializes with the MCP Server"
+session_id = "The value of the Mcp-Session-Id header in the HTTP request"
```
DaleSeo: Because the mcp-session-id header is present on both request and response, can we add trace attributes for both, so we can correlate end-to-end?

crates/apollo-mcp-server/src/server/states/starting.rs
swcollard: I'm not sure if that is possible when setting the attribute as an input to this macro? Will have to play around with it more

crates/apollo-mcp-server/src/server/states/starting.rs
```diff
@@ -222,14 +222,25 @@ impl Starting {
...
+                                    "status_code",
+                                    tracing::field::display(response.status()),
+                                );
+                                if let Some(session_id) = response
+                                    .headers()
+                                    .get("mcp-session-id")
+                                    .and_then(|v| v.to_str().ok())
+                                {
+                                    span.record("session_id", tracing::field::display(session_id));
+                                }
```
DaleSeo, line 243: You made it! 😄

PR #428, opened 2025-10-07
Add support for forwarding headers from MCP clients to GraphQL APIs

### LINKED ISSUE #198: How to Pass the headers Dynamically to the Apollo MCP Server
I have different AI Applications forwarding the traffic to the MCP Servers whuch having diffrent Client Names how to pass the headers to mcp server dynamically and forward to the supergraph.

eg header is ai-client-name: chat-app

DaleSeo:
Fixes #198

Adds opt-in support for dynamic header forwarding, which enables metadata for A/B testing, feature flagging, geo information from CDNs, or internal instrumentation to be sent from MCP clients to downstream GraphQL APIs. It automatically blocks hop-by-hop headers according to the guidelines in [RFC 7230, section 6.1](https://datatracker.ietf.org/doc/html/rfc7230#section-6.1), and it only works with the Streamable HTTP transport.

You can configure using the `forward_headers` setting:

```yaml
forward_headers:
  - x-tenant-id
  - x-experiment-id
  - x-geo-country
```

Please note that this feature is not intended for passing through credentials as documented in the best practices page.

## E2E Test

Configure the MCP server to target the Platform API and enable CORS as below, then verify end‑to‑end with MCP Inspector:

```yaml
endpoint: https://api.apollographql.com/api/graphql
cors:
  enabled: true
  match_origins:
    - "^http://localhost:[0-9]+$"
  allow_headers:
    - accept
    - content-type
    - mcp-protocol-version
    - mcp-session-id
    - traceparent
    - tracestate
    - x-custom-auth-headers
    - X-API-Key
    - apollographql-client-name
    - apollographql-client-version
forward_headers:
  - X-API-Key
  - apollographql-client-name
  - apollographql-client-version
transport:
  type: streamable_http
  port: 8000
```

With the 3 required headers set in MCP Inspector, tool execution succeeds:

2025-10-08 at 17 06 00

If you configure MCP Inspector with Connection Type = Direct, you can also inspect the MCP call in the browser:

2025-10-08 at 17 20 27

If any required header is missing, the server returns an error:

2025-10-08 at 17 18 18

### DIFF
--- a/.changesets/feat_forward_headers.md
+++ b/.changesets/feat_forward_headers.md
@@ -1,0 +1,14 @@
+### Add support for forwarding headers from MCP clients to GraphQL APIs - @DaleSeo PR #428
+
+Adds opt-in support for dynamic header forwarding, which enables metadata for A/B testing, feature flagging, geo information from CDNs, or internal instrumentation to be sent from MCP clients to downstream GraphQL APIs. It automatically blocks hop-by-hop headers according to the guidelines in [RFC 7230, section 6.1](https://datatracker.ietf.org/doc/html/rfc7230#section-6.1), and it only works with the Streamable HTTP transport.
+
+You can configure using the `forward_headers` setting:
+
+```yaml
+forward_headers:
+  - x-tenant-id
+  - x-experiment-id
+  - x-geo-country
+```
+
+Please note that this feature is not intended for passing through credentials as documented in the best practices page.


crates/apollo-mcp-server/src/headers.rs
```diff
@@ -0,0 +1,296 @@
...
+use std::str::FromStr;
+
+use headers::HeaderMapExt;
+use http::Extensions;
+use reqwest::header::{HeaderMap, HeaderName};
+
+use crate::auth::ValidToken;
+
+/// List of header names to forward from MCP clients to GraphQL API
+pub type ForwardHeaders = Vec<String>;
```
swcollard, line 11: Do we want any additional validation on the strings being used here? Like size limits, or excludes `mcp-session-id`, `authorization`, `traceparent`, and `tracestate` which we tend to handle more as special cases?

DaleSeo: Great question, @swcollard ! Which of them do you think we should block from forwarding? I guess forwarding `mcp-session-id`, `traceparent` and `tracestate` to downstream APIs could be helpful from the observability perspective? 🤔  We recommend not using this feature for passing through `authorization` in the best practices docs, but I think we should still leave it to the user's discretion based on customer feedback.

swcollard: > We recommend not using this feature for passing through authorization in the best practices docs, but I think we should still leave it to the user's discretion based on customer feedback.

This feels sufficient to me. As long as its documented as not a best practice.

mcp-session-id, traceparent and tracestate we are already passing downstream so they don't need to be defined in this list.

From doing some searching, http doesn't have any official limits on how large and how many headers there can be so maybe it is fine to leave this as an unbounded list.

damassi:
This feature is great, though the docs are a bit unclear about needing to ungate via `cors` in order for this to work.

rrazvd:
@DaleSeo the LLM can't provide a value for a specific header through the tool? like x-tenant-id ?

PR #433, opened 2025-10-08
chore: Rmcp sdk update 0.8.x

swcollard:
Checking the changelog we are now 2 minor versions behind the SDK. There were no breaking changes so this seemed like an easier update. There were several changes around oauth we should validate before release

https://github.com/modelcontextprotocol/rust-sdk/releases


DaleSeo (approved):
Thanks for the upgrade! Glad it has no breaking changes this time. 😄

PR #436, opened 2025-10-08
docs: add container destination path for config file

DaleSeo:
The Docker run example is missing the container destination path for the config file.

### DIFF
--- a/docs/source/deploy.mdx
+++ b/docs/source/deploy.mdx
@@ -71,7 +71,7 @@ schema:
 docker run \
   -it --rm \
   --name apollo-mcp-server \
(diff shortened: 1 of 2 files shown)


docs/source/deploy.mdx
```diff
@@ -72,7 +72,7 @@ docker run \
   -it --rm \
   --name apollo-mcp-server \
   -p 5050:5050 \
```
swcollard: I've lost track of all of our conversations about this, but is 5050:5050 correct, or should this be 5050:5000?

DaleSeo: Good catch! 🤦‍♂️

PR #460, opened 2025-10-20
feat: adding config option to specify metadata values for telemetry exporters

alocay:
Adding a `metadata` (for `grpc` protocol) and `headers` (for `http/protobuf` protocol) config option for the telemetry yaml config section.

### DIFF
--- a/crates/apollo-mcp-server/src/runtime/telemetry.rs
+++ b/crates/apollo-mcp-server/src/runtime/telemetry.rs
@@ -596,7 +596,7 @@ mod tests {
                 ("some-key".to_string(), "some-value".to_string()),
                 ("another-key".to_string(), "another-value".to_string()),
             ]
-                .into_iter(),
+            .into_iter(),
         );
 
         let map: MetadataMap = parsers::metadata_map_from_str(de).unwrap();
@@ -626,9 +626,12 @@ mod tests {
 
         let telemetry_config: Telemetry = serde_yaml::from_str(y).unwrap();
         let metric_exporter = telemetry_config
-            .exporters.unwrap()
-            .metrics.unwrap()
-            .otlp.unwrap();
+            .exporters
+            .unwrap()
+            .metrics
+            .unwrap()
+            .otlp
+            .unwrap();
 
         match metric_exporter {
             MetricTelemetryExporter::Grpc { metadata, .. } => {
@@ -642,7 +645,9 @@ mod tests {
                     Some(&MetadataValue::try_from("another-value").unwrap())
                 );
             }
-            MetricTelemetryExporter::HttpProtobuf { .. } => { panic!("unexpected protocol") }
+            MetricTelemetryExporter::HttpProtobuf { .. } => {
+                panic!("unexpected protocol")
+            }
         }
     }
 }


crates/apollo-mcp-server/src/runtime/telemetry.rs
```diff
@@ -346,84 +419,83 @@ mod tests {
...
             }),
-            None,
         );
-        let result = init_meter_provider(&config.telemetry);
+        let result = init_tracer_provider(&config.telemetry);
         assert!(result.is_ok());
     }
 
     #[tokio::test]
-    async fn unknown_protocol_raises_tracer_provider_error() {
```
alocay, line 392: No longer needed since protocol is parsed to an enum value and cannot be an unknown string value as it was in the previous `OTLPTracingExporter` struct. The error will occur immediately when the yaml is parsed instead of here when the value is used.

docs/source/config-file.mdx
```diff
@@ -282,10 +284,20 @@ transport:
...
-| :---------  | :-------- | :-------------------------- | :--------------------------------------------------------------- |
-| `endpoint`  | `URL`     | `http://localhost:4317`     | URL to export data to. Requires full path.                       |
-| `protocol`  | `string`  | `grpc`                      | Protocol for export. `grpc` and `http/protobuf` are supported.   |
+| Option       | Type                                         | Default                 | Description                                                    |
+|:-------------|:---------------------------------------------|:------------------------|:---------------------------------------------------------------|
+| `endpoint`   | `URL`                                        | `http://localhost:4317` | URL to export data to. Requires full path.                     |
+| `protocol`   | `string`                                     | `grpc`                  | Protocol for export. `grpc` and `http/protobuf` are supported. |
+| `metadata`   | `Key/Value Pairs`                            |                         | The metadata key values for GRPC protocol                      |
```
alocay: @mabuyo Is there a better way to indicate that `metadata` is only used with the `protocol` value is `grpc` and `headers` is used if the `protocol` is `http/protobuf`?

alocay: I added an extra example for this below

DaleSeo:
Nix CI seems to be failing because of this error:

2025-10-20 at 11 29 08

docs/source/config-file.mdx
```diff
@@ -313,6 +325,37 @@ operations:
...
+  exporters:
+    metrics:
+      otlp:
+        endpoint: "http://localhost:4317"
+        protocol: "http/protobuf"
+        headers:
+          key-name: some-value
+          key-two: another-value
+```
+
```
mabuyo: Let's put these in an `ExpansionPanel` component (a section that's collapsed by default) and closer to the config it's related to.

docs/source/config-file.mdx
```diff
@@ -282,10 +284,20 @@ transport:
...
-| `endpoint`  | `URL`     | `http://localhost:4317`     | URL to export data to. Requires full path.                       |
-| `protocol`  | `string`  | `grpc`                      | Protocol for export. `grpc` and `http/protobuf` are supported.   |
+| Option       | Type                                         | Default                 | Description                                                    |
+|:-------------|:---------------------------------------------|:------------------------|:---------------------------------------------------------------|
+| `endpoint`   | `URL`                                        | `http://localhost:4317` | URL to export data to. Requires full path.                     |
+| `protocol`   | `string`                                     | `grpc`                  | Protocol for export. `grpc` and `http/protobuf` are supported. |
+| `metadata`   | `Key/Value Pairs`                            |                         | The metadata key values for GRPC protocol                      |
+| `headers`    | `Key/Value Pairs`                            |                         | The header key values for HTTP/Protobuf protocol               |
```
mabuyo:
```suggestion
| `metadata`   | Key-value pairs                            |                         | The metadata key-value pairs. Used only if `protocol` is set to `grpc`.                      |
| `headers`    | Key-value pairs                            |                         | The header key-value pairs. Used only if `protocol` is set to `http/protobuf`.               |
```

Will need to fix spacing with the rest of the table.

Does this feel a bit clearer? I also removed the codefont for the type since it isn't really a type but more of a definition (correct me if I'm wrong!).

crates/apollo-mcp-server/src/runtime/telemetry.rs
```diff
@@ -38,67 +39,132 @@ pub struct Exporters {
...
 
+/// Telemetry exporter options
+#[derive(Debug, Deserialize, JsonSchema)]
+#[serde(tag = "protocol", rename_all = "lowercase")]
+pub enum TelemetryExporter {
+    /// GRPC Exporter
+    Grpc {
+        endpoint: String,
+        #[serde(default)]
+        temporality: MetricTemporality,
```
DaleSeo, line 50: Does this field apply only to metrics exporters and not tracing? If that's the case, when the enum is used for tracing, the `temporality` will be ignored silently. 

What do you think about making this field optional and adding runtime validation based on the exporter type? Another option could be to split it into `MetricExporter` and `TracingExporter`.

nicholascioli: I agree with splitting into metrics vs tracing

crates/apollo-mcp-server/src/runtime/telemetry.rs
```diff
@@ -38,67 +39,132 @@ pub struct Exporters {
...
 
-#[derive(Debug, Deserialize, JsonSchema)]
+#[derive(Debug, Default, JsonSchema, Serialize)]
+#[serde(rename_all = "lowercase")]
 pub enum MetricTemporality {
+    #[default]
     Cumulative,
     Delta,
+    LowMemory,
 }
```
nicholascioli: Why not use serde's remote crate deserialization here? Then you can remove the custom `deserialize` and `From` impl.

graphql/TheSpaceDevs/config.yaml
```diff
@@ -22,3 +22,11 @@ introspection:
...
   enabled: true
   allow_any_origin: true
+#telemetry:
+#  exporters:
+#    metrics:
+#      otlp:
+#        endpoint: "http://127.0.0.1:4317"
+#        protocol: "grpc"
+#        metadata:
+#          api-key: my-super-secret-key
```
nicholascioli: Was this meant to be here but commented out?

crates/apollo-mcp-server/src/runtime/telemetry.rs
alocay: Using that broke the ability to use `serde(default)` it seems. I could not get the two to work together.

graphql/TheSpaceDevs/config.yaml
alocay: I left it as an example in the project, we can remove if we don't like it.

crates/apollo-mcp-server/src/runtime/telemetry.rs
alocay: I separated the two, let me know what you think.

docs/source/config-file.mdx
alocay: @mabuyo I updated the docs, let me know.

docs/source/config-file.mdx
```diff
@@ -266,11 +266,33 @@ transport:
...
+| `protocol`    | `string`            | `grpc`                  | Protocol for export. Only `grpc` and `http/protobuf` are supported.                                                                                                                                     |
+| `temporality` | `MetricTemporality` | `Cumulative`            | Optional OTel property that refers to the way additive quantities are expressed, in relation to time, indicating whether reported values incorporate previous measurements (Cumulative) or not (Delta). |
+| `metadata`    | `Key/Value Pairs`   |                         | The metadata key values for GRPC protocol                                                                                                                                                               |
+| `headers`     | `Key/Value Pairs`   |                         | The header key values for HTTP/Protobuf protocol                                                                                                                                                        |
```
mabuyo: I missed these ones, can we update these as well to remove the codefont and follow the same description as below?

alocay: I believe this is fixed, let me know.

crates/apollo-mcp-server/src/runtime/telemetry.rs
```diff
@@ -345,89 +450,141 @@ mod tests {
...
-            None,
         );
-        let result = init_meter_provider(&config.telemetry);
+        let result = init_tracer_provider(&config.telemetry);
         assert!(result.is_ok());
     }
 
     #[tokio::test]
-    async fn unknown_protocol_raises_tracer_provider_error() {
+    async fn grpc_metric_exporter_with_metadata_returns_valid_tracer_provider() {
```
DaleSeo: It's testing a trace exporter. I'd rename it for clarity.

```suggestion
    async fn grpc_trace_exporter_with_metadata_returns_valid_tracer_provider() {
```

graphql/TheSpaceDevs/config.yaml
```diff
@@ -22,3 +22,11 @@ introspection:
...
   enabled: true
   allow_any_origin: true
+#telemetry:
+#  exporters:
+#    metrics:
+#      otlp:
+#        endpoint: "http://127.0.0.1:4317"
+#        protocol: "grpc"
+#        metadata:
+#          some-key: some-value
```
DaleSeo, line 32: Is this for convenience? like to comment in and out?

alocay: Yeah i left it there since I've been constantly trying to remember the structure of this every time I work on it. I figured it would be helpful in the future as a reference. I can remove if we don't like it.

alocay:
> @alocay Thanks for addressing my feedback but NIX CI is still failing on the PR 🥲
> 
> Image

Not sure why it fails, it passed just last time the checks ran.

alocay:
I think it's fixed now @DaleSeo

graphql/TheSpaceDevs/config.yaml
DaleSeo: I'm okay with this. If it’s for convenience, we could add it to the other config file as well.

PR #463, opened 2025-10-21
Provide log format options

alocay:
Adding configuration options for logging output format. The available options are `full` (default), `compact`, `json`, and `pretty`. 

```
logging:
  level: info
  format: compact
```

### DIFF
--- a/crates/apollo-mcp-server/src/runtime/logging.rs
+++ b/crates/apollo-mcp-server/src/runtime/logging.rs
@@ -8,14 +8,15 @@ mod format_style;
 mod log_rotation_kind;
 mod parsers;
 
-use crate::runtime::logging::defaults::default_format;
 use crate::runtime::logging::format_style::FormatStyle;
 use log_rotation_kind::LogRotationKind;
 use schemars::JsonSchema;
 use serde::Deserialize;
 use std::path::PathBuf;
 use tracing::Level;
+use tracing_appender::non_blocking::WorkerGuard;
 use tracing_appender::rolling::RollingFileAppender;
+use tracing_subscriber::fmt::MakeWriter;
 use tracing_subscriber::fmt::writer::BoxMakeWriter;
 use tracing_subscriber::{EnvFilter, Layer as LayerTrait, Registry};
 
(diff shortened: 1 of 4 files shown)


crates/apollo-mcp-server/src/runtime/logging.rs
```diff
@@ -36,6 +38,9 @@ pub struct Logging {
     /// [default: Hourly]
     #[serde(default = "defaults::default_rotation")]
     pub rotation: LogRotationKind,
+
+    #[serde(default = "defaults::default_format")]
+    pub format: FormatStyle,
```
nicholascioli: `FormatStyle` already implements `Default`, so just use it instead.

```suggestion
    #[serde(default)]
    pub format: FormatStyle,
```

crates/apollo-mcp-server/src/runtime/logging.rs
```diff
@@ -44,17 +49,13 @@ impl Default for Logging {
             level: defaults::log_level(),
             path: None,
             rotation: defaults::default_rotation(),
+            format: default_format(),
```
nicholascioli:
```suggestion
            format: Default::default(),
```

crates/apollo-mcp-server/src/runtime/logging/defaults.rs
```diff
@@ -8,3 +9,7 @@ pub(super) const fn log_level() -> Level {
 pub(super) const fn default_rotation() -> LogRotationKind {
     LogRotationKind::Hourly
 }
+
+pub(super) const fn default_format() -> FormatStyle {
+    FormatStyle::Full
+}
```
nicholascioli: Nit: Not needed since `FormatStyle` implements `Default`

crates/apollo-mcp-server/src/runtime/logging/format_style.rs
```diff
@@ -0,0 +1,106 @@
...
+use serde::{Deserialize, Deserializer};
+
+#[derive(Debug, Default, JsonSchema, Clone, PartialEq, Eq)]
+pub enum FormatStyle {
+    #[default]
+    Full,
+    Compact,
+    Json,
+    Pretty,
+}
```
nicholascioli: Nit: Make sure to add docs here for each field since the JSON schema will expose that info

crates/apollo-mcp-server/src/runtime/logging/format_style.rs
```diff
@@ -0,0 +1,106 @@
...
+            "compact" => Ok(Self::Compact),
+            "json" => Ok(Self::Json),
+            "pretty" => Ok(Self::Pretty),
+            other => Err(serde::de::Error::unknown_variant(
+                other,
+                &["full", "compact", "json", "pretty"],
+            )),
+        }
+    }
+}
```
nicholascioli: No need for the custom deserializer here. Just use `#[derive(Deserialize)]` with `#[serde(rename_all = "snake_case")]` on the actual type.

nicholascioli: Oh, I see that you want it to be case-insensitive. I would recommend just having it be case sensitive so that you don't have to maintain these extra custom derive rules (and corresponding tests).

crates/apollo-mcp-server/src/runtime/logging.rs
```diff
@@ -4,18 +4,20 @@
...
 use log_rotation_kind::LogRotationKind;
 use schemars::JsonSchema;
 use serde::Deserialize;
 use std::path::PathBuf;
 use tracing::Level;
 use tracing_appender::rolling::RollingFileAppender;
-use tracing_subscriber::EnvFilter;
-use tracing_subscriber::fmt::Layer;
 use tracing_subscriber::fmt::writer::BoxMakeWriter;
+use tracing_subscriber::{EnvFilter, Layer as LayerTrait, Registry};
```
nicholascioli, line 21: Nit: This should be split up like the other imports.

crates/apollo-mcp-server/src/runtime/logging.rs
```diff
@@ -44,17 +49,13 @@ impl Default for Logging {
...
 }
 
 type LoggingLayerResult = (
-    Layer<
-        tracing_subscriber::Registry,
-        tracing_subscriber::fmt::format::DefaultFields,
-        tracing_subscriber::fmt::format::Format,
-        BoxMakeWriter,
-    >,
+    Box<dyn tracing_subscriber::layer::Layer<Registry> + Send + Sync>,
```
nicholascioli: `Layer` is imported above already.

```suggestion
    Box<dyn Layer<Registry> + Send + Sync>,
```

alocay: This has to be LayerTrait instead but I'll make that change.

crates/apollo-mcp-server/src/runtime/logging/format_style.rs
alocay: Done

crates/apollo-mcp-server/src/runtime/logging.rs
alocay: Can you clarify why?

crates/apollo-mcp-server/src/runtime/logging.rs
```diff
@@ -44,40 +50,106 @@ impl Default for Logging {
...
+        }
+
+        if let Some(writer) = self.writer {
+            let layer = tracing_subscriber::fmt::layer();
+            let formatted_layer = match logging.format {
+                FormatStyle::Full => layer
+                    .with_writer(writer)
+                    .with_ansi(self.ansi_enabled)
+                    .with_target(false)
+                    .boxed(),
```
alocay, line 118: @nicholascioli is there a cleaner way of doing this? Since full, compact, json, and pretty all return a different type I could not figure out how to return from this match block without erasing the type (calling boxed). Let me know if there is a better way.

crates/apollo-mcp-server/src/runtime/logging.rs
```diff
@@ -44,40 +50,106 @@ impl Default for Logging {
...
-    >,
-    Option<tracing_appender::non_blocking::WorkerGuard>,
+    Box<dyn LayerTrait<Registry> + Send + Sync>,
+    Option<WorkerGuard>,
 );
 
-impl Logging {
-    pub fn env_filter(logging: &Logging) -> Result<EnvFilter, anyhow::Error> {
-        let mut env_filter = EnvFilter::from_default_env().add_directive(logging.level.into());
+pub struct LoggingLayerBuilder {
```
DaleSeo, line 63: We can derive `Default` for LoggingLayerBuilder instead of manually implementing it, which is more idiomatic:

If we add it here,

```suggestion
#[derive(Default)]
pub struct LoggingLayerBuilder {
```

L69 will look like this:

```rust
impl LoggingLayerBuilder {
    pub fn new() -> Self {
        Self::default()
    }
}
```

crates/apollo-mcp-server/src/runtime/logging.rs
```diff
@@ -44,40 +50,106 @@ impl Default for Logging {
...
+        Self {
+            writer: None,
+            worker_guard: None,
+            ansi_enabled: false,
+        }
+    }
+
+    // This primarily to be used by unit tests to allow for dependency injection.
+    // If no writer is provided when build() is called, one will be created using the logging config options.
+    #[allow(dead_code)]
```
DaleSeo, line 80: Adding production code just for tests isn't ideal. How about only compiling these methods during tests?

```suggestion
    #[cfg(test)]
```

PR #469, opened 2025-10-24
docs: headers and header forwarding

mabuyo:
- Add short section on how to use the `headers` config
- Add caution note to header forwarding (moving the info from "best practices" to the relevant config section)

### DIFF
--- a/docs/source/config-file.mdx
+++ b/docs/source/config-file.mdx
@@ -44,19 +44,19 @@ These fields are under the top-level `graphos` key and define your GraphOS graph
 | `apollo_registry_url`     | `URL`    |         | The URL to use for Apollo's registry                                                                            |
 | `apollo_uplink_endpoints` | `URL`    |         | List of uplink URL overrides. You can also provide this with the `APOLLO_UPLINK_ENDPOINTS` environment variable |
 
-### Headers
+### Static headers
 
-The `headers` option allows you to specify a list of static, hard-coded headers and value. These headers are included in all GraphQL requests.
+The `headers` option enables you to specify a list of static, hard-coded headers and values. These are included in all GraphQL requests.
 
 ```yaml title="mcp.yaml"
 headers: { "apollographql-client-name": "my-mcp-server" }
 ```
 
-To forward header values from the client, use the [forward headers](#forward-headers) option.
+To forward dynamic header values from the client, use the [`forward_headers` option](#forwarding-headers) instead.
 
-### Forward headers
+### Forwarding headers
 
-The `forward_headers` option allows you to forward specific headers from incoming MCP client requests to your GraphQL API. 
+The `forward_headers` option allows you to forward specific headers from incoming MCP client requests to your GraphQL API.
 
 This is useful for:
 - Multi-tenant applications (forwarding tenant IDs)
@@ -66,6 +66,8 @@ This is useful for:
 - Internal instrumentation (forwarding correlation IDs)
 
 Header names should match exactly but are case-insensitive.
+
+Header values are forwarded as-is, according to what the MCP client provides.
 
 Hop-by-hop headers (like `connection`, `transfer-encoding`) are automatically blocked for security.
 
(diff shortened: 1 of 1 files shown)


docs/source/config-file.mdx
```diff
@@ -44,19 +44,50 @@ These fields are under the top-level `graphos` key and define your GraphOS graph
...
+- Client identification (forwarding client names)
 - Internal instrumentation (forwarding correlation IDs)
 
 Supports exact header names only. Hop-by-hop headers (like `connection`, `transfer-encoding`) are automatically blocked for security.
 
+```yaml title="mcp.yaml"
+forward_headers:
+  - x-geo-country
+  - x-client-name
+```
```
mabuyo, line 79: Couple questions here:
- What are common header names that would be forwarded?
- Does casing matter?

docs/source/config-file.mdx
```diff
@@ -44,19 +44,50 @@ These fields are under the top-level `graphos` key and define your GraphOS graph
 | `apollo_registry_url`     | `URL`    |         | The URL to use for Apollo's registry                                                                            |
 | `apollo_uplink_endpoints` | `URL`    |         | List of uplink URL overrides. You can also provide this with the `APOLLO_UPLINK_ENDPOINTS` environment variable |
 
-### Forward Headers
+### Headers
+
+The `headers` option allows you to specify a list of hard-coded headers and values to include in all GraphQL requests.
+
+```yaml title="mcp.yaml"
+headers: { "apollo-client-name": "my-mcp-server" }
```
mabuyo: I added this section because I was confused on the difference between `forward_headers` and `headers`. Wanted to confirm if this example code snippet is correct!

docs/source/best-practices.mdx
```diff
@@ -41,9 +41,3 @@ To maintain clear trust boundaries, MCP servers must only accept tokens explicit
 Forwarding client tokens downstream is not allowed.
 
 Apollo MCP Server supports OAuth 2.1 authentication that follows best practices and aligns with the MCP authorization model. See our [authorization guide](/apollo-mcp-server/auth) for implementation details.
-
```
mabuyo, line 44: The section before this is "Avoid token passthrough" which feels very much related to header forwarding. Can we remove the section above? Maybe add some of its details to the caution note in the config file?

mabuyo (commented):
Left some comments for engineering to review 🙏🏽

docs/source/config-file.mdx
```diff
@@ -44,19 +44,50 @@ These fields are under the top-level `graphos` key and define your GraphOS graph
 | `apollo_registry_url`     | `URL`    |         | The URL to use for Apollo's registry                                                                            |
 | `apollo_uplink_endpoints` | `URL`    |         | List of uplink URL overrides. You can also provide this with the `APOLLO_UPLINK_ENDPOINTS` environment variable |
 
-### Forward Headers
+### Headers
+
+The `headers` option allows you to specify a list of hard-coded headers and values to include in all GraphQL requests.
```
DaleSeo: Can we clarify that these static headers are attached to every single GraphQL request, so it's clearer this feature is different from header forwarding?

DaleSeo (approved):
Looks good! Just a small suggestion.

docs/source/config-file.mdx
DaleSeo: > What are common header names that would be forwarded?

We have no data on that yet but I can imagine the following:
  - x-tenant-id
  - x-experiment-id
  - x-geo-country

> Does casing matter?

No. HTTP headers are case-insensitive.

docs/source/best-practices.mdx
DaleSeo: I'm fine with that. I think it really depends on whether we want to centralize best practices in one place or not.

docs/source/config-file.mdx
mabuyo: Header forwarding is _not_ attached to every single GraphQL request?

DaleSeo: Hmm, how should I put this? 🤔

The distinction is this: `headers` is applied at the server level and is static in that values are fixed when the server starts. In contrast, `forward_headers` operates at the request level and are dynamic in that values are determined at call time.

Not sure if this makes sense. 🥲

mabuyo: Yes this makes sense and matches my understanding!

PR #471, opened 2025-10-28
Fix fragment field validation in schema tree shaking

DaleSeo:
Schema tree shaking was incorrectly handling GraphQL fragments (both inline fragments and fragment spreads) by attempting to expand them immediately without considering their type conditions. This caused "field not found" errors when operations used fragments on union types or interfaces.

```
2025-10-27T13:21:08.249811Z  INFO Apollo MCP Server v1.1.1 // (c) Apollo Graph, Inc. // Licensed under MIT
2025-10-27T13:21:08.265104Z  INFO load_tool: Tool GetEvents loaded with a character count of 275. Estimated tokens: 68
2025-10-27T13:21:08.265483Z  INFO load_tool: Tool GetSessions loaded with a character count of 521. Estimated tokens: 130
2025-10-27T13:21:08.265865Z  INFO load_tool: Tool getNearbyPlaces loaded with a character count of 3766. Estimated tokens: 941
2025-10-27T13:21:08.266050Z ERROR load_tool: field name not found
2025-10-27T13:21:08.266054Z ERROR load_tool: field year not found
2025-10-27T13:21:08.266055Z ERROR load_tool: field start_date not found
```

This PR fixes the fragment handling to process fragments with their proper type context.

- Added `retain_fragments()` function to handles both inline fragments and fragment spreads by evaluating them against their specific type conditions
- Modified `selection_set_to_fields()` function: to only extracts direct field selections, returning empty vectors for fragments (which are handled separately)
- Updated `retain_type()` function to processes fragments first with correct type context, then collects direct fields

## Testing

Add two unit tests to account for the changes. ~I'll ask the issue reporter to test the fix manually.~ It has been confirmed that the fix worked.

image

### DIFF
--- a/crates/apollo-mcp-server/src/schema_tree_shake.rs
+++ b/crates/apollo-mcp-server/src/schema_tree_shake.rs
@@ -552,13 +552,13 @@ fn retain_fragments(
                     .get(fragment_spread.fragment_name.as_str())
                     .map(|fragment_def| {
                         (
-                            fragment_def.type_condition.as_str().to_string(),
+                            fragment_def.type_condition.as_str(),
                             fragment_def.selection_set.clone(),
                         )
                     });
 
                 if let Some((type_condition, fragment_selection_set)) = fragment_info
-                    && let Some(target_type) = tree_shaker.schema.types.get(type_condition.as_str())
+                    && let Some(target_type) = tree_shaker.schema.types.get(type_condition)
                 {
                     retain_type(
                         tree_shaker,
@@ -569,18 +569,15 @@ fn retain_fragments(
                 }
             }
             Selection::InlineFragment(inline_fragment) => {
-                if let Some(type_condition) = &inline_fragment.type_condition {
-                    let type_condition_str = type_condition.as_str().to_string();
-                    if let Some(target_type) =
-                        tree_shaker.schema.types.get(type_condition_str.as_str())
-                    {
-                        retain_type(
-                            tree_shaker,
-                            target_type,
-                            Some(&inline_fragment.selection_set),
-                            depth_limit,
-                        );
-                    }
+                if let Some(type_condition) = &inline_fragment.type_condition
+                    && let Some(target_type) = tree_shaker.schema.types.get(type_condition.as_str())
+                {
+                    retain_type(
+                        tree_shaker,
+                        target_type,
+                        Some(&inline_fragment.selection_set),
+                        depth_limit,
+                    );
                 }
             }
             Selection::Field(field) => {


crates/apollo-mcp-server/src/schema_tree_shake.rs
```diff
@@ -551,6 +538,61 @@ fn build_argument_name_to_value_map(arguments: &[Node<Argument>]) -> HashMap<&st
...
+                        Some(&fragment_selection_set),
+                        depth_limit,
+                    );
+                }
+            }
+            Selection::InlineFragment(inline_fragment) => {
+                if let Some(type_condition) = &inline_fragment.type_condition {
+                    let type_condition_str = type_condition.as_str().to_string();
+                    if let Some(target_type) =
+                        tree_shaker.schema.types.get(type_condition_str.as_str())
```
esilverm: Seems unnecessary to convert the `type_condition` to a `str` and then a `string` when we will just convert it back to a `str` afterwards. I think we can remove this variable and just use `type_condition.as_str()` inline instead.

crates/apollo-mcp-server/src/schema_tree_shake.rs
```diff
@@ -551,6 +538,61 @@ fn build_argument_name_to_value_map(arguments: &[Node<Argument>]) -> HashMap<&st
...
+                    .get(fragment_spread.fragment_name.as_str())
+                    .map(|fragment_def| {
+                        (
+                            fragment_def.type_condition.as_str().to_string(),
+                            fragment_def.selection_set.clone(),
+                        )
+                    });
+
+                if let Some((type_condition, fragment_selection_set)) = fragment_info
+                    && let Some(target_type) = tree_shaker.schema.types.get(type_condition.as_str())
```
esilverm: Similar thing here, is there a reason we can't just preserve the `type_condition` or the result of `type_condition.as_str()` so we don't have to re-as_str() it later?

esilverm (commented):
Should we maybe add an additional test to validate that we _can_ catch errors if the user attempts to add an invalid field to a fragment?

crates/apollo-mcp-server/src/schema_tree_shake.rs
alocay: If by later you mean within the `Selection::InlineFragment` block then it's a different match case. I think doing `as_str()` in both is fine.

crates/apollo-mcp-server/src/schema_tree_shake.rs
alocay: I agree, this does look a bit odd.

DaleSeo: Good catch! Removed the unnecessary conversion.

crates/apollo-mcp-server/src/schema_tree_shake.rs
DaleSeo: Changed to preserve the `&str` directly.

esilverm (approved):

PR #475, opened 2025-11-04
Improve operation update concurrency

dylan-apollo:


crates/apollo-mcp-server/src/server/states/running.rs
```diff
@@ -94,13 +94,18 @@ impl Running {
             operations.len(),
             serde_json::to_string_pretty(&operations)?
         );
-        *self.operations.lock().await = operations;
+        let mut operations_lock = self.operations.write().await;
+        *operations_lock = operations;
 
         // Update the schema itself
         *self.schema.lock().await = schema;
```
DaleSeo: Shouldn't we move this above `self.operations.write().await`?

I noticed that in the `update_schema()` function, we run `operations.write()` first and then `schema.lock()`. But in the `update_operations()` function below, we acquire the locks in the opposite order.

When one thread calls `update_schema()` and another thread calls `update_operations()`, this inconsistent lock ordering could potentially lead to a deadlock? 🤔

DaleSeo (commented):
Hey @dylan-apollo, thanks for the refactoring! I really appreciate the performance gain it will provide!

This PR isn't hitting our patch coverage target because no one has written tests for this core part of the codebase. 😂 This is not a required check but it could be a great chance for you to showcase your testing/mocking skills if you want to give it a try. Or, this could lead to another refactoring? 😆

Image

crates/apollo-mcp-server/src/server/states/running.rs
dylan-apollo: Oh good catch, and there's actually another bug in here too I think—because the locks aren't both held together, two threads could compete, _not_ deadlock, but end up with the schema not actually matching the operations 😱. I'll make sure one lock is held for the whole update process so that can't happen.

dylan-apollo: Done! The update schema or operations event might be a bit slower now, but it'll make sure that we never:

1. End up with a schema that doesn't match the operations
2. Read stale operations, update schema, then write back the stale operations

DaleSeo:
> ❌ **Changeset file missing for PR**
> 
> All changes should include an associated changeset file. Please refer to [README](https://github.com/apollographql/apollo-mcp-server/blob/main/.changesets/README.md) for more information on generating changesets.

Hey @dylan-apollo, you need to run `cargo xtask changeset create` and push the changeset to pass this check. To learn more, see https://github.com/apollographql/apollo-mcp-server/tree/main/.changesets.

dylan-apollo:
@DaleSeo changeset is added, I think this is ready for final review

.changesets/fix_confetti_receive_colon_education.md
```diff
@@ -0,0 +1,3 @@
+### Improved performance of parallel tool calls ([PR #475](https://github.com/apollographql/apollo-mcp-server/pull/475))
```
DaleSeo:
```suggestion
### Improved performance of parallel tool calls - @dylan-apollo PR #475
```

crates/apollo-mcp-server/src/server/states/running.rs
```diff
@@ -437,10 +448,67 @@ mod tests {
...
+
+        // Check that our local copy of operations is updated, representing what the server sees
+        let updated_operations = operations.read().await;
 
         assert_eq!(updated_operations.len(), 1);
         assert_eq!(updated_operations.first().unwrap().as_ref().name, "Valid");
     }
+
+    #[tokio::test]
+    async fn changing_schema_invalidates_outdated_operations() {
```
DaleSeo, line 461: Yay! 🎉

crates/apollo-mcp-server/src/server/states/running.rs
```diff
@@ -92,29 +94,34 @@ impl Running {
...
             operations.len(),
-            serde_json::to_string_pretty(&operations)?
+            serde_json::to_string_pretty(&operations).unwrap_or_default()
         );
-        *self.operations.lock().await = operations;
+        *operations_lock = operations;
 
         // Update the schema itself
-        *self.schema.lock().await = schema;
+        *self.schema.write().await = schema;
```
DaleSeo: Would switching lines 99 and 102 still work? If it does, would that be safer?

PR #477, opened 2025-11-08
docs: Add testing and deployment workflow guide

kbychu:
Adds a lightweight guide showing how to test and promote MCP server changes from local development through production. Connects existing documentation without being prescriptive about implementation details.

### DIFF
--- a/docs/source/_sidebar.yaml
+++ b/docs/source/_sidebar.yaml
@@ -10,6 +10,8 @@ items:
     href: "."
   - label: "Quickstart"
     href: "./quickstart"
+  - label: "Testing & Deployment Workflow"
+    href: "./testing-deployment-workflow"
   - label: "Define tools"
     href: "./define-tools"
   - label: "Configuration"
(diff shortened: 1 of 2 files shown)


docs/source/_sidebar.yaml
```diff
@@ -10,6 +10,8 @@ items:
     href: "."
   - label: "Quickstart"
     href: "./quickstart"
+  - label: "Testing & Deployment Workflow"
+    href: "./testing-deployment-workflow"
```
DaleSeo, line 14: @kbychu I really like this addition to the document, but I'm not sure it should be the 3rd item in the sidebar and the label looks a bit too verbose compare to existing ones. Maybe it would be better to move it under the deployment section?

@mabuyo Any thoughts?

Image

mabuyo: I agree it shouldn't be at the top! I think it should actually be under Guides. The contents even start with `This guide outlines...` It definitely feels more like a conceptual guide to me. 

I also want to rename it: "Development to Production Workflows". It's less about deployment specifically and more talking about how you work with the MCP server in different environments.

@kbychu any concerns here? I can make the changes as I do a more thorough review.

kbychu: @mabuyo Sounds good to me! No concerns.

PR #479, opened 2025-11-10
Docs: Suggestions for SDLC Guide

mabuyo:
@kbychu didn't want to commit these big changes to your branch without consulting and it was easier to create a new branch than make in-line suggestions!

Changes:
- Moves the guide under "Guides"
- Rename the page title to be a bit more accurate
- Prefer using inline links instead of "See XYZ page for ABC", because of how many links we use
- Prefer not to bold a lot of the actions
- Use numbered list to signify the order of steps
- Terminology consistency


docs/source/guides/dev-to-prod-workflows.mdx
```diff
@@ -0,0 +1,97 @@
...
+
+1. [Deploy an MCP server](/apollo-mcp-server/deploy) pointing to each variant.
+
+1. [Connect your AI agent](/apollo-mcp-server/quickstart#step-3-connect-to-an-mcp-client) to each environment and validate behavior.
+
+## Production environments
+
+1. Define your tools using either [local operation files](/apollo-mcp-server/define-tools#from-operation-files) or [persisted queries](/apollo-mcp-server/define-tools#from-graphos-managed-persisted-queries).
+
+    <ExpansionPanel title="Choosing the right source for your tools">
```
mabuyo, line 57: Is there anything more here that we can provide for guidance? Is it that Graph-OS managed PQ is enterprise-only?

kbychu: I think adoption of persisted query is just hard...if you haven't done it already, I am not sure it's worth the effort for this use case. Not sure how to reflect this in the docs

alocay (approved):

PR #481, opened 2025-11-11
Allow custom tools to be named the same as builtin tools

dylan-apollo:


esilverm:
This looks good, perhaps we should allow the ability to alias these built-in tool names as well in case someone wants both

dylan-apollo:
> Possibly something worthy of a followup?

Yeah I think so—probably a config value under each one of those tools' config blocks to set a different name? Maybe even a different description if they want to tweak how agents use them.

PR #485, opened 2025-11-13
docs: add metrics telemetry configuration examples for New Relic

DaleSeo:


docs/source/config-file.mdx
```diff
@@ -335,6 +335,26 @@ telemetry:
           key-two: another-value
 ```
 
+For services like New Relic, use `delta` temporality and pass your API key via metadata:
```
mabuyo: What do we mean here by _like_ New Relic? Can we make it a bit more general? Or is this specifically for New Relic?

DaleSeo: Hey @mabuyo, thanks for you feedback! I've made it a bit general.

PR #493, opened 2025-11-14
Implement deduplication of operations

### LINKED ISSUE #491: Folder and explicit path overlap can create duplicate tools
If your configuration uses local sourced operations and the paths overlap, Apollo MCP Server will create duplicate tools which will cause an error for any Agent connecting to that server.

Example `mcp.yaml` configuration:

```yaml
operations:
  source: local
  paths:
    - tools
    - ./tools/GetCustomerAccountsTransactions.graphql
```

**Expected Outcome:**
Only one tool named `GetCustomerAccountsTransactions` is returned from the `tools/list` call 

**Observed Outcome:**
Two tools named `GetCustomerAccountsTransactions` are returned from the `tools/list` call

Image

**Version information**

Apollo MCP Server: 1.2
Apollo Router: 2.8.2
Rover: 0.36.2

DaleSeo:
Fixes #491 

When configuring Apollo MCP Server with local operations, if the configuration includes both:
- A directory path (e.g., `tools`)
- An explicit file path within that directory (e.g., `./tools/GetCustomerAccountsTransactions.graphql`)

The server would create duplicate tools with the same name, causing errors for MCP clients. This PR adds deduplication logic based on canonical file paths before emitting the `OperationsUpdated` event.

### DIFF
--- a/crates/apollo-mcp-server/src/operations/operation_source.rs
+++ b/crates/apollo-mcp-server/src/operations/operation_source.rs
@@ -156,7 +156,15 @@ impl OperationSource {
                                                 let canonical_path = PathBuf::from(source_path)
                                                     .canonicalize()
                                                     .unwrap_or_else(|_| PathBuf::from(source_path));
-                                                seen_paths.insert(canonical_path)
+                                                let is_new =
+                                                    seen_paths.insert(canonical_path.clone());
+                                                if !is_new {
+                                                    tracing::debug!(
+                                                        ?canonical_path,
+                                                        "Filtered duplicate operation"
+                                                    );
+                                                }
+                                                is_new
                                             } else {
                                                 // If there's no source path, include the operation
                                                 true


alocay (approved):

crates/apollo-mcp-server/src/operations/operation_source.rs
```diff
@@ -145,9 +145,34 @@ impl OperationSource {
...
+                                        .values()
+                                        .flatten()
+                                        .filter(|op| {
+                                            if let Some(source_path) = &op.source_path {
+                                                // Try to canonicalize the path, fall back to the original if it fails
+                                                let canonical_path = PathBuf::from(source_path)
+                                                    .canonicalize()
+                                                    .unwrap_or_else(|_| PathBuf::from(source_path));
+                                                let is_new =
+                                                    seen_paths.insert(canonical_path.clone());
```
esilverm, line 160: This code seems to build two parallel sets of paths. Perhaps instead we should be iterating over the `state.values()` and adding it to the `HashSet` and then converting that into our result vec. Otherwise it feels like a misuse of memory

PR #500, opened 2025-11-25
Load and serve basic apps

dylan-apollo:
This PR introduces the concept of "apps" which, right now, correspond to [OpenAPI Apps SDK](https://developers.openai.com/apps-sdk), but can be more broadly understood as [MCP Apps](https://blog.modelcontextprotocol.io/posts/2025-11-21-mcp-apps/) once that is stabilized.

At startup, the new `apps::load_from_path` function is called, right now looking for a hard-coded `apps/` folder, though we can make that configurable in the future. Each app is a directory within that `apps/` folder which contains an `.application-manifest.json`. An app then registers as a tool (exposed in `list_tools` and `call_tool`) and a resource (`list_resources` and `read_resource`).

Apps won't actually be loaded unless the cargo feature `"apps"` is enabled. I only gated the actual runtime component with that feature flag right now, but we could optionally skip compiling the entire module if we want.

The only immediate effect of the default build of this server after this PR is that the server will now report it has resource capabilities. Since the resource list will always be empty, I think that change is very minor. But we could disable that for now as well if it's concerning.

### DIFF
--- a/crates/apollo-mcp-server/src/apps.rs
+++ b/crates/apollo-mcp-server/src/apps.rs
@@ -1,0 +1,226 @@
+use std::fs::read_to_string;
+use std::path::Path;
+
+use apollo_compiler::{Schema, validation::Valid};
+use rmcp::model::{Meta, RawResource, Resource};
+use serde::Deserialize;
+use tracing::debug;
+use url::Url;
+
+use crate::{
+    custom_scalar_map::CustomScalarMap,
+    operations::{MutationMode, Operation, RawOperation},
+};
+
+/// An app, which consists of a tool and a resource to be used together.
+#[derive(Clone, Debug)]
+pub(crate) struct App {
+    /// The GraphQL operation that defines the app's tool
+    pub(crate) operation: Operation,
+    /// The HTML resource that serves as the app's UI
+    pub(crate) resource: String,
+    /// The URI of the app's resource
+    pub(crate) uri: Url,
+}
+
+impl App {
+    pub(crate) fn resource(&self) -> Resource {
+        Resource::new(
+            RawResource {
+                name: self.operation.tool.name.clone().into(),
+                uri: self.uri.to_string(),
+                mime_type: Some("text/html+skybridge".to_string()),
+                // TODO: load all this from a manifest file
+                title: None,
+                description: None,
+                icons: None,
+                size: None,
+            },
+            None,
+        )
+    }
+}
+
+const MANIFEST_FILE_NAME: &str = ".application-manifest.json";
+
+pub(crate) fn load_from_path(
+    path: &Path,
+    schema: &Valid<Schema>,
+    custom_scalar_map: Option<&CustomScalarMap>,
+    mutation_mode: MutationMode,
+    disable_type_description: bool,
+    disable_schema_description: bool,
+) -> Result<Vec<App>, String> {
+    let Ok(apps_dir) = path.read_dir() else {
+        return Ok(Vec::new());
+    };
+
+    let mut apps = Vec::new();
+    for app_dir in apps_dir {
+        let app_dir = match app_dir {
+            Ok(app_dir) => app_dir,
+            Err(err) => {
+                debug!("Failed to read app directory, ignoring: {}", err);
+                continue;
+            }
+        };
+        let path = app_dir.path();
+        if !path.is_dir() {
+            debug!("{} is not a directory, ignoring", path.to_string_lossy());
+            continue;
+        }
+
+        let Ok(manifest) = read_to_string(path.join(MANIFEST_FILE_NAME)) else {
+            debug!(
+                "No manifest file found in {}, ignoring",
+                path.to_string_lossy()
+            );
+            continue;
+        };
+
+        let manifest: Manifest = serde_json::from_str(&manifest).map_err(|err| {
+            format!(
+                "Failed to parse manifest from {}: {}",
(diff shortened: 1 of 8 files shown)


dylan-apollo:
Because these changes shouldn't be user-facing yet, I'm not including a changeset. We can add one in describing the new apps feature once we're ready for folks to test it out.

crates/apollo-mcp-server/src/server/states/running.rs
```diff
@@ -338,10 +342,53 @@ impl ServerHandler for Running {
...
+        let request_uri = Url::parse(&request.uri).map_err(|err| {
+            ErrorData::resource_not_found(
+                format!("Requested resource has an invalid URI: {err}"),
+                None,
+            )
+        })?;
+        let Some(app) = self
+            .apps
+            .iter()
+            .find(|app| app.uri.path() == request_uri.path())
```
DaleSeo: Any reason to only match the path?

crates/apollo-mcp-server/src/server/states/running.rs
```diff
@@ -361,6 +408,7 @@ impl ServerHandler for Running {
             capabilities: ServerCapabilities::builder()
                 .enable_tools()
                 .enable_tool_list_changed()
+                .enable_resources()
```
DaleSeo: > The only immediate effect of the default build of this server after this PR is that the server will now report it has resource capabilities. Since the resource list will always be empty, I think that change is very minor. But we could disable that for now as well if it's concerning.

I agree this isn't a big deal, but just to be safe, I suggest not enabling resources when apps is empty. This will help prevent issues for MCP clients that don't support resources. There are lots of them listed here: https://modelcontextprotocol.io/clients

crates/apollo-mcp-server/src/apps.rs
```diff
@@ -0,0 +1,219 @@
...
+            )
+        })?;
+
+        let Some(operation_str) = manifest
+            .operations
+            .iter()
+            .find_map(|op| op.prefetch_id.is_some().then(|| op.body.to_string()))
+        else {
+            // TODO: Allow applications with only post-fetch operations
+            return Err("Exactly one prefetch operation must be defined".into());
```
DaleSeo, line 95: It seems like this error won't occur when there are multiple prefetch operations, right? because `find_map` will take the first matching one if multiple exists?

crates/apollo-mcp-server/src/apps.rs
```diff
@@ -0,0 +1,219 @@
...
+        let path = app_dir.path();
+        if !path.is_dir() {
+            // Don't care about standalone files in this directory
+            continue;
+        }
+
+        let Ok(manifest) = read_to_string(path.join(MANIFEST_FILE_NAME)) else {
+            // A directory without a manifest file is not an application, might be some other build artifact
+            continue;
+        };
```
DaleSeo: How about we add some debug logs before `continue`?

crates/apollo-mcp-server/src/server/states/running.rs
dylan-apollo: Yes, we put a hash in the fragment to bust ChatGPT's resource cache, so that will not match during redeploys. If you deploy v2 of an app and v1's tool/resource definitions are still in-memory in someone's session, you don't want to get errors when someone requests the v1 resource (unless it was truly removed).

crates/apollo-mcp-server/src/server/states/running.rs
```diff
@@ -338,10 +342,53 @@ impl ServerHandler for Running {
                 .chain(self.search_tool.as_ref().iter().map(|e| e.tool.clone()))
                 .chain(self.explorer_tool.as_ref().iter().map(|e| e.tool.clone()))
                 .chain(self.validate_tool.as_ref().iter().map(|e| e.tool.clone()))
+                .chain(self.apps.iter().map(|app| app.operation.as_ref().clone()))
                 .collect(),
         })
     }
 
+    async fn list_resources(
```
DaleSeo, line 391: Do we have any o11y requirements? If we do, you can consider adding `#[tracing::instrument...` on this.

crates/apollo-mcp-server/src/apps.rs
dylan-apollo: Yeah, this was enforced in the front end tooling which builds the manifest. We will actually have a follow-up PR to support multiple prefetch operations per app (coming soon!)

DaleSeo:
Can we add some tests for the new functions added in `running.rs` to meet the coverage target? You can check what is not covered [here](https://app.codecov.io/gh/apollographql/apollo-mcp-server/pull/500?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=apollographql).

crates/apollo-mcp-server/src/server/states/running.rs
dylan-apollo: We don't have any specific requirements, but I added similar instruments to the ones on the equivalent tool functions.

DaleSeo:
> Because these changes shouldn't be user-facing yet, I'm not including a changeset. We can add one in describing the new apps feature once we're ready for folks to test it out.

Tip: If you attach the following label to your PR, this annoying comment will stop being added.

2025-11-26 at 12 10 22

DaleSeo (approved):
Thanks for addressing my feedback!

PR #502, opened 2025-11-26
App URL entrypoints

dylan-apollo:
In addition to local resource files loaded into memory, users can define _remote_ resources for apps that are dynamically fetched each time they're requested. This enables users to deploy their front-end code wherever they wish, like to a CDN.


crates/apollo-mcp-server/src/apps.rs
DaleSeo:
```suggestion
    fn test_local_resource() {
```

DaleSeo (approved):
Looks good to me! I just have a small suggestion for renaming.

PR #504, opened 2025-11-26
Multiple tool definitions per app

dylan-apollo:
This PR decouples the tool definition from the app definition, and allows the user to configure multiple tools per app (and per operation).

Each tool is effectively an action that the agent wants to take. It's very common in ChatGPT apps to have several actions all result in the same UI.

The reasons to allow multiple tools per _operation_ are:

1. Allow multiple descriptions, which may help an agent pick a tool.
2. Allow _inputs_ that don't actually go to the GraphQL operation, only to the UI (coming in a follow-up PR)

### DIFF
--- a/crates/apollo-mcp-server/src/apps.rs
+++ b/crates/apollo-mcp-server/src/apps.rs
@@ -1,8 +1,8 @@
-use std::fs::read_to_string;
 use std::path::Path;
+use std::{fs::read_to_string, sync::Arc};
 
 use apollo_compiler::{Schema, validation::Valid};
-use rmcp::model::{Meta, RawResource, Resource};
+use rmcp::model::{Meta, RawResource, Resource, Tool};
 use serde::Deserialize;
 use tracing::debug;
 use url::Url;
@@ -15,12 +15,22 @@ use crate::{
 /// An app, which consists of a tool and a resource to be used together.
 #[derive(Clone, Debug)]
 pub(crate) struct App {
-    /// The GraphQL operation that defines the app's tool
-    pub(crate) operation: Operation,
+    pub(crate) name: String,
     /// The HTML resource that serves as the app's UI
     pub(crate) resource: AppResource,
     /// The URI of the app's resource
     pub(crate) uri: Url,
+    /// Entrypoint tools for this app
+    pub(crate) tools: Vec<AppTool>,
+}
+
+/// An MCP tool which serves as an entrypoint for an app.
+#[derive(Clone, Debug)]
+pub(crate) struct AppTool {
+    /// The GraphQL operation that's executed when the tool is called. Its data is injected into the UI
+    pub(crate) operation: Arc<Operation>,
+    /// The MCP tool definition
+    pub(crate) tool: Tool,
 }
 
 #[derive(Clone, Debug)]
@@ -33,7 +43,7 @@ impl App {
     pub(crate) fn resource(&self) -> Resource {
         Resource::new(
             RawResource {
-                name: self.operation.tool.name.clone().into(),
+                name: self.name.clone(),
                 uri: self.uri.to_string(),
                 mime_type: Some("text/html+skybridge".to_string()),
                 // TODO: load all this from a manifest file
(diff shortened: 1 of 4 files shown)


crates/apollo-mcp-server/src/apps.rs
```diff
@@ -92,38 +102,69 @@ pub(crate) fn load_from_path(
...
+                        "Failed parsing tools: No operation in {path}",
+                        path = path.to_string_lossy()
+                    ));
+                }
+                Ok(Some(op)) => Arc::new(op),
+            };
+
+            for tool in operation_def.tools {
+                let tool = Tool {
+                    name: format!("{name}--{}", tool.name).into(),
```
DaleSeo, line 146: Any chance an app or tool name contains `--`?

crates/apollo-mcp-server/src/apps.rs
```diff
@@ -92,38 +102,69 @@ pub(crate) fn load_from_path(
...
+                    meta: Some(meta.clone()),
+                    description: Some(
+                        if let Some(app_description) = manifest.description.clone() {
+                            format!("{} {}", app_description, tool.description).into()
```
DaleSeo, line 150: What if descriptions include spaces?

crates/apollo-mcp-server/src/apps.rs
```diff
@@ -15,12 +15,22 @@ use crate::{
...
     /// The URI of the app's resource
     pub(crate) uri: Url,
+    /// Entrypoint tools for this app
+    pub(crate) tools: Vec<AppTool>,
+}
+
+/// An MCP tool which serves as an entrypoint for an app.
+#[derive(Clone, Debug)]
+pub(crate) struct AppTool {
+    /// The GraphQL operation that's executed when the tool is called. It's data is injected into the UI
```
DaleSeo:
```suggestion
    /// The GraphQL operation that's executed when the tool is called. Its data is injected into the UI
```

crates/apollo-mcp-server/src/apps.rs
```diff
@@ -15,12 +15,22 @@ use crate::{
...
+    pub(crate) operation: Arc<Operation>,
```
DaleSeo, line 31: Have you considered introducing something like `AppOperation`?

crates/apollo-mcp-server/src/apps.rs
```diff
@@ -280,4 +316,65 @@ mod test_load_from_path {
...
+            None,
+            MutationMode::All,
+            false,
+            false,
+        )
+        .expect("Failed to load apps");
+        assert_eq!(apps.len(), 1);
+        let app = &apps[0];
+        assert_eq!(app.tools.len(), 3);
+        assert_eq!(app.tools[0].tool.name, "MultipleToolsApp--Tool1");
```
DaleSeo, line 360: Hmm... This chaining API isn't very ergonomic to me.  I expected `app.tools[0]` to return the tool directly and needed to look at the `AppTool` struct again.

crates/apollo-mcp-server/src/apps.rs
dylan-apollo: These don't need to be deterministic, just had to choose some separator. We'll probably change it over time if we realize there's a better way to have ChatGPT recognized it.

crates/apollo-mcp-server/src/apps.rs
dylan-apollo: These don't need to be deterministic, just had to choose some separator. We'll probably change it over time if we realize there's a better way to have ChatGPT recognized it.

crates/apollo-mcp-server/src/apps.rs
dylan-apollo: It comes from wanting to re-use a lot of existing structures, like `Operation` and `rmcp::Tool`. Most of what `Operation` does we don't need anymore, but refactoring all that code is a big task.

crates/apollo-mcp-server/src/apps.rs
dylan-apollo: You mean to use instead of `Operation`? That's probably ideal, but it'd be a lot of refactoring which should probably happen separate.

PR #505, opened 2025-11-26
Allow multiple operations per App tool

dylan-apollo:
This builds on #504, so that PR should be merged first.

An app now consists of multiple entrypoints (tools) from #504. Additionally, every time an app loads, it can load _any number_ of operations (which we call prefetch operations).

This PR changes the execution flow of an app from "execute the equivalent operation" to:

1. Execute the initial operation
2. In _parallel_, execute any additional prefetch operations (that don't match that first one)
3. Combine all results under new nested keys in the final results, so the UI code knows where to find each one

### DIFF
--- a/crates/apollo-mcp-server/src/apps.rs
+++ b/crates/apollo-mcp-server/src/apps.rs
@@ -30,6 +30,12 @@ pub(crate) struct App {
     pub(crate) prefetch_operations: Vec<PrefetchOperation>,
 }
 
+#[derive(Clone, Debug)]
+pub(crate) enum AppResource {
+    Local(String),
+    Remote(Url),
+}
+
 /// An MCP tool which serves as an entrypoint for an app.
 #[derive(Clone, Debug)]
 pub(crate) struct AppTool {
@@ -46,12 +52,6 @@ pub(crate) struct PrefetchOperation {
     pub(crate) operation: Arc<Operation>,
     /// A unique ID for the operation that the UI will use to look up its data
     pub(crate) prefetch_id: String,
-}
-
-#[derive(Clone, Debug)]
-pub(crate) enum AppResource {
-    Local(String),
-    Remote(Url),
 }
 
 impl App {
--- a/crates/apollo-mcp-server/src/apps/execution.rs
+++ b/crates/apollo-mcp-server/src/apps/execution.rs
@@ -32,7 +32,7 @@ pub(crate) async fn find_and_execute_app(
     None
 }
 
-pub(crate) async fn execute_app(
+async fn execute_app(
     app: &App,
     tool: &AppTool,
     headers: &HeaderMap,


crates/apollo-mcp-server/src/server/states/running.rs
```diff
@@ -368,21 +369,16 @@ impl ServerHandler for Running {
                     .execute(graphql_request)
                     .with_context(Context::current())
                     .await
-            } else if let Some(tool) = self
-                .apps
-                .iter()
-                .find_map(|app| app.tools.iter().find(|tool| tool.tool.name == tool_name))
+            } else if let Some(res) = find_and_execute_app(
```
DaleSeo, line 372: Really like this function encapsulating all the complexity of app execution away. However, I noticed there's a bit of inconsistency now in how operations and apps are handled.

- For operations, we have separate find and execute steps.
- For apps, we combine find and execute.

Would it make sense to refactor operations to follow the same pattern? This could make the logic in `call_tool` more uniform and easier to follow.

Not blocking, just a thought for improving consistency across the codebase.

crates/apollo-mcp-server/src/apps.rs
```diff
@@ -33,6 +39,15 @@ pub(crate) struct AppTool {
...
+#[derive(Clone, Debug)]
+pub(crate) struct PrefetchOperation {
+    /// The operation to execute
+    pub(crate) operation: Arc<Operation>,
+    /// A unique ID for the operation that the UI will use to look up its data
+    pub(crate) prefetch_id: String,
+}
+
 #[derive(Clone, Debug)]
 pub(crate) enum AppResource {
```
DaleSeo: nit: I thought it would be helpful to put this before `AppTool` so they appear in the same order as they do in `App`.

crates/apollo-mcp-server/src/apps/execution.rs
```diff
@@ -0,0 +1,291 @@
...
+        for tool in &app.tools {
+            if tool.tool.name == tool_name {
+                return Some(execute_app(app, tool, headers, arguments, endpoint).await);
+            }
+        }
+    }
+    None
+}
+
+pub(crate) async fn execute_app(
```
DaleSeo:
```suggestion
async fn execute_app(
```

crates/apollo-mcp-server/src/apps/execution.rs
```diff
@@ -0,0 +1,291 @@
...
+        prefetch_calls.push(
+            prefetch
+                .operation
+                .execute(graphql_request)
+                .with_context(Context::current())
+                .map_ok(|res| (prefetch.prefetch_id.clone(), res)),
+        );
+    }
+
+    let prefetch_results = try_join_all(prefetch_calls.into_iter()).await?;
```
DaleSeo, line 75: With `try_join_all`, it looks like if any prefetch operation fails, the entire tool call fails. Is this the desired behavior?

crates/apollo-mcp-server/src/server/states/running.rs
andrewmcgivery: I think that refactoring would absolutely make sense but my preference would be to have that as a separate PR to keep this one scoped to the app changes. I don't want to scope creep this PR and accidentally break normal operations in the process 😅

crates/apollo-mcp-server/src/apps/execution.rs
andrewmcgivery: I think this is a "good enough" behaviour for the purposes of this PR but this is a good callout and I do want to revisit this with @dylan-apollo when they are back because it's possible that we would want (and could allow for) partial errors in tool responses since they are split out into separate parts of the `structuredContent`.

crates/apollo-mcp-server/src/server/states/running.rs
DaleSeo: OK, I'll follow up on the refactoring! 😄

PR #507, opened 2025-11-26
Allow defining additional inputs (not related to operations)

dylan-apollo:
Based on #506, that one should be merged first.

This allows the UI-side of the app to define additional tool inputs that the agent should provide for the UI-only. Those inputs are not sent to any GraphQL operations, they are simply passed through (the UI gets access to them automatically on the `window` object).

The filtering we're _already_ doing with #505 will prevent these extra arguments from ending up in the GraphQL requests.

### DIFF
--- a/crates/apollo-mcp-server/src/apps.rs
+++ b/crates/apollo-mcp-server/src/apps.rs
@@ -173,7 +173,7 @@ pub(crate) fn load_from_path(
                     ),
                     input_schema: if let Some(extra_inputs) = tool.extra_inputs {
                         let mut merged = operation.tool.input_schema.as_ref().clone();
-                        merge_inputs(&mut merged, extra_inputs);
+                        merge_inputs(&mut merged, extra_inputs)?;
                         Arc::new(merged)
                     } else {
                         operation.tool.input_schema.clone()
@@ -227,7 +227,10 @@ pub(crate) fn load_from_path(
     Ok(apps)
 }
 
-fn merge_inputs(orig: &mut Map<String, Value>, extra: Vec<ExtraInputDefinition>) {
+fn merge_inputs(
+    orig: &mut Map<String, Value>,
+    extra: Vec<ExtraInputDefinition>,
+) -> Result<(), String> {
     let mut properties = match orig.remove("properties") {
         Some(Value::Object(props)) => props,
         _ => Map::default(),
@@ -239,6 +242,13 @@ fn merge_inputs(orig: &mut Map<String, Value>, extra: Vec<ExtraInputDefinition>)
     };
 
     for extra_input in extra {
+        if properties.contains_key(&extra_input.name) {
+            return Err(format!(
+                "Extra input with name '{}' failed to process because another input with this name was already processed. Make sure your extra_input names are unique, both from each other and any graphql variables you may have.",
+                extra_input.name
+            ));
+        }
+
         if extra_input.required {
             let value = Value::String(extra_input.name.clone());
             if !required.contains(&value) {
@@ -257,6 +267,8 @@ fn merge_inputs(orig: &mut Map<String, Value>, extra: Vec<ExtraInputDefinition>)
 
     orig.insert("properties".to_string(), Value::Object(properties));
     orig.insert("required".to_string(), Value::Array(required));
+
+    Ok(())
 }
 
 #[derive(Clone, Deserialize)]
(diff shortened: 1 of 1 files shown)


crates/apollo-mcp-server/src/apps.rs
```diff
@@ -218,6 +225,50 @@ pub(crate) fn load_from_path(
...
+    let mut properties = orig
+        .remove("properties")
+        .and_then(|val| {
+            if let Value::Object(properties) = val {
+                Some(properties)
+            } else {
+                None
+            }
+        })
+        .unwrap_or_default();
```
DaleSeo: Suggesting simpler code:

```suggestion
    let mut properties = match orig.remove("properties") {
        Some(Value::Object(props)) => props,
        _ => Map::default(),
    };
```

crates/apollo-mcp-server/src/apps.rs
```diff
@@ -218,6 +225,50 @@ pub(crate) fn load_from_path(
...
+    let mut required = orig
+        .remove("required")
+        .and_then(|val| {
+            if let Value::Array(required) = val {
+                Some(required)
+            } else {
+                None
+            }
+        })
+        .unwrap_or_default();
```
DaleSeo: ditto

```suggestion
    let mut required = match orig.remove("required") {
        Some(Value::Array(req)) => req,
        _ => Vec::default(),
    };
```

crates/apollo-mcp-server/src/apps.rs
```diff
@@ -218,6 +225,50 @@ pub(crate) fn load_from_path(
...
+        }
+
+        properties.insert(
+            extra_input.name,
+            json!({
+                "description": extra_input.description,
+                "type": extra_input.value_type
+            }),
+        );
+    }
```
DaleSeo, line 266: What if `extra_input.name` matches an existing property name?

crates/apollo-mcp-server/src/apps.rs
```diff
@@ -258,6 +309,18 @@ struct OperationDefinition {
...
+    #[serde(rename = "extraInputs", default)]
+    extra_inputs: Option<Vec<ExtraInputDefinition>>,
+}
+
+#[derive(Clone, Deserialize)]
+struct ExtraInputDefinition {
+    name: String,
+    description: String,
+    #[serde(rename = "type")]
+    value_type: String,
```
DaleSeo: Shouldn't we validate against known JSON Schema types like `string`, `number` and etc?

crates/apollo-mcp-server/src/apps.rs
andrewmcgivery: Hm. Good question. The result would be that the property is overwritten. This could happen in two scenarios...

1. The graphql variable name is the same as an extra_input name or
2. The user has defined the same extra_input name twice

I suspect we should probably throw an error in this scenario because that is probably not what the user is intending to do.

I'll take a look at this. :)

crates/apollo-mcp-server/src/apps.rs
andrewmcgivery: Thats probably a good idea. I'll make this into an enum instead of a string so that it is validated as part of the parsing of the JSON.

crates/apollo-mcp-server/src/apps.rs
andrewmcgivery: Done!

crates/apollo-mcp-server/src/apps.rs
andrewmcgivery: Done!

crates/apollo-mcp-server/src/apps.rs
```diff
@@ -218,6 +227,50 @@ pub(crate) fn load_from_path(
...
+        Some(Value::Array(req)) => req,
+        _ => Vec::default(),
+    };
+
+    for extra_input in extra {
+        if properties.contains_key(&extra_input.name) {
+            return Err(format!(
+                "Extra input with name '{}' failed to process because another input with this name was already processed. Make sure your extra_input names are unique, both from each other and any graphql variables you may have.",
+                extra_input.name
+            ));
```
DaleSeo, line 249: 👍

crates/apollo-mcp-server/src/apps.rs
```diff
@@ -258,6 +311,42 @@ struct OperationDefinition {
...
+        })
+    }
+}
+
+#[derive(Clone, Deserialize)]
+struct ExtraInputDefinition {
+    name: String,
+    description: String,
+    #[serde(rename = "type")]
+    value_type: JsonSchemaType,
```
DaleSeo, line 347: 🎉

PR #514, opened 2025-12-03
Fix broken non nullable return types in minified schema

esilverm:
When minifying return types, we were mistakenly branching only at list and non-null without handling the non-null list option. Additionally, any non-nullable types within the list brackets were being treated as nullable. As a result, the LLM would fail to build a valid operation in the case that an input object falls into these categories and because the minified schema is the source of truth, the LLM would endlessly iterate without the ability to know that they were missing a required parameter (even if mentioned in the description).

This adjusts the handling by properly checking all cases and recursively building list return types.

Addresses AIR-71

### DIFF
--- a/crates/apollo-mcp-server/src/introspection/minify.rs
+++ b/crates/apollo-mcp-server/src/introspection/minify.rs
@@ -226,7 +226,8 @@ fn type_name(ty: &Type) -> String {
     match ty {
         Type::List(inner) => format!("[{}]", type_name(inner)),
         Type::NonNullList(inner) => format!("[{}]!", type_name(inner)),
-        Type::Named(name) => shorten_scalar_names(name.as_str()).to_string(),
+        #[allow(clippy::useless_format)]
+        Type::Named(name) => format!("{}", shorten_scalar_names(name.as_str())),
         Type::NonNullNamed(name) => format!("{}!", shorten_scalar_names(name.as_str())),
     }
 }


crates/apollo-mcp-server/src/introspection/minify.rs
```diff
@@ -223,13 +223,11 @@ fn format_interfaces(
...
-    if ty.is_list() {
-        format!("[{name}]")
-    } else if ty.is_non_null() {
-        format!("{name}!")
-    } else {
-        name.to_string()
+    match ty {
+        Type::List(inner) => format!("[{}]", type_name(inner)),
+        Type::NonNullList(inner) => format!("[{}]!", type_name(inner)),
+        Type::Named(name) => shorten_scalar_names(name.as_str()).to_string(),
```
DaleSeo: nit: Had to look twice to understand why `format!()` isn't used here like the other arms. Using it consistently might make the pattern easier to scan at a glance.

```suggestion
        Type::Named(name) => format!("{}", shorten_scalar_names(name.as_str())),
```

DaleSeo (approved):
Great fix, @esilverm! It might be worth adding a specific test case for deeply nested types like `[[ID!]!]!` to prevent any regressions.

crates/apollo-mcp-server/src/introspection/minify.rs
esilverm: I see what you mean, but when I do this I get a clippy warning for useless `format!` so I'll stick with what I have and add the specific test case to the snapshot!

PR #515, opened 2025-12-03
Add release-experimental-yml workflow to allow releasing experimental builds of the binary

andrewmcgivery:


.github/workflows/release-experimental.yml
```diff
@@ -0,0 +1,233 @@
...
+
+jobs:
+  validate:
+    name: Validate version format
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Validate experimental version pattern
+        run: |
+          if ! echo "${{ inputs.version }}" | grep -qE '^[0-9]+\.[0-9]+\.[0-9]+-experimental\.[0-9]+$'; then
+            echo "Error: Version must match pattern: X.Y.Z-experimental.N"
```
DaleSeo, line 26: Like this safeguard!

.github/workflows/release-experimental.yml
```diff
@@ -0,0 +1,233 @@
...
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: Version to publish (e.g., 1.2.1-experimental.1)
+        required: true
+        type: string
+      features:
+        description: Cargo features (comma-separated, e.g., apps)
+        required: false
```
DaleSeo, line 11: Shouldn't be required?

.github/workflows/release-experimental.yml
```diff
@@ -0,0 +1,233 @@
...
+        run: |
+          ./release/x86_64-unknown-linux-musl/dist/config-schema > artifacts/config.schema.json
+
+      - name: Upload release artifacts
+        uses: softprops/action-gh-release@v2
+        with:
+          tag_name: v${{ env.VERSION }}
+          files: artifacts/*
+          prerelease: ${{ contains(env.VERSION, '-experimental.') }}
+          make_latest: false # this runs for each combination in the matrix - don't mark as latest until all are done
```
DaleSeo: This comment doesn't apply to experimental releases since they won't ever be marked as the latest.

.github/workflows/release-experimental.yml
```diff
@@ -0,0 +1,233 @@
...
+            build-${{ runner.os }}-
+          purge: true
+          purge-prefixes: release-${{ matrix.bundle }}-
+          purge-created: 0
+          purge-primary-key: never
+          gc-max-store-size: 5G
+
+      - name: Build binaries
+        env:
+          APOLLO_MCP_FEATURES: ${{ inputs.features }}
```
DaleSeo: It would be helpful if the name clearly indicates that this environment variable is only used at build time.

DaleSeo (approved):
I only have non-blocking comments.

.github/workflows/release-experimental.yml
andrewmcgivery: I left it as optional just incase someone just wants to do an experimental build of what is currently in dev, even if what they have there isn't behind feature flags!

PR #517, opened 2025-12-04
Abstract away operation execution logic from the server's running state

DaleSeo:
This is a follow-up to my review comment https://github.com/apollographql/apollo-mcp-server/pull/505#discussion_r2569985220 about making operations and apps more consistent. I abstracted the operation execution logic from the server's running state, following the pattern used in apps. This change helped me write tests and identify a subtle bug where the execute tool wasn't propagating the OTel context.

### DIFF
--- a/.changesets/maint_operation_execution.md
+++ b/.changesets/maint_operation_execution.md
@@ -1,0 +1,3 @@
+### Abstract away operation execution logic from the server's running state - @DaleSeo PR #517
+
+I abstracted the operation execution logic from the server's running state, following the pattern used in apps. This change helped me write tests and identify a subtle bug where the execute tool wasn't propagating the OTel context.


.changesets/maint_operation_execution.md
```diff
@@ -0,0 +1,3 @@
+### Abstract away operation execution logic from the server's running state - @DaleSeo PR #517
```
andrewmcgivery, line 1: Does this actually need a changeset? I wouldn't expect this internal reflect to be relevant to customers.

andrewmcgivery (approved):
For what it's worth... LGTM :D

.changesets/maint_operation_execution.md
DaleSeo: Yeah customers won't care but we've just been keeping maintenance changesets for our own tracking.

PR #524, opened 2025-12-05
Bug: raw operation variables is always null so inputs check was always failing on app tool graphql calls

andrewmcgivery:


crates/apollo-mcp-server/src/operations/operation.rs
```diff
@@ -29,6 +29,8 @@ pub struct Operation {
     pub(crate) tool: Tool,
     pub(crate) inner: RawOperation,
     operation_name: String,
+    // Holds the raw variables defined in an operation
+    pub(crate) variable_definitions: HashMap<String, String>,
```
DaleSeo: Since we're just using the keys for the filter check, I guess we simplify this to `variable_names: HashSet<String>`?

crates/apollo-mcp-server/src/apps/execution.rs
```diff
@@ -127,13 +127,7 @@ fn filter_inputs_for_operation(
...
         inputs
             .iter()
-            .filter(|(key, _)| {
-                operation
-                    .inner
-                    .variables
-                    .as_ref()
-                    .is_some_and(|variables| variables.contains_key(*key))
-            })
+            .filter(|(key, _)| operation.variable_definitions.contains_key(*key))
```
DaleSeo: The `tool.input_schema` already has the variable definitions. Have you thought about checking that instead?

The code will look something like this:

```rs
fn filter_inputs_for_operation(
    inputs: Option<&JsonObject>,
    operation: &Operation,
) -> Option<JsonObject> {
    let inputs = inputs?;
    let properties = operation.tool.input_schema.get("properties")?.as_object()?;
    Some(
        inputs
            .iter()
            .filter(|(key, _)| properties.contains_key(*key))
            .map(|(key, value)| (key.clone(), value.clone()))
            .collect(),
    )
}
```

This might be a simpler approach that avoids modifying the `Operation` struct entirely.

DaleSeo:
> ❌ **Changeset file missing for PR**
> 
> All changes should include an associated changeset file. Please refer to [README](https://github.com/apollographql/apollo-mcp-server/blob/main/.changesets/README.md) for more information on generating changesets.

Tip: If you don't want to include a changeset, you can attach the following label to your PR.

image

crates/apollo-mcp-server/src/apps/execution.rs
andrewmcgivery: `inputs` and `input_schema` are going to be exactly the same... and this is not what we're checking here. :)

We have to explicitly check against the actual graphql variables because the `inputs` for app tools is a mixture of two things:

- The GraphQL variable inputs
- Any defined `extra_inputs`

We need to filter down all the inputs passed in to _only_ the graphql variables, the `extra_inputs` should be ignored and not passed downstream as part of the GQL operation.

see: https://apollographql.atlassian.net/wiki/spaces/graphdev/pages/1958543516/4.+Action+Oriented+Apps+Declaring+Tools#Declaring-Tools

crates/apollo-mcp-server/src/operations/operation.rs
andrewmcgivery: Yea, we absolutely could. I had included the variable types as a "just incase we need this" but we absolutely do not need it right now so I'm happy to simplify this if we'd prefer!

crates/apollo-mcp-server/src/apps/execution.rs
DaleSeo: Hmm... I think there might be a misunderstanding! I was referring to `operation.tool.input_schema`, not `AppTool.tool.input_schema` here. In `filter_inputs_for_operation`, we receive `&tool.operation` (the Operation), so `operation.tool.input_schema.get("properties")` should give us just the GraphQL variables, not the extra_inputs.

Does that make sense, or am I missing something?

PR #526, opened 2025-12-09
docs: Reorganize CORS documentation

jcaudle:
This change reorganizes some of the guidance in the CORS documentation to provide a more guided approach to developers trying to implement Apollo MCP Server.

### DIFF
--- a/docs/source/cors.mdx
+++ b/docs/source/cors.mdx
@@ -37,7 +37,7 @@ cors:
 
 If you set `allow_any_origin` to `true` (the default is `false`), Apollo MCP Server will respond with `Access-Control-Allow-Origin: *` in all CORS responses. This means that _any_ website can initiate browser connections. 
 
-<Note>It is important to note that your client will not be able to authenticate to your MCP Server if you use this configuration.</Note>
+<Note>Clients will not be able to authenticate with your MCP Server if you set `allow_any_origin` to `true`.</Note>
 
 ### Origins
 


docs/source/cors.mdx
```diff
@@ -1,68 +1,71 @@
...
   enabled: true
+```
 
-  # Set to true to allow any origin
-  # (Defaults to false)
+If `enabled` is not set to `true`, all other CORS options are ignored.
+
+### Configuring Allowed Origins
+
+You must specify which browser origins are allowed to access your MCP Server.
```
mabuyo, line 28: Can we add one more sentence after this just introducing the subheadings below? To signal that there's more than one way to do so.

Wondering if we can also match headings to look like this:

- Why it matters
- Enabling CORS
- Configuring allowed origins
    - Allow any origin
    - Specific origins
    - Match origins
- Passing credentials
- All cors options
- Common configurations
    - Development setup
    - Production setup
    - Public API setup 
- Browser integration example

So going up to an H2 for Configuring allowed origins and H3 for subheadings so they show up in the ToC. (I slightly tweaked the heading titles too) What do you think?

docs/source/cors.mdx
```diff
@@ -1,68 +1,71 @@
...
+  enabled: true
   allow_any_origin: true
+```
 
-  # List of accepted origins
-  # (Ignored if allow_any_origin is true)
-  #
-  # An origin is a combination of scheme, hostname and port.
-  # It does not have any path section, so no trailing slash.
+If you set `allow_any_origin` to `true` (the default is `false`), Apollo MCP Server will respond with `Access-Control-Allow-Origin: *` in all CORS responses. This means that _any_ website can initiate browser connections. It is important to note that your client will not be able to authenticate to your MCP Server if you use this configuration.
```
mabuyo: Could we pull out the "it is important to note" to a `<Note>` component?

docs/source/cors.mdx
```diff
@@ -1,68 +1,71 @@
...
     - https://www.your-app.example.com
-
-  # List of origin patterns (regex matching)
-  match_origins:
-    - "^https://([a-z0-9]+[.])*api[.]example[.]com$" # any host that uses https and ends with .api.example.com
+    - https://staging.your-app.example.com
 ```
 
-You can also disable CORS entirely by setting `enabled` to `false` or omitting the `cors` section:
+Setting `origins` to any list of URLs restricts access to only those specific servers. If you have a known list of web applications accessing your MCP server, you should use this option. An origin is a combination of protocol, hostname, and port. It does not have a path section, so there should be no trailing slash.
```
mabuyo:
```suggestion
Set `origins` to a list of URLs to restrict access to specific servers. Use this option if you have a known list of web applications accessing your MCP server. An origin is a combination of protocol, hostname, and port. Origins do not include a path, so do not include a trailing slash.
```

docs/source/cors.mdx
```diff
@@ -1,68 +1,73 @@
...
+```
 
-  # List of accepted origins
-  # (Ignored if allow_any_origin is true)
-  #
-  # An origin is a combination of scheme, hostname and port.
-  # It does not have any path section, so no trailing slash.
+If you set `allow_any_origin` to `true` (the default is `false`), Apollo MCP Server will respond with `Access-Control-Allow-Origin: *` in all CORS responses. This means that _any_ website can initiate browser connections. 
+
+<Note>It is important to note that your client will not be able to authenticate to your MCP Server if you use this configuration.</Note>
```
mabuyo:
```suggestion
<Note>Clients will not be able to authenticate with your MCP Server if you set `allow_any_origin` to `true`.</Note>
```

PR #528, opened 2025-12-11
App Tool Filtering

andrewmcgivery:
- When "app" query param is provided, only expose the tools appropriate for that app
- Refactor list_tools to have a list_tools_impl method to make testing easier

The purpose of this change is to make it easier to expose multiple apps from the same MCP server but avoid having multiple apps tools exposed at the same time, confusing the LLM.

I chose a url param (`/mcp?app=the-store`) because it's the only universal thing all clients/platforms have, compared to a header which isn't universally supported (including ChatGPT NOT supporting it).

The result is that the app tools are _only_ exposed when the url param is provided and _only_ the tools for that app are provided (+ the execute tool).

## Out of scope
I want to do the same thing for resources but I'll do that as a separate PR as to not blow this one up too big. :)

### DIFF
--- a/crates/apollo-mcp-server/src/server/states/running.rs
+++ b/crates/apollo-mcp-server/src/server/states/running.rs
@@ -885,6 +885,7 @@ mod tests {
         let result = running.list_tools_impl(extensions).await.unwrap();
 
         assert_eq!(result.tools.len(), 1);
+        assert_eq!(result.tools[0].name, "GetId");
         assert_eq!(result.next_cursor, None);
     }
 


crates/apollo-mcp-server/src/server/states/running.rs
```diff
@@ -803,4 +859,49 @@ mod tests {
...
+        let request = axum::http::Request::builder()
+            .uri("http://localhost?app=MyApp")
+            .body(())
+            .unwrap();
+        let (parts, _) = request.into_parts();
+        extensions.insert(parts);
+
+        let result = running.list_tools_impl(extensions).await.unwrap();
+
+        assert_eq!(result.tools.len(), 1);
```
DaleSeo, line 887: Consider also asserting the tool name to ensure the correct tool is returned.

```suggestion
        assert_eq!(result.tools.len(), 1);
        assert_eq!(result.tools[0].name, "GetId");
```

crates/apollo-mcp-server/src/server/states/running.rs
```diff
@@ -412,34 +493,9 @@ impl ServerHandler for Running {
...
-                .chain(self.explorer_tool.as_ref().iter().map(|e| e.tool.clone()))
-                .chain(self.validate_tool.as_ref().iter().map(|e| e.tool.clone()))
-                .chain(
-                    self.apps
-                        .iter()
-                        .flat_map(|app| app.tools.iter().map(|tool| tool.tool.clone())),
-                )
-                .collect(),
-        })
+        self.list_tools_impl(context.extensions).await
```
DaleSeo, line 498: > Refactor list_tools to have a list_tools_impl method to make testing easier

What was the specific difficulty with testing `list_tools` directly? Was it constructing RequestContext<RoleServer> or something else?

I'm a bit hesitant about this `xxx` -> `xxx_impl` approach becoming a pattern where `list_tools` becomes a one-liner delegating to `list_tools_impl`. The `_impl` suffix doesn't represent a meaningful abstraction and it's purely a testing artifact. We're testing `list_tools_impl`, not `list_tools`. If the delegation breaks, tests will still pass.

PR #529, opened 2025-12-15
deps: update rust to v1.92.0

DaleSeo:
Just updating the Rust version to match Rover.


DaleSeo:
Giving up due to the Nix issue for now.

PR #535, opened 2026-01-05
Allow opting out of audience validation

### LINKED ISSUE #438: Allow empty audiences
**Expected:** an empty `audiences` array will not attempt any audience validation

**Actual:** Passing in an empty `audiences` array to config results in failed validation at https://github.com/apollographql/apollo-mcp-server/blob/cd72dc1777b82118423f0fe757671c6c408c12d6/crates/apollo-mcp-server/src/auth/valid_token.rs#L114

DaleSeo:
Fixes #438

## Problem

Previously, passing an empty `audiences` array in the auth config would cause JWT validation to fail. Users expected that an empty array would skip audience validation entirely, but the `jsonwebtoken` library's default behavior (`validate_aud = true`) caused tokens to be rejected.

## Solution

Added an explicit `allow_any_audience` configuration option that follows the same pattern as CORS's `allow_any_origin`. When set to `true`, audience validation is skipped entirely.

```yaml
auth:
  servers:
    - https://auth.example.com
    
  # Validate specific audiences (default)
  audiences: ["my-api"]
  allow_any_audience: false

  # Or skip audience validation entirely
  audiences: []
  allow_any_audience: true## Changes
```

### DIFF
--- a/crates/apollo-mcp-server/src/auth.rs
+++ b/crates/apollo-mcp-server/src/auth.rs
@@ -15,6 +15,7 @@ use networked_token_validator::NetworkedTokenValidator;
 use schemars::JsonSchema;
 use serde::Deserialize;
 use tower_http::cors::{Any, CorsLayer};
+use tracing::warn;
 use url::Url;
 
 mod networked_token_validator;
@@ -34,6 +35,7 @@ pub struct Config {
     pub servers: Vec<Url>,
 
     /// List of accepted audiences for the OAuth tokens
+    #[serde(default)]
     pub audiences: Vec<String>,
 
     /// Allow any audience (skip validation) - use with caution
@@ -58,6 +60,12 @@ pub struct Config {
 
 impl Config {
     pub fn enable_middleware(&self, router: Router) -> Router {
+        if self.allow_any_audience {
+            warn!(
+                "allow_any_audience is enabled - audience validation is disabled. This reduces security."
+            );
+        }
+
         /// Simple handler to encode our config into the desired OAuth 2.1 protected
         /// resource format
         async fn protected_resource(State(auth_config): State<Config>) -> Json<ProtectedResource> {
--- a/docs/source/config-file.mdx
+++ b/docs/source/config-file.mdx
@@ -245,7 +245,7 @@ These fields are under the top-level `transport` key, nested under the `auth` ke
 | Option                           | Type           | Default | Description                                                                                        |
 | :------------------------------- | :------------- | :------ | :------------------------------------------------------------------------------------------------- |
 | `servers`                        | `List<URL>`    |         | List of upstream delegated OAuth servers (must support OIDC metadata discovery endpoint)           |
-| `audiences`                      | `List<string>` |         | List of accepted audiences from upstream signed JWTs (ignored if `allow_any_audience` is `true`)   |
+| `audiences`                      | `List<string>` | `[]`    | List of accepted audiences from upstream signed JWTs (ignored if `allow_any_audience` is `true`)   |
 | `allow_any_audience`             | `bool`         | `false` | Set to `true` to skip audience validation entirely (use with caution)                              |
 | `resource`                       | `string`       |         | The externally available URL pointing to this MCP server. Can be `localhost` when testing locally. |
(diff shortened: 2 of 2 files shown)


crates/apollo-mcp-server/src/auth/networked_token_validator.rs
```diff
@@ -32,7 +34,11 @@ fn build_oidc_url(oauth_server: &Url) -> Url {
 }
 
 impl ValidateToken for NetworkedTokenValidator<'_> {
-    fn get_audiences(&self) -> &Vec<String> {
+    fn allow_any_audience(&self) -> bool {
```
gocamille, line 37: @DaleSeo if this is set to `true` would we want a warning to be printed at runtime?

crates/apollo-mcp-server/src/auth/valid_token.rs
```diff
@@ -98,7 +101,11 @@ pub(super) trait ValidateToken {
                         continue;
                     }
                 });
-                val.set_audience(self.get_audiences());
+                if self.allow_any_audience() {
+                    val.validate_aud = false;
+                } else {
```
gocamille, line 106: @DaleSeo With this conditional, users will implicitly need two things if they want `audiences` to be empty:  they'll need the empty `audiences` array, and the `allow_any_audience` flag to be `true`.  Without both, the `audiences` array as I understand it will still fail the `jsonwebtoken` library `aud` validation. Should we also make the `audiences` field implicitly skip validation and have `audiences: []` work without `allow_any_audience`? Or is this considered a security trade-off?

crates/apollo-mcp-server/src/auth/valid_token.rs
```diff
@@ -98,7 +101,11 @@ pub(super) trait ValidateToken {
...
+                    val.set_audience(self.get_audiences());
```
gocamille, line 107: Should we also validate that `audiences` and `allow_any_audience` aren't in conflict (i.e. that someone can have `audiences: ["my-audience"]` and `allow_any_audience: true`)?

crates/apollo-mcp-server/src/auth.rs
```diff
@@ -36,6 +36,10 @@ pub struct Config {
     /// List of accepted audiences for the OAuth tokens
     pub audiences: Vec<String>,
```
gocamille, line 39: Wondering if we should add a `#[serde(default)]` here: 

```suggestion
    #[serde(default)]
    pub audiences: Vec<String>,
```

Which would allow users to omit the `audiences` field when `allow_any_audience` is `true`

crates/apollo-mcp-server/src/auth/networked_token_validator.rs
DaleSeo: Good idea! Will add a security warning.

crates/apollo-mcp-server/src/auth/valid_token.rs
DaleSeo: When `allow_any_audience: true`, the `audiences` option is completely ignored. Users don't need to set `audiences: []` as well.

I initially used an empty list to skip validation in my earlier PR #461. However, after @andrewmcgivery suggested following the same pattern as CORS, I introduced `allow_any_audience`. This way, users must explicitly acknowledge that they're disabling validation. It also aligns with how CORS is configured, where `allow_any_origin` is separate from `origins`.

I lean toward the current approach since it's more explicit about security trade-offs but let me know what you think! :)

crates/apollo-mcp-server/src/auth/valid_token.rs
DaleSeo: I followed our CORS implementation for consistency. It also ignores `origins` when `allow_any_origin: true`. Since we already log a warning at startup when `allow_any_audience` is enabled, I think that's enough to alert users. But I'm happy to add extra validation if you think it's worth it!

crates/apollo-mcp-server/src/auth.rs
DaleSeo: Love this suggestion! Now users can simply write:

```yaml
auth:
  servers:
    - https://auth.example.com
  allow_any_audience: true
```

Much cleaner. 😍 Thanks!

PR #537, opened 2026-01-06
feat(auth): Add 403 Forbidden `insufficient_scope` support per MCP Auth Spec 2025-11-25 and RFC 6750 (Section 3.1)

gocamille:
## Summary
This adds HTTP 403 Forbidden responses with `error="insufficient_scope"` per [MCP Auth Spec 2025-11-25 Section 10: Error Handling](https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#error-handling) and [RFC 6750 Section 3.1](https://www.rfc-editor.org/rfc/rfc6750.html#section-3.1).

## Changes
- `www_authenticate.rs`: Added `BearerError::InsufficientScope` enum and [error](cci:1://file:///Users/camillelawrence/Desktop/repos/apollo-mcp-server/crates/apollo-mcp-server/src/auth/www_authenticate.rs:135:4-149:5) field to `WWW-Authenticate` header
- `valid_token.rs`: Extract [scope](cci:1://file:///Users/camillelawrence/Desktop/repos/apollo-mcp-server/crates/apollo-mcp-server/src/auth/valid_token.rs:27:0-38:1)/[scp](cci:1://file:///Users/camillelawrence/Desktop/repos/apollo-mcp-server/crates/apollo-mcp-server/src/auth/valid_token.rs:503:8-509:9) claims from JWTs (handles both standard OAuth and Azure AD)
- `auth.rs`: Scope validation with fail-closed behaviour—valid tokens lacking required scopes get `403`
- `headers.rs`: Updated tests for new `ValidToken` struct

## Behavior
- **401 Unauthorized**: Missing or invalid token
- **403 Forbidden**: Valid token but insufficient scopes (includes `error="insufficient_scope"` in response)

### DIFF
--- a/.changeset/feat_add_insufficient_scope.md
+++ b/.changeset/feat_add_insufficient_scope.md
@@ -3,8 +3,8 @@
 This adds HTTP 403 Forbidden responses with `error="insufficient_scope"` per [MCP Auth Spec 2025-11-25 Section 10: Error Handling](https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#error-handling) and [RFC 6750 Section 3.1](https://www.rfc-editor.org/rfc/rfc6750.html#section-3.1).
 
 **Changes:**
-- `www_authenticate.rs`: Added `BearerError::InsufficientScope` enum and [error](cci:1://file:///Users/camillelawrence/Desktop/repos/apollo-mcp-server/crates/apollo-mcp-server/src/auth/www_authenticate.rs:135:4-149:5) field to `WWW-Authenticate` header
-- `valid_token.rs`: Extract [scope](cci:1://file:///Users/camillelawrence/Desktop/repos/apollo-mcp-server/crates/apollo-mcp-server/src/auth/valid_token.rs:27:0-38:1)/[scp](cci:1://file:///Users/camillelawrence/Desktop/repos/apollo-mcp-server/crates/apollo-mcp-server/src/auth/valid_token.rs:503:8-509:9) claims from JWTs (handles both standard OAuth and Azure AD)
+- `www_authenticate.rs`: Added `BearerError::InsufficientScope` enum and [error](crates/apollo-mcp-server/src/auth/www_authenticate.rs#L135-L149) field to `WWW-Authenticate` header
+- `valid_token.rs`: Extract [scope](crates/apollo-mcp-server/src/auth/valid_token.rs#L27-L38)/[scp](crates/apollo-mcp-server/src/auth/valid_token.rs#L503-L509) claims from JWTs (handles both standard OAuth and Azure AD)
 - `auth.rs`: Scope validation with fail-closed behaviour—valid tokens lacking required scopes get `403`
 - `headers.rs`: Updated tests for new `ValidToken` struct
 


crates/apollo-mcp-server/src/auth.rs
```diff
@@ -229,6 +246,27 @@ async fn oauth_validate(
         unauthorized_error()
     })?;
 
+    // Check if token has required scopes (fail-closed: missing scope claim = insufficient)
+    if !auth_config.scopes.is_empty() {
```
DaleSeo, line 250: Should we assume that if users set the `transport.auth.scopes` option to an empty list, they want to skip scope validation completely?

crates/apollo-mcp-server/src/auth.rs
```diff
@@ -229,6 +246,27 @@ async fn oauth_validate(
...
+        let missing_scopes: Vec<_> = auth_config
+            .scopes
+            .iter()
+            .filter(|required| !valid_token.scopes.iter().any(|s| s == *required))
```
DaleSeo, line 254: https://github.com/apollographql/apollo-mcp-server/blob/f3e8210a938bfd293b7b9f3d5d23a3fac12f3d88/crates/apollo-mcp-server/src/auth/protected_resource.rs#L33

The code above shows that the `transport.auth.scopes` option sets the `scopes_supported` field for the protected resource metadata endpoint. However, it seems like we're treating them as if they’re like scopes_required here though.

According to the MCP spec:

> The scopes included in the `WWW-Authenticate` challenge MAY match `scopes_supported`, be a subset or superset of it, or an alternative collection that is neither a strict subset nor superset.

I really dislike the ambiguity in the spec. 🤪🔫 I'm having trouble interpreting this, but I’d love to hear your thoughts on it.

crates/apollo-mcp-server/src/auth/valid_token.rs
```diff
@@ -51,6 +67,15 @@ pub(super) trait ValidateToken {
...
             pub sub: String,
+
+            /// OAuth scope claim (standard OAuth providers like Auth0, Okta)
+            /// Space-separated list of scopes
+            #[serde(default)]
+            pub scope: Option<String>,
+
+            /// Microsoft/Azure AD uses "scp" instead of "scope"
+            #[serde(default)]
+            pub scp: Option<String>,
```
DaleSeo: Interesting... I guess this isn't mentioned in the spec, but it's something you discovered? How about we wait for people to report this issue before we increase the complexity? From my experience, dealing with all these non-standard edge cases can be pretty exhausting, adding to complexity.

crates/apollo-mcp-server/src/auth.rs
```diff
@@ -229,6 +246,27 @@ async fn oauth_validate(
...
+                required = ?auth_config.scopes,
+                present = ?valid_token.scopes,
+                missing = ?missing_scopes,
+                "Token has insufficient scopes"
+            );
+            tracing::Span::current().record("reason", "insufficient_scope");
+            tracing::Span::current().record("status_code", StatusCode::FORBIDDEN.as_u16());
+            return Err(forbidden_error(&auth_config.scopes));
+        }
+    }
```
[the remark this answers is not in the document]
DaleSeo: This looks like one of those rare times when AI gives a good review. 😃

.changeset/feat_add_insufficient_scope.md
```diff
@@ -0,0 +1,16 @@
+### feat(auth): Add 403 Forbidden `insufficient_scope` support per MCP Auth Spec 2025-11-25 and RFC 6750 (Section 3.1) - @gocamille PR #537
+
+<!-- https://apollographql.atlassian.net/browse/AMS-171 -->
```
DaleSeo: I guess this is not intentional?

.changeset/feat_add_insufficient_scope.md
```diff
@@ -0,0 +1,16 @@
...
+
+## Summary
```
DaleSeo: Not sure how these H2 headings under H3 will look in `CHANGELOG.md` when we cut a release. 🤔

.changeset/feat_add_insufficient_scope.md
gocamille: OOOoooh! This must be because it was in the PR before I ran the `cargo xtask` on this, thank you for your eagle eye on this!

crates/apollo-mcp-server/src/auth.rs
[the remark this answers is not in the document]
gocamille: Implemented test coverage in [3188f91](https://github.com/apollographql/apollo-mcp-server/pull/537/commits/3188f914be7511127559e2a4fe2fc8d99377364f)

crates/apollo-mcp-server/src/auth.rs
gocamille: This (and the one below) is a really great question (and super nuanced @DaleSeo !) -- looking at the specs (both MCP and OAuth), this assumption is correct. So when `transport.auth.scopes` is set to an empty list the following would happen:
- Scope validation would be completely skipped
- The token would only need to pass signature and audience validation
- The `scopes_supported` field in the protected resource metadata would be empty (which a client could see as "no specific scopes needed")

This seems a reasonable default for those who want simple "authenticated or not" access control, but (and this is related to the [comment below](https://github.com/apollographql/apollo-mcp-server/pull/537#discussion_r2669998054) too) we may want to consider adding flexibility / complexity in a few ways: 

- Separating `scopes_supported` from `scopes_required` — like you mentioned the MCP spec explicitly says scopes in `WWW-Authenticate` "may" match `scopes_supported`, be a subset or superset of it, or an alternative collection, while right now we use one list for enforcement.  
- Supporting per-tool scope requirements (a big one) — The spec mentions "minimum scopes needed for the operation," so different MCP tools might require different scopes.
- Adding "any-of" logic may also be good here, as the current implementation requires all listed scopes while some use cases might only need at least one from a set.

What do you think of the current "all-or-nothing" approach? Should we enhance the flexibility of this to cover any of the above items? 

(If we do stay with the current approach I would update the docs to document this clearly, and possibly add a warning when the scopes list is empty)

crates/apollo-mcp-server/src/auth.rs
gocamille: Yes! This ambiguity had me in tangles 😂 ! Re: `scopes_supported` vs `scopes_required` — you're right that we're currently treating them as the same thing, I was just wondering whether they should be separated conceptually in the [above comment](https://github.com/apollographql/apollo-mcp-server/pull/537#discussion_r2676526762) too. I think using one list for both is acceptable, but a) we should document this clearly and b) revisit if users need more granular control. Especially since the spec's "MAY match, subset, superset, or alternative" leaves room for this flexibility.

crates/apollo-mcp-server/src/auth/valid_token.rs
gocamille: That's valid @DaleSeo ! This predates the `scope` standardization and it's me defensively programming from my past knowledge, but I agree it adds complexity.

.changeset/feat_add_insufficient_scope.md
gocamille: Also good catch -- I've updated both subheadings to be **bold**

crates/apollo-mcp-server/src/auth.rs
DaleSeo: Thanks for sharing your thoughts, @gocamille! This reminds me that in PR #535, we introduced the `allow_any_audience` flag instead of treating `audiences: []` as skipping validation. I think we should do the same for scopes, and it might be another case where a CORS-like configuration makes sense. Skipping scope validation reduces security, so I would suggest making it an explicit opt-out.

PR #538, opened 2026-01-06
Replace ci job with nix-less setup

andrewmcgivery:


.github/workflows/ci.yml
```diff
@@ -1,4 +1,4 @@
-name: Nix CI
+name: CI
```
DaleSeo, line 1: This already makes me want to hit approve, haha!

.github/workflows/ci.yml
```diff
@@ -10,135 +10,69 @@ on:
...
-      - name: Restore and save Nix store
-        uses: nix-community/cache-nix-action@v6
+      - uses: actions-rust-lang/setup-rust-toolchain@v1
+      - run: cargo doc
+
+  fmt:
+    name: Run Cargo Format
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v5
```
DaleSeo:
```suggestion
      - uses: actions/checkout@v6
```

.github/workflows/ci.yml
```diff
@@ -10,135 +10,69 @@ on:
...
       - uses: actions/checkout@v5
         with:
           ref: ${{ github.event.pull_request.head.sha }}
-      - uses: nixbuild/nix-quick-install-action@v30
-        with:
-          nix_conf: ${{ env.nix_conf }}
-      - name: Restore and save Nix store
-        uses: nix-community/cache-nix-action@v6
+      - uses: actions-rust-lang/setup-rust-toolchain@v1
+      - run: cargo doc
```
DaleSeo, line 30: We haven't done this before, have we? I'm curious why we should generate the API doc in CI.

.github/workflows/ci.yml
```diff
@@ -10,135 +10,69 @@ on:
...
         with:
-          primary-key: build-${{ runner.os }}-${{ hashFiles('Cargo.lock', '**/Cargo.toml', 'flake.nix', 'flake.lock', 'rust-toolchain.toml') }}
-          purge: true
-          purge-prefixes: build-${{ runner.os }}-
-          purge-created: 0
-          purge-primary-key: never
-          gc-max-store-size: 5G
-      - name: Run checks
-        run: nix flake check
+          ref: ${{ github.event.pull_request.head.sha }}
```
DaleSeo: I see the old Nix CI had the same checkout strategy, so this is preserving existing behavior. However, I wonder if we want to continue checking out the PR head commit directly.

By default, `actions/checkout` on a pull request checks out a synthetic merge commit, basically a preview of what main would look like after merging the PR. The explicit `ref` here opts out of that and checks out the exact commit the author pushed instead.

The tradeoff is that the default approach catches integration issues earlier since it tests against the current state of main, but it can also cause CI to fail unexpectedly when main changes even if the PR itself hasn't changed. The explicit SHA approach gives more predictable results tied to what the author actually pushed.

I'm inclined to use the default checkout strategy, which I believe is more conventional across other projects.  I remember we've seen issues in the main branch because they weren't caught before merging PRs.

.github/workflows/ci.yml
```diff
@@ -10,135 +10,69 @@ on:
...
-          purge-created: 0
-          purge-primary-key: never
-          gc-max-store-size: 5G
-      - name: Save flake attributes from garbage collection
-        run: nix profile install .#saveFromGC
+      - uses: actions-rust-lang/setup-rust-toolchain@v1
+      - run: cargo clippy --all-targets -- --deny warnings
 
-  check:
-    name: Run checks
```
DaleSeo, line 50: Let's make this status check optional in the branch protection settings. That way, we can merge this PR and avoid blocking others.

2026-01-07 at 10 22 47

.github/workflows/ci.yml
gocamille: I also am wondering -- would we be looking to publish this binary to crates.io in the future?

andrewmcgivery: This was already done by NIX. We can certainly remove if it we don't see the value... was just going for a 1:1 replacement as much as possible haha

.github/workflows/ci.yml
andrewmcgivery: @DaleSeo Is this something you can do? I don't have permission to do basically anything in settings lol

.github/workflows/ci.yml
DaleSeo: @andrewmcgivery Oh, I didn't know that. OK, then let's keep it for now. It shouldn't be a big deal since it runs at the same time as other jobs.

.github/workflows/ci.yml
DaleSeo: Sure, I’ve taken care of that! You guys should have permissions once we set up the new team on GitHub.

DaleSeo (approved):
Thanks @andrewmcgivery for addressing my feedback! I'm really excited about the simplified CI!

PR #539, opened 2026-01-07
feat: add support for custom environment variable expansion

### LINKED ISSUE #454: Allow Custom Environment Variable Mapping in Apollo MCP Server
The Apollo MCP Server currently enforces a strict environment variable naming convention:

```
APOLLO_MCP_{SECTION}__{SUBSECTION}__{PROPERTY}
```

This tightly couples environment variable names to the internal YAML configuration hierarchy. While this approach ensures structure and consistency, it creates friction for users deploying across diverse environments (e.g., Kubernetes, Docker, CI/CD systems, or enterprise infrastructure with custom naming policies).

**Problem:**

1. **Rigid Naming Pattern:**
   All environment variables must follow the `APOLLO_MCP_` prefix and double-underscore (`__`) nesting rule.
2. **Integration Overhead:**
   Many systems have predefined or shorter naming schemes that don’t align with this pattern, forcing users to manually rename or duplicate variables.
3. **Reduced Portability:**
   Teams migrating configurations or integrating multiple services face compatibility issues because variable names cannot be customized or remapped.
4. **No Alias Support:**
   The system does not allow mapping alternative environment variable names to existing configuration keys.

**Example Limitation:**
Currently, to configure the OTLP tracing endpoint, only this variable is supported:

```
APOLLO_MCP_TELEMETRY__EXPORTERS__TRACING__OTLP__ENDPOINT
```

It’s not possible to use a shorter or existing environment variable such as:

```
CUSTOM_OTEL_ENDPOINT
```

**Proposed Solution:**
Introduce an optional `env_mapping` section in the configuration file to allow flexible aliasing or remapping of environment variables.

Example proposal:

```yaml
telemetry:
  # Service name for telemetry data identification
  service_name: "rhg-apollo-mcp-server"
  # Service version for telemetry data
  version: "1.2.3"
  # Telemetry exporters configuration
  exporters:
    # Metrics export configuration
    metrics:
      # OTLP (OpenTelemetry Protocol) exporter for metrics
      otlp:
        # Protocol: "grpc" or "http/protobuf" (default: grpc)
        protocol: "grpc"
        endpoint: ${env.CUSTOM_OTEL_ENDPOINT}
```

The MCP server would resolve these custom mappings first, falling back to the default naming convention if not found.

**Benefits:**

* Enables flexible integration with any deployment environment
* Reduces configuration duplication and human error
* Maintains backward compatibility with existing naming patterns
* Simplifies adoption for teams already using standard environment variables

**Environment:**

* Apollo MCP Server Version: 1.0.0

**Summary:**
The current environment variable naming pattern (`APOLLO_MCP_{SECTION}__{SUBSECTION}__{PROPERTY}`) is too prescriptive for real-world use cases. Supporting a user-defined mapping system would make configuration more portable, maintainable, and developer-friendly.

gocamille:
## Summary

This PR adds support for `${env.VAR_NAME}` syntax in configuration files, allowing users to reference custom environment variables without being limited to the `APOLLO_MCP_*` naming convention.

Closes #454.

## Changes

- `runtime/env_expansion.rs` (new module) - parser for variable expansion
- `runtime.rs` (modified) - integrates expansion into the `read_config()` function
- `config-file.mdx` - updated docs with syntax, escaping, and special characters handling

- **Note** The `APOLLO_MCP_*` environment variable(s) will still take precedence over expanded custom config values (no breaking change).

### DIFF
--- a/.changeset/feat_env_var_expansion.md
+++ b/.changeset/feat_env_var_expansion.md
@@ -1,12 +1,12 @@
 ### feat: add support for custom environment variable expansion - @gocamille PR #539
 
-## Summary
+#### Summary
 
 This PR adds support for `${env.VAR_NAME}` syntax in configuration files, allowing users to reference custom environment variables without being limited to the `APOLLO_MCP_*` naming convention.
 
 Closes #454.
 
-## Changes
+#### Changes
 
 - `runtime/env_expansion.rs` (new module) - parser for variable expansion
 - `runtime.rs` (modified) - integrates expansion into the `read_config()` function


crates/apollo-mcp-server/src/runtime/env_expansion.rs
```diff
@@ -0,0 +1,266 @@
+//! Environment variable expansion for configuration files.
+//!
+//! Supports `${env.VAR_NAME}` syntax. Use `$${env.VAR}` to escape.
```
gocamille: Should we add `${env.VAR:-default}` syntax for fallback values in a follow-up?

crates/apollo-mcp-server/src/runtime.rs
```diff
@@ -297,4 +313,43 @@ mod test {
...
+
+            let config = read_config(path)?;
+
+            assert_eq!(config.endpoint.as_str(), "https://expanded:4000/");
+            Ok(())
+        });
+    }
+
+    #[test]
+    fn it_prioritizes_apollo_mcp_env_over_expanded_vars() {
```
DaleSeo, line 335: These tests are really helpful to understand the precedence behavior!

crates/apollo-mcp-server/src/env_expansion.rs
```diff
@@ -0,0 +1,275 @@
...
+//!
+//! Supports `${env.VAR_NAME}` syntax. Use `$${env.VAR}` to escape.
+//! See config-file.mdx for full documentation.
+
+use std::env;
+use std::iter::Peekable;
+use std::str::Chars;
+
+#[derive(Debug, PartialEq, thiserror::Error)]
+pub enum EnvExpansionError {
```
DaleSeo: Since only the parent module needs to access this

```suggestion
pub(super) enum EnvExpansionError {
```

crates/apollo-mcp-server/src/runtime/env_expansion.rs
```diff
@@ -0,0 +1,275 @@
...
+pub enum EnvExpansionError {
+    #[error("undefined environment variable '{name}' referenced in configuration")]
+    UndefinedVariable { name: String },
+
+    #[error("environment variable '{name}' contains non-UTF8 data")]
+    NonUnicodeValue { name: String },
+}
+
+/// Expand all `${env.VAR_NAME}` references in the given content.
+pub fn expand_env_vars(content: &str) -> Result<String, EnvExpansionError> {
```
DaleSeo:
```suggestion
pub(super) fn expand_env_vars(content: &str) -> Result<String, EnvExpansionError> {
```

crates/apollo-mcp-server/src/runtime/env_expansion.rs
```diff
@@ -0,0 +1,275 @@
...
+/// body and either expands it (if valid) or outputs it literally.
+fn expand_placeholder(
+    chars: &mut Peekable<Chars>,
+    result: &mut String,
+) -> Result<(), EnvExpansionError> {
+    chars.next(); // consume '{'
+    let body = read_placeholder_body(chars);
+
+    // Extract var name: must be "env.<name>}" format
+    let Some(var_name) = body.strip_prefix("env.").and_then(|s| s.strip_suffix('}')) else {
```
DaleSeo: So, `env.` has to be case sensitive.

crates/apollo-mcp-server/src/runtime/env_expansion.rs
DaleSeo: Apollo Router already supports that `${env.VAR:-default}` expansion (see [their docs](https://www.apollographql.com/docs/graphos/routing/configuration/yaml#variable-expansion)). Since the customer uses Apollo Router, it would be a good idea to add it. But if it's too much work, we can just wait for them to ask.

.changeset/feat_feat_env_var_expansion.md
DaleSeo, line 1: nit: I guess this double prefix `feat_` isn't intended?

.changeset/feat_feat_env_var_expansion.md
```diff
@@ -0,0 +1,15 @@
+### feat: add support for custom environment variable expansion ([Issue #454](https://github.com/apollographql/apollo-mcp-server/issues/454))
```
DaleSeo: We usually include the PR number in the changelog following this format:

```suggestion
### feat: add support for custom environment variable expansion - @gocamille PR #539
```

DaleSeo (commented):
Thanks @gocamille, for implementing this so cleanly with solid tests! I believe this should meet the customer's request. However, I wanted to mention that most features in Apollo MCP Server have been following Router's patterns since they've been refined and tested over several years. This approach also ensures a consistent experience across Apollo products.

As I was reviewing this, I took a look at how Apollo Router implements the same functionality: https://github.com/apollographql/router/blob/dev/apollo-router/src/configuration/expansion.rs

Here are a couple of differences worth noting:
-  Router performs type coercion during expansion. Environment variables that contain `true` or `10` can be converted to YAML booleans or numbers. The current implementation only does string substitution before YAML parsing, which is simpler but requires users to be careful about how values are interpreted.
-  Architecturally, Router's expansion module is part of the library and available to crate consumers, while this implementation is limited to the binary's runtime module. That works for now, but it's something to keep in mind if we expect other consumers to want this functionality.

This isn't blocking anything, just flagging it for consideration.

crates/apollo-mcp-server/src/runtime/env_expansion.rs
gocamille: Thank you @DaleSeo ! Fixed in commit https://github.com/apollographql/apollo-mcp-server/pull/539/commits/5ba738bbfde8bc9cf9b3fc60ace7c0bae1a4f0ce 👍

crates/apollo-mcp-server/src/env_expansion.rs
gocamille: Ah good catch thank you!

crates/apollo-mcp-server/src/runtime/env_expansion.rs
gocamille: This is fixed on commit 23a61180 !

crates/apollo-mcp-server/src/runtime/env_expansion.rs
gocamille: Ah good question @DaleSeo -- yes, in this initial implementation, `env.` must be case sensitive -- the updated implementation which achieves parity with apollo router by using the `shellexpand` library also follows the same convention 👍 since bash and most shells follow this convention I'm wondering if it's intuitive for users, or whether we should be explicit in the docs. What do you think?

.changeset/feat_feat_env_var_expansion.md
gocamille: Thank you yes ! Removed 👍 hopefully the use of `knope` will eliminate duplicate `changesets`.

gocamille:
Thank you so much for your thorough review @DaleSeo ! I appreciated the insight on achieving logical parity with our long-standing Router features, that makes sense! I did take that suggestion and updated the core `expand_env_vars` function to leverage the well-supported `shellexpand` library, which is the same library used by Router. As you pointed out, we now have parity on the range of features this expansion module covers: 

- The MCP server now performs type coercion during expansion with `shellexpand`, like apollo router ([a09f08c](https://github.com/apollographql/apollo-mcp-server/pull/539/commits/a09f08cce915a4267eefacd36e659398e8ce835b)); 
- Default variable expansion has also been implemented ([5ba738b](https://github.com/apollographql/apollo-mcp-server/pull/539/commits/5ba738bbfde8bc9cf9b3fc60ace7c0bae1a4f0ce)); and
- The expansion module here is now part of the library so it's available to crate consumers who want this functionality.

This PR is only targeting a very narrow band of features for environment variable expansion, but we could look into adding the following in the future to achieve parity with apollo-router (and consider other ways we'd want to extend it), including: 
  - File expansion
  - Prefix support
  - Expansion modes
  - An override system
  - Developer mode defaults

.changeset/feat_env_var_expansion.md
```diff
@@ -0,0 +1,15 @@
+### feat: add support for custom environment variable expansion - @gocamille PR #539
+
+## Summary
```
DaleSeo: Shouldn't we use `####` here instead? I'm concerned that this might break the structure when we add it to the changelog.

DaleSeo (approved):
Great improvements! 🎉

.changeset/feat_env_var_expansion.md
gocamille: Good point -- I'll update this!

PR #542, opened 2026-01-08
Execute tool visibility when running app

andrewmcgivery:
When running apps, make the execute tool executable from the app but hidden from the LLM via meta entry on the tool.

### DIFF
--- a/crates/apollo-mcp-server/src/apps/execution.rs
+++ b/crates/apollo-mcp-server/src/apps/execution.rs
@@ -431,8 +431,8 @@ mod tests {
         assert!(result.is_none());
     }
 
-    #[tokio::test]
-    async fn make_tool_private_adds_meta_when_tool_has_no_meta() {
+    #[test]
+    fn make_tool_private_adds_meta_when_tool_has_no_meta() {
         let mut tool = Tool::new("GetId", "a description", JsonObject::new());
         tool = make_tool_private(tool);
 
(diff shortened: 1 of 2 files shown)


crates/apollo-mcp-server/src/apps/execution.rs
```diff
@@ -423,4 +430,39 @@ mod tests {
 
         assert!(result.is_none());
     }
+
+    #[tokio::test]
+    async fn make_tool_private_adds_meta_when_tool_has_no_meta() {
```
DaleSeo: This test doesn't seem to await anything.

```suggestion
    #[test]
    fn make_tool_private_adds_meta_when_tool_has_no_meta() {
```

crates/apollo-mcp-server/src/apps/execution.rs
```diff
@@ -423,4 +430,39 @@ mod tests {
...
+
+        assert_eq!(meta.keys().len(), 1);
+        assert_eq!(
+            meta.get("openai/visibility").unwrap(),
+            &Value::from("private")
+        );
+    }
+
+    #[tokio::test]
+    async fn make_tool_private_modified_meta_when_tool_has_existing_meta() {
```
DaleSeo: ditto

```suggestion
    #[test]
    fn make_tool_private_modified_meta_when_tool_has_existing_meta() {
```

crates/apollo-mcp-server/src/server/states/running.rs
```diff
@@ -223,7 +223,12 @@ impl Running {
                             .await
                             .iter()
                             .map(|op| op.as_ref().clone())
-                            .chain(self.execute_tool.as_ref().iter().map(|e| e.tool.clone()))
+                            .chain(
+                                self.execute_tool
+                                    .as_ref()
+                                    .iter()
+                                    .map(|e| make_tool_private(e.tool.clone())),
```
DaleSeo, line 231: Can we add a comment explaining why the execute tool is private only in app mode? It's clear from the PR description, but a bit of context here would be helpful later when we look at the code.

PR #543, opened 2026-01-08
chore: remove direnv

DaleSeo:
`direnv` was only needed by Nix.


.vscode/extensions.json
```diff
@@ -1,8 +1,7 @@
 {
-    "recommendations": [
-        "mkhl.direnv",
```
DaleSeo, line 3: This has been causing this on VS Code-based IDEs, which I've always closed.

Image

andrewmcgivery (approved):

PR #555, opened 2026-01-14
Update rmcp to 0.12 and remove SSE transport support

DaleSeo:
This PR updates the `rmcp` dependency from v0.9 to v0.12. This upgrade aligns the Apollo MCP Server with the latest improvements in the MCP SDK and lays the groundwork for our ongoing spec compliance efforts.

As part of this update, SSE transport support has been removed. The MCP specification version [2025-03-26](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports) replaced the HTTP+SSE transport with the more flexible Streamable HTTP transport, and rmcp followed suit by removing the `SseServer` implementation in v0.11. SSE was already marked as deprecated in our codebase in favor of Streamable HTTP, and new features have not been supported for SSE for the past few months.

Users currently configured with SSE transport will receive a clear error message directing them to migrate to `streamable_http` transport, which provides the same HTTP-based communication with improved session handling capabilities.

The update also required adding the new `meta` field (as `Option<Meta>`) to several MCP model types including `ListToolsResult`, `ListResourcesResult`, and `RawResource`, reflecting changes in the update [schema reference](https://modelcontextprotocol.io/specification/2025-11-25/schema).

## Breaking Change

The `sse` transport type is no longer supported. Users must update their configuration to use `streamable_http` instead:

- Before

```yaml
transport:
  type: sse
  address: 127.0.0.1
  port: 8000
```

- After

```yaml
transport:
  type: streamable_http
  address: 127.0.0.1
  port: 8000
```

### DIFF
--- a/.changeset/remove_sse_transport.md
+++ b/.changeset/remove_sse_transport.md
@@ -1,0 +1,29 @@
+---
+default: minor
+---
+
+### Remove SSE transport support - @DaleSeo PR #555
+
+The SSE transport has been removed following the upgrade to rmcp 0.12. SSE was previously deprecated in favor of Streamable HTTP transport.
+
+**Migration**: Update your configuration to use `streamable_http` transport instead of `sse`:
+
+- Before
+
+```yaml
+transport:
+  type: sse
+  address: 127.0.0.1
+  port: 8000
+```
+
+- After
+
+```yaml
+transport:
+  type: streamable_http
+  address: 127.0.0.1
+  port: 8000
+```
+
+If you were using SSE transport, switch to `streamable_http` which provides the same HTTP-based communication with improved session handling and is the recommended transport in the MCP specification.
--- a/crates/apollo-mcp-server/src/apps/execution.rs
+++ b/crates/apollo-mcp-server/src/apps/execution.rs
@@ -774,6 +774,7 @@ mod tests {
                 description: None,
                 icons: None,
                 size: None,
+                meta: None,
             },
             None,
         );
@@ -806,6 +807,7 @@ mod tests {
                 description: None,
                 icons: None,
                 size: None,
+                meta: None,
             },
             None,
         );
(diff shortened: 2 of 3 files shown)


.changeset/remove_sse_transport.md
```diff
@@ -0,0 +1,29 @@
+---
+default: major
```
andrewmcgivery: I think this is appropriate as major and flagged as breaking.

Are we okay with doing a 2.x.x bump as a result of this? Do we want to try to line up any other breaking changes for the next release to go with this?

crates/apollo-mcp-server/src/server/states/running.rs
```diff
@@ -263,13 +264,15 @@ impl Running {
...
                 .collect(),
+            meta: None,
         })
     }
 
     fn list_resources_impl(&self) -> ListResourcesResult {
         ListResourcesResult {
             resources: self.apps.iter().map(|app| app.resource()).collect(),
             next_cursor: None,
+            meta: None,
```
andrewmcgivery, line 283: I assume `meta` ends up being mapped to `_meta` by the library?

.changeset/remove_sse_transport.md
DaleSeo: Great catch! We haven't strictly followed the semantic versioning for the major version. The v1.x.x bump was mainly a product and marketing decision, and we've had a couple of breaking changes since then. I'll change this to `default: minor` so that Knope bumps the minor version. When we release, we can manually update the changelog to include a breaking section like we've done so far:

2026-01-16 at 08 43 50

crates/apollo-mcp-server/src/server/states/running.rs
DaleSeo: That is right: https://github.com/DaleSeo/rust-sdk/blob/95d3e3f9403ea853f18ef692143cecf8bc156b20/crates/rmcp/src/model.rs#L885-L886

PR #557, opened 2026-01-14
Release workflow fixes

andrewmcgivery:
- Use Knope to manage version in server.json file
- Don't replace -rc of versions in docs
- Remove "new" from job names
- Fix verify changeset to skip release PRs


gocamille (approved):
Thank you @andrewmcgivery !!

PR #558, opened 2026-01-15
Refactor apps code to centralize meta data

andrewmcgivery:
Refactor apps code to centralize meta data so that it is set in a single place at run time. This is to make it easier to add MCP Apps support later and allow us to switch between the two on-the-fly based on our delivery target.

### DIFF
--- a/crates/apollo-mcp-server/src/apps/execution.rs
+++ b/crates/apollo-mcp-server/src/apps/execution.rs
@@ -146,8 +146,9 @@ pub(crate) fn make_tool_private(mut tool: Tool) -> Tool {
 }
 
 // Attach tool meta data when requested to allow swapping between app targets (Apps SDK, MCP Apps)
-pub(crate) fn attach_tool_metadata(app: &App, mut tool: AppTool) -> AppTool {
-    let meta = tool.tool.meta.get_or_insert_with(Meta::new);
+pub(crate) fn attach_tool_metadata(app: &App, tool: &AppTool) -> Tool {
+    let mut inner_tool = tool.tool.clone();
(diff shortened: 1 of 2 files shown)


crates/apollo-mcp-server/src/apps/execution.rs
```diff
@@ -143,13 +145,131 @@ pub(crate) fn make_tool_private(mut tool: Tool) -> Tool {
     tool
 }
 
+// Attach tool meta data when requested to allow swapping between app targets (Apps SDK, MCP Apps)
+pub(crate) fn attach_tool_metadata(app: &App, mut tool: AppTool) -> AppTool {
```
DaleSeo: This clones the entire `AppTool` (including `Arc<Operation>` + `AppLabels`), but it looks like we only need to clone the `Tool` at the end. How about we refactor this function to take `&AppTool` instead and return `Tool` instead of `AppTool`?

Then the call site becomes cleaner like:

```rs
.map(|tool| attach_tool_metadata(app, tool))
````

andrewmcgivery: I tried doing this but it ends up being an issue because then you have to pass `tool: &mut AppTool` which then requires me to do `app.tools.iter_mut()` in `running.rs` which then results in... well, eventually we get back to not being able to make `self` mutable. 😅

However, I took a slightly different approach with a similar result... I passed in &AppTool and immediately cloned the inner tool, made the meta changes on that, and then returned that inner tool. Should be a less expensive clone!

DaleSeo: Looks great! Thanks for refactoring it.

PR #561, opened 2026-01-15
Add exponential retry backoffs during initial collection fetch

gocamille:
Adds a streamlined retry mechanism (10 attempts, 5s delay) for transient errors during initial collection fetch and allows the server to continue serving cached operations when subsequent polls encounter transient network failures, preventing unnecessary crash loops.

### Testing

New tests added to verify transient error classification and initial fetch retry logic:

```
test platform_api::operation_collections::collection_poller::tests::test_is_collection_error_transient_response_error 
test platform_api::operation_collections::collection_poller::tests::test_is_collection_error_transient_header_errors 
test platform_api::operation_collections::collection_poller::tests::test_is_collection_error_transient_invalid_variables 
test platform_api::operation_collections::collection_poller::tests::test_retry_initial_fetch_success_first_try 
test platform_api::operation_collections::collection_poller::tests::test_retry_initial_fetch_permanent_error_no_retry
```

### DIFF
--- a/.changeset/add_exponential_retry_backoffs_during_initial_collection_fetch.md
+++ b/.changeset/add_exponential_retry_backoffs_during_initial_collection_fetch.md
@@ -1,0 +1,5 @@
+---
+default: patch
+---
+
+# Add exponential retry backoffs during initial collection fetch
--- a/.changeset/add_exponential_retry_backoffs_during_initial_collection_fetch.md
+++ b/.changeset/add_exponential_retry_backoffs_during_initial_collection_fetch.md
@@ -1,5 +1,0 @@
----
-default: patch
----
-
-# Add exponential retry backoffs during initial collection fetch
--- a/crates/apollo-mcp-registry/src/platform_api/operation_collections/collection_poller.rs
+++ b/crates/apollo-mcp-registry/src/platform_api/operation_collections/collection_poller.rs
@@ -174,6 +174,16 @@ async fn backoff_sleep_with_shutdown(
     }
 }
 
+/// Creates the standard retry backoff configuration for initial collection fetches.
+fn create_retry_backoff() -> backoff::ExponentialBackoff {
+    backoff::ExponentialBackoff {
+        initial_interval: INITIAL_BACKOFF,
+        max_interval: MAX_BACKOFF,
+        max_elapsed_time: Some(MAX_ELAPSED_TIME),
+        ..Default::default()
+    }
+}
+
 #[derive(Clone)]
 pub struct OperationData {
     id: String,
@@ -328,15 +338,8 @@ impl CollectionSource {
             let mut previous_updated_at = HashMap::new();
             let mut has_fetched = false;
 
-            // Note: This retry loop pattern is shared with default_collection_stream.
-            // Shutdown sleep handling is deduplicated via backoff_sleep_with_shutdown helper.
             // Retry loop for initial fetch with exponential backoff
-            let mut backoff = backoff::ExponentialBackoff {
-                initial_interval: INITIAL_BACKOFF,
-                max_interval: MAX_BACKOFF,
-                max_elapsed_time: Some(MAX_ELAPSED_TIME),
-                ..Default::default()
-            };
+            let mut backoff = create_retry_backoff();
 
             loop {
                 // Check for shutdown before each attempt
@@ -497,12 +500,7 @@ impl CollectionSource {
             let mut has_fetched = false;
 
             // Retry loop for initial fetch with exponential backoff
-            let mut backoff = backoff::ExponentialBackoff {
-                initial_interval: INITIAL_BACKOFF,
-                max_interval: MAX_BACKOFF,
-                max_elapsed_time: Some(MAX_ELAPSED_TIME),
-                ..Default::default()
-            };
+            let mut backoff = create_retry_backoff();
 
             loop {
                 // Check for shutdown before each attempt
--- a/crates/apollo-mcp-registry/src/platform_api/operation_collections/collection_poller.rs
+++ b/crates/apollo-mcp-registry/src/platform_api/operation_collections/collection_poller.rs
@@ -1,9 +1,11 @@
+use backoff::backoff::Backoff;
 use futures::Stream;
 use graphql_client::GraphQLQuery;
 use reqwest::header::{HeaderMap, HeaderName, HeaderValue};
 use secrecy::ExposeSecret;
 use std::collections::HashMap;
 use std::pin::Pin;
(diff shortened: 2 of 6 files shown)


crates/apollo-mcp-registry/src/platform_api/operation_collections/collection_poller.rs
```diff
@@ -34,6 +36,9 @@ use operation_collection_query::{
 };
 
 const MAX_COLLECTION_SIZE_FOR_POLLING: usize = 100;
```
gocamille, line 37: Note that these new values are used for initial fetches and not for polling and used by the `create_retry_backoff` helper function below.

crates/apollo-mcp-registry/src/platform_api/operation_collections/collection_poller.rs
```diff
@@ -152,6 +157,30 @@ fn is_collection_error_transient(error: &CollectionError) -> bool {
...
+) -> bool {
+    tokio::select! {
+        _ = tokio::time::sleep(duration) => true,
+        _ = sender.closed() => {
+            tracing::debug!("Shutdown detected during backoff sleep");
+            false
+        }
+    }
+}
+
```
gocamille, line 199: This tracks how long we've been retrying and calculates increasing wait times with the `backoff` crate, so that each subsequent call to `next_backoff()` returns the next wait duration, or `None` if we've exceeded the 5-minute timeout we set in `MAX_ELAPSED_TIME`. Note that it's only used during the initial fetch — the polling loop doesn't use this backoff 👍

crates/apollo-mcp-registry/src/platform_api/operation_collections/collection_poller.rs
```diff
@@ -304,64 +333,102 @@ impl CollectionSource {
...
-                    | OperationCollectionResult::ValidationError(ValidationError { message }) => {
-                        if let Err(e) = sender
-                            .send(CollectionEvent::CollectionError(CollectionError::Response(
-                                message,
-                            )))
-                            .await
-                        {
-                            tracing::debug!(
-                                "failed to send error to collection stream. This is likely to be because the server is shutting down: {e}"
-                            );
```
gocamille, line 327: We now loop here in order to retry the initial fetch if it fails. The original logic stays the same, the only differences are that we will flag that we've completed an initial fetch (`has_fetched` = true) and we will retry if it fails.

crates/apollo-mcp-registry/src/platform_api/operation_collections/collection_poller.rs
```diff
@@ -304,64 +333,102 @@ impl CollectionSource {
...
-                            err
-                        );
-                    } else {
-                        tracing::error!(
-                            "Failed to fetch initial operation collection with permanent error: {err}"
-                        );
-                        if let Err(e) = sender.send(CollectionEvent::CollectionError(err)).await {
-                            tracing::debug!(
-                                "failed to send error to collection stream. This is likely to be because the server is shutting down: {e}"
+                    },
```
gocamille: We've expanded the error handling here so we get a lot more information about what went wrong, specifically we want to know whether the error is transient (if so we can retry), if the backoff has already been exhausted, or if there's another permanent error.

crates/apollo-mcp-registry/src/platform_api/operation_collections/collection_poller.rs
```diff
@@ -414,94 +481,132 @@ impl CollectionSource {
...
-                        match variant.mcp_default_collection {
-                            DefaultCollectionResult::OperationCollection(collection) => {
-                                let should_poll = write_init_response(
-                                    &sender,
-                                    &mut previous_updated_at,
-                                    collection.operations.iter().map(OperationData::from),
-                                )
-                                .await;
-                                if !should_poll {
-                                    return;
```
gocamille, line 451: Same as above, we now loop here in order to retry the initial fetch if it fails.

gocamille:
A note about the code coverage: 

This patch coverage percentage is expected due to our current testing infrastructure. What we have tested are:
- The new `retry_initial_fetch` function that covers the business logic in this PR (covered with a mockfetcher)
- `is_collection_error_transient` — still is tested for all error types under our control. the transient case for Request errors just delegates to `reqwest's` own methods and testing that would be testing the crate, not our logic.

What is not tested and why:
- Retry loop internals — these run inside spawned `async` tasks and would need integration tests with a mock server. the logic is straightforward and the helpers it uses are covered.
- Request transient errors since they delegate to the `reqwest` crate's own methods.

The latter definitely strengthens the case for reviewing the testing architecture!

.changeset/add_exponential_retry_backoffs_during_initial_collection_fetch.md
```diff
@@ -0,0 +1,5 @@
+---
+default: patch
```
andrewmcgivery: This feels like more of a minor than a patch since it's not a bug fix unless I'm misunderstanding the change.

crates/apollo-mcp-registry/src/platform_api/operation_collections/collection_poller.rs
gocamille: Note that this comment is outdated after the refactor in 352ab4ea8624d8e073e42c4720593156be74534c.

crates/apollo-mcp-registry/src/platform_api/operation_collections/collection_poller.rs
gocamille: Note that this comment is outdated after the refactor in 352ab4ea8624d8e073e42c4720593156be74534c.

crates/apollo-mcp-registry/src/platform_api/operation_collections/collection_poller.rs
gocamille: Note that this comment is outdated after the refactor in 352ab4ea8624d8e073e42c4720593156be74534c.

crates/apollo-mcp-registry/src/platform_api/operation_collections/collection_poller.rs
gocamille: Note that this comment is outdated after the refactor in 352ab4ea8624d8e073e42c4720593156be74534c.

.changeset/add_exponential_retry_backoffs_during_initial_collection_fetch.md
gocamille: Good point! I've updated it!

PR #562, opened 2026-01-15
Apps: Swap out mime type based on detected app target (Apps SDK vs MCP Apps)

andrewmcgivery:


crates/apollo-mcp-server/src/server/states/running.rs
```diff
@@ -268,15 +268,17 @@ impl Running {
         })
     }
 
-    fn list_resources_impl(&self) -> ListResourcesResult {
-        ListResourcesResult {
+    fn list_resources_impl(&self, extensions: Extensions) -> Result<ListResourcesResult, McpError> {
```
DaleSeo, line 271: The `list_resources` endpoint now dynamically sets the MIME type based on appTarget, but `read_resource` still hardcodes `text/html+skybridge`. Is this intentional, or should `get_app_resource` also respect the app target?

https://github.com/apollographql/apollo-mcp-server/blob/0f016b8b649b6d309bf3e1110ba418b3bbb3c7d7/crates/apollo-mcp-server/src/apps/execution.rs#L268

crates/apollo-mcp-server/src/apps/execution.rs
```diff
@@ -262,6 +271,34 @@ pub(crate) async fn get_app_resource(
...
+                .find(|(key, _)| key == "appTarget")
+                .map(|(_, value)| value.into_owned())
+        });
+
+    match app_target_param {
+        Some(app_target) if app_target.to_lowercase() == "openai" => Ok(AppTarget::AppsSDK),
+        Some(app_target) if app_target.to_lowercase() == "mcp" => Ok(AppTarget::MCPApps),
+        Some(app_target) => Err(McpError::new(
+            ErrorCode::INVALID_REQUEST,
+            format!("App target {app_target} not recognized."),
```
DaleSeo: Listing the valid values will help users quickly fix the issue.

```suggestion
            format!("App target {app_target} not recognized. Valid values are 'openai' or 'mcp'."),
```

crates/apollo-mcp-server/src/server/states/running.rs
```diff
@@ -703,7 +705,8 @@ mod tests {
     #[tokio::test]
     async fn resource_list_includes_app_resources() {
         let resources = running_with_apps(AppResource::Local("abcdef".to_string()), None, None)
-            .list_resources_impl()
+            .list_resources_impl(Extensions::new())
```
DaleSeo, line 708: Consider adding a test that explicitly passes an app target to `list_resources_impl` to verify both paths are exercised.

crates/apollo-mcp-server/src/apps/execution.rs
```diff
@@ -741,11 +778,102 @@ mod tests {
             None,
         );
 
-        let result = attach_resource_mime_type(resource);
+        let mut extensions = Extensions::new();
+        let request = axum::http::Request::builder()
+            .uri("http://localhost?appTarget=openai")
```
DaleSeo, line 783: Is using a query parameter for platform selection a standard pattern? I did some research and couldn't find other MCP servers doing it this way. [MCP-UI](https://mcpui.dev/guide/mcp-apps), for example, uses build-time adapter selection where you configure the target platform once, rather than switching per-request.

I'm wondering if we should consider using the `Accept` header for content negotiation instead of a query parameter. Clients could request `Accept: text/html+skybridge` for OpenAI or `Accept: text/html;profile=mcp-app` for MCP Apps. This follows standard HTTP conventions and would be more discoverable. We could still fall back to a default if no specific header is provided.

Alternatively, if runtime switching isn't a hard requirement, build-time configuration might be simpler and more predictable.

crates/apollo-mcp-server/src/server/states/running.rs
andrewmcgivery: Yes, I purposefully did not change this in this PR as I wanted to keep this PR small and targetted. I will have follow up PRs to update the tool meta data and resource meta data. :)

crates/apollo-mcp-server/src/apps/execution.rs
andrewmcgivery: 1. Runtime switching is a requirement as part of our "write once, deploy anywhere" strategy. If we were to make it fixed via config, users would have to run multiple instances of the MCP server which is not desirable. Again, the goal is that they can serve to multiple targets off of a single codebase and a single MCP Server.
2. You won't find other MCP Servers doing it this way cause nobody is (yet) trying to solve this problem of serving to both at the same time. Remember also that MCP Apps is in _draft_ so we're very, very early.
3. We cannot use headers because they are not universally accepted/configurable by hosts/clients. ChatGPT for example does not have the ability to add headers when registering an app. The URL is the only thing that is universal and this is the same reason why we used the `app` URL param for switching which app we are serving.
4. Worth mentioning, we also intend to have "auto detection" once the MCP Apps capability advertising is standardized and implemented into all the SDKs (it's currently blocked by another SEP). Once we have that, the url-based approach may not be required anymore or may be something that users won't have to (usually) use.

DaleSeo: That makes sense. Thanks for the context, @andrewmcgivery! 

Just one last minor thing: `appTarget` is similar to the existing `app` parameter. Would something like `uiRuntime` or `clientTarget` be clearer to avoid confusion?

andrewmcgivery: I purposefully had made it `appTarget` to sort of "namespace" any params having to do with apps to make it really clear what they were for.... just incase we have other params unrelated to apps in the future for whatever reason.

I'm not against changing it but I do like that with `app` included in the name of the param, there is no ambiguity what it is for. :D

PR #563, opened 2026-01-15
Restore tarball directory structure for Rover compatibility

DaleSeo:
This PR fixes a regression that broke Rover's ability to download and run the MCP server binary.

When the release workflow was rewritten recently, the tarball directory structure was changed. The original NIX-based workflow packaged the binary at `dist/`, but the new workflow put it under `artifact/` instead.

Rover expects the binary to be located at `dist/apollo-mcp-server` inside the tarball. When users run `rover dev --mcp`, the plugin downloads successfully but Rover fails to find the binary at the expected path, producing the following error: 

```sh
$ rover dev --mcp mcp.yaml
error: unable to find dependency: "error: could not find a file at the path '/var/folders/v2/5j2zr9qs7c76ph0yxn53d_9r0000gp/T/apollo-mcp-server7j25GC/dist/apollo-mcp-server'

Caused by:
    No such file or directory (os error 2)
"
```

The fix simply changes `archive_name` from `artifact` to `dist` in both the `build-artifacts` and `generate-config-schema` jobs, restoring the original tarball structure that Rover expects.

### DIFF
--- a/.changeset/restore_tarball_directory_structure_for_rover_compatibility.md
+++ b/.changeset/restore_tarball_directory_structure_for_rover_compatibility.md
@@ -1,0 +1,9 @@
+---
+default: patch
+---
+
+### Restore tarball directory structure for Rover compatibility - @DaleSeo PR #563
+
+Fixes a regression that broke Rover's ability to download and run the MCP server binary.
+
+When the release workflow was rewritten recently, the tarball directory structure was changed. The original NIX-based workflow packaged the binary at dist/, but the new workflow put it under artifact/ instead.


.github/workflows/release-bins.yml
andrewmcgivery, line 256: Probably need to update this too

gocamille (approved):

.github/workflows/release-bins.yml
andrewmcgivery: Actually... nvm, I don't think we need to cause `artifact` != `artifacts`

PR #567, opened 2026-01-16
MCP Apps: Resource metadata

andrewmcgivery:
Add swapping of resource metadata and mime type based on the app target.

### DIFF
--- a/crates/apollo-mcp-server/src/apps/execution.rs
+++ b/crates/apollo-mcp-server/src/apps/execution.rs
@@ -241,6 +241,7 @@ pub(crate) async fn get_app_resource(
     let mut meta: Option<Meta> = None;
     if let Some(csp) = &app.csp_settings {
         match app_target {
+            // Note that the difference in which keys are set here and the camelCase vs snake_key is on purpose. These are differences between the two specs.
             AppTarget::AppsSDK => {
                 meta.get_or_insert_with(Meta::new).insert(
                     "openai/widgetCSP".into(),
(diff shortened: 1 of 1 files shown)


crates/apollo-mcp-server/src/apps/execution.rs
```diff
@@ -235,13 +240,35 @@ pub(crate) async fn get_app_resource(
...
+                );
+            }
+            AppTarget::MCPApps => {
+                meta.get_or_insert_with(Meta::new).insert(
+                    "csp".into(),
+                    json!({
+                        "connectDomains": csp.connect_domains,
+                        "resourceDomains": csp.resource_domains,
+                        "frameDomains": csp.frame_domains,
+                        "baseUriDomains": csp.base_uri_domains
```
DaleSeo, line 263: MCPApps needs `baseUriDomains`, but it doesn't use `redirect_domains`. Also, do they have different letter casing for the keys? A brief comment would help those who aren't as familiar with these subtle API differences.

crates/apollo-mcp-server/src/apps/execution.rs
```diff
@@ -876,4 +925,202 @@ mod tests {
...
+        assert!(csp.get("base_uri_domains").is_none());
+        assert!(meta.get("openai/widgetDescription").is_some());
+        assert!(meta.get("openai/widgetDomain").is_some());
+        assert!(meta.get("openai/widgetPrefersBorder").is_some());
+        // AppsSDK should not have ui nesting
+        assert!(meta.get("ui").is_none());
+    }
+
+    #[tokio::test]
+    async fn get_app_resource_returns_openai_format_when_target_is_openai() {
```
DaleSeo, line 931: How is this test different from the previous one?

crates/apollo-mcp-server/src/apps/execution.rs
andrewmcgivery: Yea... they have slightly different CSP settings, including don't have all the same keys, and one uses camalCase and one uses snake_case... fun times lol

I'll add a comment.

crates/apollo-mcp-server/src/apps/execution.rs
andrewmcgivery: Good catch. That would be Claude being stupid and Andrew being stupid not catching it lol 😅

PR #570, opened 2026-01-19
Add MCP Apps tool meta data

andrewmcgivery:


crates/apollo-mcp-server/src/apps/execution.rs
```diff
@@ -142,36 +142,65 @@ fn filter_inputs_for_operation(
...
-pub(crate) fn make_tool_private(mut tool: Tool) -> Tool {
+pub(crate) fn make_tool_private(mut tool: Tool, app_target: &AppTarget) -> Tool {
     let meta = tool.meta.get_or_insert_with(Meta::new);
-    meta.insert("openai/visibility".into(), "private".into());
+    match app_target {
+        AppTarget::AppsSDK => meta.insert("openai/visibility".into(), "private".into()),
+        AppTarget::MCPApps => {
+            let mut ui = Meta::new();
+            ui.insert("visibility".into(), json!(["app"]));
+            meta.insert("ui".into(), serde_json::to_value(ui).unwrap_or_default())
```
gocamille, line 152: What would the `default` value be in this instance, and is it something that MCPApps expects?

crates/apollo-mcp-server/src/apps/execution.rs
```diff
@@ -142,36 +142,65 @@ fn filter_inputs_for_operation(
...
+        && let Some(tool_invocation_invoked) = &tool.labels.tool_invocation_invoked
+    {
         meta.insert(
             "openai/toolInvocation/invoked".into(),
             tool_invocation_invoked.clone().into(),
         );
     }
 
+    // In the case of MCP Apps, the meta data is nested under `_meta.ui`
+    if matches!(app_target, AppTarget::MCPApps) {
```
gocamille, line 202: I am a noob to the lifecycle of MCPApps 🙏 but could there be a possibility that this overrides the previous set of metadata unexpectedly (or is that expected)?

gocamille:
Thank you @andrewmcgivery I'm excited for this! I had some questions but this looks great! 

I was also looking at `AppTarget` more and wondered if moving the `match` statements currently in the adapter functions (`make_tool_private`, `get_mime_type`, etc.) and have an implementation for `AppTarget`: 

```
impl AppTarget {
    pub fn private_visibility(&self) -> Value {
        match self {
            Self::AppsSDK => json!("private"),
            Self::MCPApps => json!({"visibility": ["app"]}),
        }
    }
    pub fn mime_type(&self) -> &'static str {
        match self {
            Self::AppsSDK => "text/html+skybridge",
            Self::MCPApps => "text/html;profile=mcp-app",
        }
    }
...cont'd
```

Then inside the adapter functions: 

```
pub(crate) fn make_tool_private(mut tool: Tool, app_target: &AppTarget) -> Tool {
    let meta = tool.meta.get_or_insert_with(Meta::new);
    meta.insert(
        app_target.visibility_key().into(),
        app_target.private_visibility(),
    );
    tool
}
```

I thought of this as possibly easier to grok and extend for future us!

crates/apollo-mcp-server/src/apps/execution.rs
andrewmcgivery: `Value::Null`... but technically that would never happen because the thing we are giving it is also derived from a `Value` so this can't actually fail. :)

crates/apollo-mcp-server/src/apps/execution.rs
andrewmcgivery: Nope! This is the one and only function that is setting `_meta.ui`, by design. The tool list call is completely stateless call each time it is called so there is no way this could be overwritten. I also very purposefully moved everything into central functions in this file to avoid losing track of these being set other places in the codebase. :)

andrewmcgivery:
> Thank you @andrewmcgivery I'm excited for this! I had some questions but this looks great!
> 
> I was also looking at `AppTarget` more and wondered if moving the `match` statements currently in the adapter functions (`make_tool_private`, `get_mime_type`, etc.) and have an implementation for `AppTarget`:
> 
> ```
> impl AppTarget {
>     pub fn private_visibility(&self) -> Value {
>         match self {
>             Self::AppsSDK => json!("private"),
>             Self::MCPApps => json!({"visibility": ["app"]}),
>         }
>     }
>     pub fn mime_type(&self) -> &'static str {
>         match self {
>             Self::AppsSDK => "text/html+skybridge",
>             Self::MCPApps => "text/html;profile=mcp-app",
>         }
>     }
> ...cont'd
> ```
> 
> Then inside the adapter functions:
> 
> ```
> pub(crate) fn make_tool_private(mut tool: Tool, app_target: &AppTarget) -> Tool {
>     let meta = tool.meta.get_or_insert_with(Meta::new);
>     meta.insert(
>         app_target.visibility_key().into(),
>         app_target.private_visibility(),
>     );
>     tool
> }
> ```
> 
> I thought of this as possibly easier to grok and extend for future us!

Probably not. I had considered this earlier and actually started down this path but... This assumes there is a 1:1 relationship and that every key exists in both standards.. and unfortunately, neither are true. I also expect there to continue to be these differences as the MCP Apps spec id developed.

Further, I am planning to do some refactoring of the app files in a separate PR to split out the tools and resources to make this execution file less of a hot mess. `AppTarget` is going to end up in a separate file from tools and resources and so having all the implementation details with AppTarget would make that refactor not work.

crates/apollo-mcp-server/src/apps/execution.rs
```diff
@@ -142,36 +142,65 @@ fn filter_inputs_for_operation(
...
-pub(crate) fn attach_tool_metadata(app: &App, tool: &AppTool) -> Tool {
+pub(crate) fn attach_tool_metadata(app: &App, tool: &AppTool, app_target: &AppTarget) -> Tool {
     let mut inner_tool = tool.tool.clone();
     let meta = inner_tool.meta.get_or_insert_with(Meta::new);
-    meta.insert(
-        "openai/outputTemplate".to_string(),
-        app.uri.to_string().into(),
-    );
-    meta.insert("openai/widgetAccessible".to_string(), true.into());
+    let mut ui = Meta::new();
```
DaleSeo: It would be a bit better to put this under the `AppTarget::MCPApps` arm since it's only used there.

andrewmcgivery: Used the `None` and `get_or_insert_with` pattern to save the allocation. :)

PR #576, opened 2026-01-20
Fix transient error handling in collection polling loop

DaleSeo:
The MCP server currently restarts when it encounters network errors while polling operation collections from GraphOS. This causes unnecessary server restarts during temporary connectivity issues.

This PR applies the same transient error handling pattern to both polling loops (`collection_id_stream` and `default_collection_stream`). Transient errors are now logged as warnings and the server retries on the next poll interval (30 seconds), while permanent errors like permission denied or not found still trigger a restart.

### DIFF
--- a/.changeset/collection_polling_loop.md
+++ b/.changeset/collection_polling_loop.md
@@ -1,0 +1,9 @@
+---
+default: patch
+---
+
+### Fix transient error handling in collection polling loop - @DaleSeo PR #576
+
+The MCP server currently restarts when it encounters network errors while polling operation collections from GraphOS. This causes unnecessary server restarts during temporary connectivity issues.
+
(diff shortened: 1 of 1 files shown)


crates/apollo-mcp-registry/src/platform_api/operation_collections/collection_poller.rs
```diff
@@ -393,10 +393,23 @@ impl CollectionSource {
                         tracing::debug!("Operation collection unchanged");
                     }
                     Err(err) => {
-                        if let Err(e) = sender.send(CollectionEvent::CollectionError(err)).await {
-                            tracing::debug!(
-                                "failed to send error to collection stream. This is likely to be because the server is shutting down: {e}"
+                        if is_collection_error_transient(&err) {
+                            // Log transient errors but don't send CollectionError to prevent server restart
```
gocamille, line 397: Should we log these errors at the `info` level rather than the `warn` level since this is an issue the user cannot act upon?

DaleSeo: I added these warnings because if this issue keeps happening, users might think it's not just temporary and will contact Apollo.

PR #577, opened 2026-01-20
Extract collection error handling to improve testability

DaleSeo:
I didn't meet the test coverage target in PR #576 because adding unit tests was difficult. The error handling logic was deeply nested inside `tokio::task::spawn` async blocks within the polling loops. The code was tightly coupled with GraphQL requests, meaning any test would need to mock the entire request/response flow.

This PR refactors the error handling logic in the `operation_poller` module into `error` module so we can write focused unit tests properly.

### DIFF
--- a/crates/apollo-mcp-registry/src/platform_api/operation_collections/collection_poller.rs
+++ b/crates/apollo-mcp-registry/src/platform_api/operation_collections/collection_poller.rs
@@ -327,7 +327,8 @@ impl CollectionSource {
                     }
                 },
                 Err(err) => {
-                    if err.is_permanent(&sender).await {
+                    if !err.is_transient() {
+                        err.send_to_stream(&sender).await;
                         return;
                     }
                 }
@@ -364,7 +365,8 @@ impl CollectionSource {
                         tracing::debug!("Operation collection unchanged");
                     }
                     Err(err) => {
-                        if err.is_permanent(&sender).await {
+                        if !err.is_transient() {
+                            err.send_to_stream(&sender).await;
                             break;
                         }
                     }
@@ -449,7 +451,8 @@ impl CollectionSource {
                     }
                 },
                 Err(err) => {
-                    if err.is_permanent(&sender).await {
+                    if !err.is_transient() {
+                        err.send_to_stream(&sender).await;
                         return;
                     }
                 }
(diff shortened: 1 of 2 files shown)


crates/apollo-mcp-registry/src/platform_api/operation_collections/collection_poller.rs
```diff
@@ -288,15 +274,15 @@ impl CollectionSource {
...
             }
             CollectionSource::Default(ref graph_ref, ref platform_api_config) => {
-                self.default_collection_stream(graph_ref.clone(), platform_api_config.clone())
+                self.stream_default_collection(graph_ref.clone(), platform_api_config.clone())
             }
         }
     }
 
-    fn collection_id_stream(
+    fn stream_collection_by_id(
```
DaleSeo, line 285: Also renamed these two functions for clarity:
- `collection_id_stream` -> `stream_collection_by_id`
- `default_collection_stream` -> `stream_default_collection`

crates/apollo-mcp-registry/src/platform_api/operation_collections/error.rs
```diff
@@ -17,3 +20,169 @@ pub enum CollectionError {
     #[error("invalid variables: {0}")]
     InvalidVariables(String),
 }
+
+impl CollectionError {
+    /// Returns `true` if the error is permanent and sends it to the collection stream.
+    /// Returns `false` if the error is transient (will be retried).
+    #[allow(clippy::wrong_self_convention)] // Intentionally consumes self to send over channel
+    pub(super) async fn is_permanent(self, sender: &Sender<CollectionEvent>) -> bool {
```
andrewmcgivery: The automated Claude flagged this and I think I agree that having a boolean method named "is_xxxx" that has a side effect is a bit of a confusing anti-pattern. But I understand your intent here is to encapsulate the check and the result of the check so it doesn't get repeated several times in the other file.

To avoid splitting it, maybe just renamed it to like `send_if_permanent` would address this without having to change any of the actual code here.

DaleSeo: That's interesting, @andrewmcgivery! My AI agent also suggested some other function names including `send_if_permanent`, but I didn't like any of them. I thought they might confuse callers in a different way since they don't clearly show a boolean return value. I took this as a sign that I might've been too greedy and put too much in a single function. I've broken it down into two functions. Hope you like that refactor!

PR #579, opened 2026-01-21
Claude Code Review Setup

andrewmcgivery:
- Code review using Claude Code github action
- Currently it is fed our "rust best practices" authored in 2025 (we have room to make tweaks there as needed)
- Only will review PRs with changes to `*.rs` files
- Defaults to Sonnet but auto-upgrades to Opus for complex PRs (>500 lines)
- Can manually set model using labels `ai-review-model-opus` or `ai-review-model-sonnet`
- Can skip completely with label `skip-ai-review`

### DIFF
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -2,6 +2,8 @@ name: Code Review
 on:
   pull_request:
     types: [opened, synchronize, ready_for_review]
+    paths:
+      - '**/*.rs'
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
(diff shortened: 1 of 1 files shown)


.github/workflows/claude-code-review.yml
```diff
@@ -0,0 +1,98 @@
...
+      - name: Checkout code
+        uses: actions/checkout@v6
+
+      - name: Check if review should run
+        id: should_review
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          # Check if any .rs files were changed
+          HAS_RS_FILES=$(gh pr diff ${{ github.event.pull_request.number }} --name-only | grep -c '\.rs$' || true)
```
jerelmiller: I think you can simplify this and move this check up to the top-level `on`. See https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#example-including-paths:

```yml
on:
  pull_request:
    types: [opened, synchronize, ready_for_review]
  push:
    paths:
      - '*.rs'
```

andrewmcgivery: Good call! I used `on.pull_request.paths` instead of `push` but still a really good simplification!

jerelmiller: Ah I missed it could be nested under `pull_request`. Good catch!

gocamille (approved):

PR #581, opened 2026-01-21
Update all workspace crate versions in `Cargo.lock` during release prep

DaleSeo:
While I was releasing v1.5.0, I checked out the release branch and saw that my IDE updated `Cargo.lock` because the Prepare Release workflow didn't update all the workspace crate versions. This PR fixes that issue.

I also have a couple of suggestions:

-  Change the branch naming from `release` to `release/$version` to avoid reusing the same branch for each release.
-  Automatically apply the `release` label to release PRs so it's easier to find them and keep consistency with previous release PRs.

### DIFF
--- a/.github/workflows/prep-release.yml
+++ b/.github/workflows/prep-release.yml
@@ -26,8 +26,3 @@ jobs:
       - run: knope prepare-release --verbose
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-
-      - name: Label release PR
-        run: gh pr edit --add-label release
-        env:
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/knope.toml
+++ b/knope.toml
@@ -45,6 +45,11 @@ template = "Release $version"
 [workflows.steps.body]
 template = "This PR was created by Knope. Merging it will create a new release\n\n$changelog"
 
+[[workflows.steps]]
+type = "Command"
+command = "gh pr edit --add-label release"
+shell = true
+
 [[workflows]]
 name = "release"
 


knope.toml
```diff
@@ -15,7 +17,7 @@ name = "prepare-release"
 
 [[workflows.steps]]
 type = "Command"
-command = "git switch -c release"
+command = "git switch -c release/$version"
```
andrewmcgivery: This won't work the way you think it will 😅 

`$version` here will be the *current* version because `PrepareRelease` hasn't been run yet... so your branch name here and then later on like 32 will be different names.

knope.toml
```diff
@@ -2,6 +2,8 @@
 versioned_files = [
     "Cargo.toml",
     { path = "Cargo.lock", dependency = "apollo-mcp-server" },
+    { path = "Cargo.lock", dependency = "apollo-mcp-registry" },
+    { path = "Cargo.lock", dependency = "apollo-schema-index" },
```
andrewmcgivery, line 6: Did you double check this with `knope prepare-release --dry-run --verbose` so make sure this is going to work as expected? :)

.github/workflows/prep-release.yml
```diff
@@ -26,3 +26,8 @@ jobs:
       - run: knope prepare-release --verbose
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Label release PR
+        run: gh pr edit --add-label release
```
andrewmcgivery: I would probably add this as a step to our knope file instead of a step in the github action.

knope.toml
andrewmcgivery: (it should work but worth running the command quickly locally to make sure it doesn't immediately blow up haha)

knope.toml
DaleSeo: Good catch! Fixed by creating a temporary branch first, then renaming it after `PrepareRelease` bumps the version.

knope.toml
DaleSeo: Oh, I did not know the command. Let me give that a try.

.github/workflows/prep-release.yml
DaleSeo: I considered that first but didn't figure out how to install `gh` on Knope and pass the github token secret. I guess this is probably not too bad because we run `git config` outside Knope as well?

andrewmcgivery: You can set the environment variable on the gh actions step the same way:

```
env:
          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

And if you set a command to be run via shell `shell=true`, it should Just Work ™️

knope.toml
DaleSeo: Pretty cool! 👍

2026-01-21 at 17 20 16

.github/workflows/prep-release.yml
DaleSeo: Interesting... Now I wonder why you want to have these commands in the Knope config file. All these commands could be combined into a single step in GitHub Actions, and I prefer not to tightly integrate with Knope due to the learning curve. I mean, most devs are much more familiar with GitHub Actions than with Knope. The constellation's Knope config looks very minimal unlike ours.

andrewmcgivery: It's mostly about how you encapsulate the concept of "prepare release" as a "task"/"workflow".

With the exception of setting the github config, all the other git commands for "prepare release" are being run via knope. The Github config is a bit of a "pre-requisite" step to "prepare release" rather than a part of "prepare release" itself.

If the tagging the PR is part of the process of "prepare release", it should be part of that encapsulation.

PR #582, opened 2026-01-21
Refactor apps code into app, manifest, resource, and tool

andrewmcgivery:
- Refactor apps code into app, manifest, resource, and tool
- Add additional tests for app_target
- Some minor function renames
- `get_app_target` refactored to `AppTarget::try_from`

### DIFF
--- a/crates/apollo-mcp-server/src/apps/app.rs
+++ b/crates/apollo-mcp-server/src/apps/app.rs
@@ -1,0 +1,201 @@
+use std::sync::Arc;
+
+use rmcp::model::{ErrorCode, Extensions, RawResource, Resource, Tool};
+use url::Url;
+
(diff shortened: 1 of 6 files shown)


andrewmcgivery:
> @andrewmcgivery thank you! I've combed through this refactor and appreciate the separation, especially organizing the Resource and Tool logic for constructing the app metadata based on the user's target choice 🎉
> 
> Would we want to document in the manifest file which fields are target-specific? I noticed a few that are silently ignored depending on `AppTarget` (i.e. Open AI Apps SDK only has `toolInvocation/invoking` and `toolInvocation/invoked` labels, `widgetSettings.description`, and others; MCP Apps has `csp.base_uri_domains`)
> 
> Not a blocker! This implementation follows both specs as expected. Ship it 🚢

This is largely going to be more of a docs issue which... is a later Andrew problem. 😆

PR #584, opened 2026-01-21
feat(auth): support RFC 8414 and OIDC Discovery for auth server metadata

gocamille:
#### Summary 

Update authorization server metadata discovery to support both OAuth 2.0 AS Metadata (RFC 8414) and OpenID Connect Discovery 1.0, per [MCP Auth Spec 2025-11-25](https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#authorization-server-metadata-discovery).

#### Background

The previous implementation only attempted a single discovery URL pattern. The latest spec requires trying multiple URL patterns to ensure compatibility with various OAuth providers:

- RFC 8414 (path-insertion): `/.well-known/oauth-authorization-server/{path}`
- OIDC Discovery (path-insertion): `/.well-known/openid-configuration/{path}`
- OIDC Discovery (legacy path-appending): `/{path}/.well-known/openid-configuration`

#### Changes

- New `build_discovery_urls` function which builds discovery URLs in priority order per MCP spec
- New `discover_jwks` function 
- URL normalization to strip query strings/fragments per [RFC 8414 Section 3.1](https://datatracker.ietf.org/doc/html/rfc8414#section-3.1), which requires us to collapse double slashes

### DIFF
--- a/docs/source/auth.mdx
+++ b/docs/source/auth.mdx
@@ -57,3 +57,27 @@ If you set `allow_any_audience` to `true` (the default is `false`), Apollo MCP S
 Skipping audience validation reduces security. Only use `allow_any_audience: true` when you trust all tokens issued by your configured OAuth servers, regardless of their intended audience.
 
 </Caution>
+
+## Performance considerations
+
+### Discovery timeout
+
+Authorization server metadata is discovered using OAuth 2.0 Authorization Server Metadata (RFC 8414) and OpenID Connect Discovery. The MCP Server tries multiple discovery URL patterns in sequence until one succeeds.
+
+The `discovery_timeout` setting controls how long to wait for each discovery URL attempt. The default is 5 seconds per URL.
+
+```yaml title="mcp.yaml"
+transport:
+  type: streamable_http
+  auth:
+    servers:
+      - https://auth.example.com
+    discovery_timeout: 10s  # Increase timeout for slower networks (default: 5s)
+```
+
+**Considerations:**
+- Discovery happens once at startup when validating the first JWT token
+- With multiple fallback URLs (RFC 8414, OIDC Discovery), the cumulative timeout can be 10-15 seconds if all URLs fail
+- Increase the timeout if your OAuth server is on a slow network or responds slowly
+- Decrease the timeout in high-performance environments where fast failure is preferred
+- Discovery results are cached by the underlying JWKS library, so subsequent token validations are fast
--- a/docs/source/config-file.mdx
+++ b/docs/source/config-file.mdx
@@ -242,17 +242,18 @@ For Apollo MCP Server `≤v1.0.0`, the default `port` value is `5000`. In `v1.1.
 
 These fields are under the top-level `transport` key, nested under the `auth` key. Learn more about [authorization and authentication](/apollo-mcp-server/auth).
 
-| Option                            | Type           | Default | Description                                                                                        |
-| :-------------------------------- | :------------- | :------ | :------------------------------------------------------------------------------------------------- |
-| `servers`                         | `List<URL>`    |         | List of upstream delegated OAuth servers (must support OIDC metadata discovery endpoint)           |
-| `audiences`                       | `List<string>` | `[]`    | List of accepted audiences from upstream signed JWTs (ignored if `allow_any_audience` is `true`)   |
-| `allow_any_audience`              | `bool`         | `false` | Set to `true` to skip audience validation entirely (use with caution)                              |
-| `resource`                        | `string`       |         | The externally available URL pointing to this MCP server. Can be `localhost` when testing locally. |
-| `resource_documentation`          | `string`       |         | Optional link to more documentation relating to this MCP server                                    |
-| `scopes`                          | `List<string>` |         | List of queryable OAuth scopes from the upstream OAuth servers                                     |
-| `disable_auth_token_passthrough`  | `bool`         | `false` | Optional flag to disable passing validated Authorization header to downstream API                  |
-| `tls.ca_cert`                     | `string`       |         | Path to a CA certificate to trust (PEM format).                                                    |
-| `tls.danger_accept_invalid_certs` | `bool`         | `false` | Accepts invalid TLS certificates. Set this to `true` for development or testing purposes only.     |
+| Option                            | Type           | Default | Description                                                                                                                                       |
+| :-------------------------------- | :------------- | :------ | :------------------------------------------------------------------------------------------------------------------------------------------------ |
+| `servers`                         | `List<URL>`    |         | List of upstream delegated OAuth servers (must support OIDC metadata discovery endpoint)                                                          |
+| `audiences`                       | `List<string>` | `[]`    | List of accepted audiences from upstream signed JWTs (ignored if `allow_any_audience` is `true`)                                                  |
+| `allow_any_audience`              | `bool`         | `false` | Set to `true` to skip audience validation entirely (use with caution)                                                                             |
+| `resource`                        | `string`       |         | The externally available URL pointing to this MCP server. Can be `localhost` when testing locally.                                                |
+| `resource_documentation`          | `string`       |         | Optional link to more documentation relating to this MCP server                                                                                   |
+| `scopes`                          | `List<string>` |         | List of queryable OAuth scopes from the upstream OAuth servers                                                                                    |
+| `disable_auth_token_passthrough`  | `bool`         | `false` | Optional flag to disable passing validated Authorization header to downstream API                                                                 |
+| `discovery_timeout`               | `Duration`     | `5s`    | Timeout for authorization server metadata discovery requests. Supports human-readable durations (e.g., "5s", "10s", "30s").                       |
+| `tls.ca_cert`                     | `string`       |         | Path to a CA certificate to trust (PEM format).                                                                                                   |
+| `tls.danger_accept_invalid_certs` | `bool`         | `false` | Accepts invalid TLS certificates. Set this to `true` for development or testing purposes only.                                                    |
 
 Below is an example configuration using `StreamableHTTP` transport with authentication:
 
@@ -288,6 +289,9 @@ transport:
       - read
       - mcp
       - profile
+
+    # Timeout for authorization server metadata discovery (default: 5s)
+    discovery_timeout: 5s
 
     # Optional TLS configuration for connecting to OAuth servers
     tls:


crates/apollo-mcp-server/src/auth/networked_token_validator.rs
```diff
@@ -29,15 +31,101 @@ impl<'a> NetworkedTokenValidator<'a> {
...
+        .collect()
+}
+
+/// Attempts discovery from multiple URLs sequentially, returning first success
+async fn discover_jwks(client: &reqwest::Client, issuer: &Url) -> Option<Jwks> {
+    let urls = build_discovery_urls(issuer);
+
+    for url in &urls {
+        let result = tokio::time::timeout(
+            Duration::from_secs(5),
```
andrewmcgivery: Bit of a magic number here.... should this be configurable? I know in Router land stuff life this often gets requested to be configurable.

gocamille: This is a good point! I'm always weary of these and have seen a few of these magic numbers in the repo, I would tend to agree we should expose this to be configurable 👍

crates/apollo-mcp-server/src/auth/networked_token_validator.rs
```diff
@@ -29,15 +31,101 @@ impl<'a> NetworkedTokenValidator<'a> {
...
-    discovery_url
+/// Constructs discovery URLs. Returns URLs in priority order:
+/// 1. RFC 8414 (path-insertion): `/.well-known/oauth-authorization-server/{path}`
+/// 2. OIDC Discovery (path-insertion): `/.well-known/openid-configuration/{path}`
+/// 3. OIDC Discovery (legacy path-appending): `/{path}/.well-known/openid-configuration`
+///
+/// # URL Normalization
+/// Query strings and fragments are stripped per RFC 8414 Section 3.1.
+/// The spec does not define behavior for these, and most implementations strip them.
+fn build_discovery_urls(issuer: &Url) -> Vec<Url> {
```
DaleSeo: I like the new function and parameter name. 👍

crates/apollo-mcp-server/src/auth/networked_token_validator.rs
```diff
@@ -29,15 +31,101 @@ impl<'a> NetworkedTokenValidator<'a> {
...
+            origin, path_suffix
+        ),
+        // Priority 2: OIDC Discovery path-insertion
+        format!("{}/.well-known/openid-configuration{}", origin, path_suffix),
+    ];
+
+    // Priority 3: OIDC Discovery legacy path-appending (only if there IS a path)
+    if !path.is_empty() {
+        urls.push(format!(
+            "{}/{}/.well-known/openid-configuration",
```
DaleSeo, line 100: I noticed the fallback behavior changed slightly. Previously, the path-appending style used oauth-authorization-server:

```rs
{path}/.well-known/oauth-authorization-server
```

But now the path-appending fallback (priority 3) uses openid-configuration:

```rs
{path}/.well-known/openid-configuration
```

If any customers have been relying on the previous discovery behavior, this could be a breaking change for them.

Should we consider adding `{path}/.well-known/oauth-authorization-server` as a 4th fallback for backwards compatibility? Curious if you've tested against the providers we were previously supporting.

crates/apollo-mcp-server/src/auth/networked_token_validator.rs
```diff
@@ -29,15 +31,101 @@ impl<'a> NetworkedTokenValidator<'a> {
...
+    let path = normalized
+        .path()
+        .trim_matches('/')
+        .split('/')
+        .filter(|s| !s.is_empty())
+        .collect::<Vec<_>>()
+        .join("/");
+
+    let Some(host) = normalized.host_str() else {
+        trace!(issuer = %issuer, "Issuer URL has no host, cannot build discovery URLs");
```
DaleSeo: I'm a bit concerned that this trace log might go unnoticed in practice. Most users won't run at trace level unless they're actively debugging. If someone accidentally configures a URL without a host, auth will silently fail and the only hint will be buried in trace logs.

A URL without a host should be a clear configuration mistake rather than an expected edge case. Would it make sense to validate this at config load time instead? That would give users a clear startup error rather than a runtime mystery.

.gitignore
```diff
@@ -9,6 +9,7 @@
 # will have compiled files and executables
 debug/
 target/
+target-install/
```
DaleSeo: What is this for?

.changeset/support_oauth_20_and_openid_connect_metadata_discovery.md
```diff
@@ -0,0 +1,5 @@
+---
+default: minor
+---
+
+# Support OAuth 2.0 and OpenID Connect metadata discovery
```
DaleSeo, line 5:
```suggestion
### Support OAuth 2.0 and OpenID Connect metadata discovery - @gocamille PR #584

// description
```

andrewmcgivery: I wouldn't mess with the heading level. Knope makes it a h1 on purpose and will convert it to the proper heading level in changelogs later as part of it's release process.

crates/apollo-mcp-server/src/auth/networked_token_validator.rs
gocamille: @DaleSeo Good catch and I did look into this while building! I noticed the comment in commit [1fcd0fc](https://github.com/apollographql/apollo-mcp-server/commit/1fcd0fc) for Keycloak specifically. The fallback order change is intentional per MCP Auth Spec 2025-11-25.

For Keycloak, which was the vendor referenced in our project, this is not a breaking change. Keycloak serves metadata at `{realm}/.well-known/openid-configuration`, which is exactly what [Priority 3](https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#authorization-server-metadata-discovery) hits. The previous `oauth-authorization-server` path-appending only worked because some Keycloak versions serve both endpoints—and `openid-configuration` is the canonical one 👍 

One note though: Keycloak users will see 2 failed requests (Priority 1-2 are path-insertion URLs Keycloak doesn't serve) before Priority 3 succeeds. But the `404`s are fast (~50ms each), so total overhead is ~100ms.

I'm happy to add a Priority 4 for `{path}/.well-known/oauth-authorization-server` if we want strict backwards compatibility, though it's not spec-mandated and Keycloak doesn't need it.

crates/apollo-mcp-server/src/auth/networked_token_validator.rs
gocamille: @DaleSeo you're right! This should be a startup error, not a trace log. A host-less URL is a config mistake, not a runtime edge case 👍 

I've added validation in [5d8c2bf](https://github.com/apollographql/apollo-mcp-server/pull/584/commits/5d8c2bfd4e8863e83bcfe87a75db37624af2a649) to reject auth server URLs without hosts. The new `build_discovery_urls()` can then assume valid input.

.gitignore
gocamille: I was testing a target install and accidentally committed this directory so added it to `.gitignore`! However this is a rare occurrence and mistake, so I am happy to remove this!

.changeset/support_oauth_20_and_openid_connect_metadata_discovery.md
gocamille: Things I learn!

crates/apollo-mcp-server/src/auth/networked_token_validator.rs
DaleSeo: I get this from the link provided:

2026-01-22 at 14 33 06

DaleSeo: NVM, I found the validation logic.

crates/apollo-mcp-server/src/auth.rs
```diff
@@ -42,6 +42,12 @@ pub enum TlsConfigError {
     CertificateParse { path: PathBuf },
     #[error("Failed to build HTTP client: {0}")]
     ClientBuild(#[from] reqwest::Error),
+    #[error("Invalid auth server URL at index {index} ({url}): {reason}")]
+    InvalidServerUrl {
+        index: usize,
+        url: String,
+        reason: String,
```
DaleSeo: Is the `reason` field intended for future validation cases? Currently it's only ever "URL has no host". If there are no other validations planned, a simpler variant like `ServerUrlMissingHost { index, url }` might be cleaner, but happy to leave as-is if you have other checks in mind.

gocamille:
## Review Summary

This PR has already received a thorough review. I've analyzed the code against Apollo's Rust best practices and found that the previous review comprehensively covered the main issues.

### Confirmed Findings

The previous review correctly identified:
- **Line 56**: The `expect()` usage with SAFETY comment could be more robust (Chapter 4.2)
- **Test assertions**: Could use more descriptive error messages for easier debugging
- **Test naming**: Generally follows conventions well

### Additional Observations

**Performance consideration**: The `discover_jwks` function (lines 102-126) applies a 5-second timeout **per discovery URL**. With 2-3 URLs in the fallback chain, the cumulative worst-case timeout is **10-15 seconds** if all URLs fail sequentially. This is likely acceptable for a one-time initialization operation, but worth considering if this code path is hit per-request (though JWKS caching would typically prevent this).

### Assessment

**Strengths:**
- ✅ Excellent test coverage with comprehensive edge cases (URL normalization, provider patterns, fallback behavior, error handling)
- ✅ Clean implementation of RFC 8414 and OIDC Discovery spec with proper fallback logic
- ✅ Appropriate error logging with trace/warn levels
- ✅ Proper URL normalization (query strings, fragments, path handling)
- ✅ Early validation of server URLs at startup to fail fast

**Test Coverage**: ✅ Excellent - covers edge cases, error paths, and integration scenarios

**Recommendation**: **Approve with suggestions**

The implementation is solid and production-ready. The issues raised in the previous review are minor and don't block merging. The cumulative timeout consideration is worth noting but likely not a practical concern.

---
_Reviewed by Claude Code Sonnet 4.5_

crates/apollo-mcp-server/src/auth/networked_token_validator.rs
DaleSeo: Thanks for the thorough investigation, @gocamille! If you're confident, I'm okay skipping Priority 4. Just in case someone hits an edge case, I'd suggest we highlight this behavior change in the changeset file. Something like noting that the discovery order changed and users relying on legacy `{path}/.well-known/oauth-authorization-server` should verify their provider supports `openid-configuration`. 

For context, we tend to keep features deprecated rather than removed since the MCP spec changes quickly, and users often struggle to keep up. For example, we kept SSE transport around for 10 months after it was dropped from the spec. So I'm a bit cautious about removing fallback behaviors, but I'll defer to your judgment here since you've tested it.

.changeset/support_oauth_20_and_openid_connect_metadata_discovery.md
DaleSeo: Sorry about the incorrect feedback, @gocamille! 😅

crates/apollo-mcp-server/src/auth/networked_token_validator.rs
```diff
@@ -19,25 +22,116 @@ impl<'a> NetworkedTokenValidator<'a> {
...
+    let path = normalized
+        .path()
+        .trim_matches('/')
+        .split('/')
+        .filter(|s| !s.is_empty())
+        .collect::<Vec<_>>()
+        .join("/");
+
+    let Some(host) = normalized.host_str() else {
+        unreachable!("server URL must have a host (validated at startup)")
```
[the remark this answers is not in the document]
andrewmcgivery: We should probably in general avoid having panic-able code in our codebase. I'd rather double up validation and have this method return a Result<Vec<Url>> than have a `unreachable` or `expect`.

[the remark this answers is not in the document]
gocamille: Good point @andrewmcgivery (and claude 😂 )! I've updated this on [5d778c7](https://github.com/apollographql/apollo-mcp-server/pull/584/commits/5d778c71e03fbbc0c6aec1b9aa4e511f73cf3bc4)

andrewmcgivery:
@gocamille one small thing I just thought about... probably want to update the docs with the new configuration option!

crates/apollo-mcp-server/src/auth.rs
gocamille: @DaleSeo you're right that currently there's only one validation case! I kept the reason field for extensibility in case we add other URL validations in the future (e.g., unsupported schemes like `file://` or malformed paths, etc.), but I agree it's speculative.

For now I've gone ahead and simplified it based on your suggestions 👍 !

PR #595, opened 2026-01-23
Update release workflows for unique branch names

DaleSeo:
PR #581 changed the Prepare Release workflow to create unique branch names like `release/1.6.0` instead of reusing a single `release` branch. However, the release workflows weren't updated to match, so they still checked. This caused the v1.6.0 release workflows to not trigger when the release PR was merged.

### DIFF
--- a/.github/workflows/release-bins.yml
+++ b/.github/workflows/release-bins.yml
@@ -5,6 +5,12 @@ on:
     types: [closed]
     branches:
       - main
+  workflow_dispatch:
+    inputs:
+      version:
+        description: Version to publish (e.g., 1.6.0)
+        required: true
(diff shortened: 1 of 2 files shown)


.github/workflows/release-bins.yml
```diff
@@ -5,6 +5,12 @@ on:
     types: [closed]
     branches:
       - main
+  workflow_dispatch:
+    inputs:
+      version:
+        description: Version to publish (e.g., 1.6.0)
+        required: true
+        type: string
```
DaleSeo, line 13: @gocamille Added this so that we can run these workflows manually for the v1.6.0 release.

PR #597, opened 2026-01-26
fix(ci): remove manual release workflow trigger

gocamille:
This pr removes the `workflow_dispatch` trigger that was added for the v1.6.0 release in PR #595.

The core release workflows (`release-container.yml`, `release-bins.yml`) should not be triggered manually—they should always be called by entry point workflows.

#### Changes
- Removed `workflow_dispatch` trigger from `release-container.yml`
- Updated comment to reflect the remaining triggers (PR merge, `workflow_call`)

#### Verification
- [x] Canary builds still work (`release-canary.yml` passes version via `workflow_call`)
- [x] Stable releases still work (PR merge from `release/` branch works)
- [x] No manual trigger available (as is intended)

### DIFF
--- a/.github/workflows/release-container.yml
+++ b/.github/workflows/release-container.yml
@@ -5,12 +5,6 @@ on:
     types: [closed]
     branches:
       - main
-  workflow_dispatch:
-    inputs:
-      version:
-        description: Version to publish (e.g., 1.6.0)
-        required: true
-        type: string
   workflow_call:
     inputs:
       version:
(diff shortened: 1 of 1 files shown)


gocamille:
@DaleSeo I really like the treatment the `release-experimental.yaml` file ! I think that fail-fast `validation` step is preferable over the more permissive way we handle it in `release-container`. I'm going to update it accordingly.

gocamille:
@DaleSeo we've removed the `workflow_dispatch` trigger from #595 as the core release workflows should be triggered via PR merge or `workflow_call`—not manually—so that version validation stays in the entry point workflows 👍

gocamille:
> Don't forget to update the PR title and description to reflect the new direction! 😄

🤔 I did...that's odd that the description updated but not the title! It may have been an access/expiry issue! Thank you for the heads-up!

PR #609, opened 2026-01-30
Remove app name prefix from tool names

andrewmcgivery:
The app name being prefixed to the tool name is a holdover from prior to having the app=xxx query param. It was to prevent 2 apps with the same tool name from colliding. This is no longer required since we only serve the tools for 1 app given the app query param.


jerelmiller (approved):
🎉

PR #617, opened 2026-02-02
Providing custom instructions/hints for GraphQL query generation

### LINKED ISSUE #346: Providing custom instructions/hints for GraphQL query generation

DaleSeo:
Closes #346

This PR adds configurable hint text for built‑in tools so teams can inject schema‑specific guidance into the model context without changing schema descriptions. The hints are appended to existing tool descriptions to preserve default safety guidance, and they’re formatted on a new line with a `Hint:` label for visibility in tool listings.

### DIFF
--- a/crates/apollo-mcp-server/src/introspection/tools/description.rs
+++ b/crates/apollo-mcp-server/src/introspection/tools/description.rs
@@ -1,10 +1,12 @@
-pub(crate) fn append_description_hint(default: &str, hint: Option<&str>) -> String {
+use std::borrow::Cow;
+
+pub(crate) fn append_description_hint<'a>(default: &'a str, hint: Option<&str>) -> Cow<'a, str> {
     match hint.and_then(|value| {
         let trimmed = value.trim();
         (!trimmed.is_empty()).then_some(trimmed)
     }) {
-        Some(hint) => format!("{default}\nHint: {hint}"),
-        None => default.to_string(),
+        Some(hint) => Cow::Owned(format!("{default}\nHint: {hint}")),
+        None => Cow::Borrowed(default),
     }
 }
 
--- a/crates/apollo-mcp-server/src/introspection/tools/introspect.rs
+++ b/crates/apollo-mcp-server/src/introspection/tools/introspect.rs
@@ -51,7 +51,8 @@ impl Introspect {
             root_mutation_type.as_deref(),
             minify,
         );
-        let description = append_description_hint(&default_description, description_hint);
+        let description =
+            append_description_hint(&default_description, description_hint).into_owned();
         Self {
             schema,
             allow_mutations: root_mutation_type.is_some(),
--- a/crates/apollo-mcp-server/src/introspection/tools/search.rs
+++ b/crates/apollo-mcp-server/src/introspection/tools/search.rs
@@ -77,7 +77,8 @@ impl Search {
                 ""
             }
         );
-        let description = append_description_hint(&default_description, description_hint);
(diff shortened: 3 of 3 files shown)


.changeset/custom-introspection-hints.md
```diff
@@ -0,0 +1,14 @@
...
+
+# Added configurable hints for introspection tools
+
+Apollo MCP Server now supports configurable hint text for the built-in introspection tools (`execute`, `introspect`, `search`, and `validate`). These hints are appended to the tool descriptions so you can guide query generation without changing schema descriptions.
+
+```yaml
+introspection:
+  execute:
+    enabled: true
+    description: "Use carts(where: { status: ACTIVE }) for active carts."
```
jerelmiller: Would it make sense to use `hint` somewhere in here? (please take my feedback with a grain of salt though)

either:

```yml
introspection:
  execute:
    enabled: true
    hint:
      description: "Use carts(where: { status: ACTIVE }) for active carts."
```

or even:

```yml
introspection:
  execute:
    enabled: true
    hint: "Use carts(where: { status: ACTIVE }) for active carts."
```

Description makes me think you're setting the full description, not a piece of it.

DaleSeo: Good call, @jerelmiller! I'm glad you brought that up. I did consider `hint` initially, but kept `description` so we can support full overrides later without changing the config key. It preserves backward compatibility while leaving room to evolve behavior. If we switch to `hint` now, we lose that flexibility unless we support both.

jerelmiller: > so we can support full overrides later without changing the config key

I'd argue this is a breaking change so if we go this route, we need to be very clear. Making that change later using the same key means suddenly those that relied on the description from elsewhere will now need to update this description field to include the full text in order to maintain existing functionality.

If that's acceptable, then I'm ok with this. If we want something less breaking, it might be best to use hint now and deprecate it in the future when introducing description to pave a way for migrating from one to the other.

DaleSeo: That's a fair callout, @jerelmiller. I think I overthought it. I've updated the config key to `hint` as suggested.

andrewmcgivery (approved):

PR #620, opened 2026-02-04
ci: run e2e tests in CI

DaleSeo:
Adds new e2e job to ci.yml that runs mcp-server-tester for both local-operations and pq-manifest test suites. Also Fixed path resolution bug in that overwrote TEST_DIR and bumped mcp-server-tester from 1.4.0 to 1.4.1.

## Testing
The tests ran successfully on this PR: https://github.com/apollographql/apollo-mcp-server/actions/runs/21687772789/job/62539292043?pr=620

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,3 +70,24 @@ jobs:
           token: ${{ secrets.CODECOV_TOKEN }}
           files: codecov.json
           fail_ci_if_error: true
+
+  e2e:
+    name: End-to-End Tests
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+
+      - uses: actions-rust-lang/setup-rust-toolchain@v1
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: '20'
+
+      - name: Build release binary
+        run: cargo build --release
+
+      - name: Run e2e tests (local-operations)
+        run: ./e2e/mcp-server-tester/run_tests.sh local-operations
+
+      - name: Run e2e tests (pq-manifest)
+        run: ./e2e/mcp-server-tester/run_tests.sh pq-manifest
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -70,3 +70,24 @@ jobs:
           token: ${{ secrets.CODECOV_TOKEN }}
           files: codecov.json
           fail_ci_if_error: true
+
+  e2e:
```
andrewmcgivery, line 74: Do we have any concerns with CI times from adding this? How long do these generally take?

DaleSeo: I reran the job several times. It seems to take ~1.5 minutes. I wouldn't worry yet.

.github/workflows/ci.yml
```diff
@@ -70,3 +70,24 @@ jobs:
...
+        with:
+          node-version: '20'
+
+      - name: Build release binary
+        run: cargo build --release
+
+      - name: Run e2e tests (local-operations)
+        run: ./e2e/mcp-server-tester/run_tests.sh local-operations
+
+      - name: Run e2e tests (pq-manifest)
```
andrewmcgivery, line 92: Could we run this `e2e` job as a matrix so that we can run both `local-operations` and `pq-manifest` in parallel?

Your earlier comment said these are pretty fast so we might not need to worry about this but making this matrix would also be low lift so figure worth asking. :D

DaleSeo: The thing to consider is that `cargo build --release` is by far the most expensive part of this job. With a matrix, you'd be doing that Rust release build twice. Given that the tests themselves are fast, I'd lean toward keeping it sequential for now and revisiting when more test suites are added and the total test time becomes meaningful.

andrewmcgivery (approved):

PR #636, opened 2026-02-10
Adding docs for MCP Apps

pragl:
DO NOT MERGE until we've removed the feature flag for MCP Apps.

Adding 0 to 1 docs for MCP Apps.

### DIFF
--- a/docs/source/_sidebar.yaml
+++ b/docs/source/_sidebar.yaml
@@ -48,6 +48,8 @@ items:
         href: "./mcp-apps-prerequisites"
       - label: "Quickstart"
         href: "./mcp-apps-quickstart"
+      - label: "MCP App Development"
+        href: "./mcp-apps-development"
       - label: "Architecture"
         href: "./mcp-apps-architecture"
       - label: "Reference"
--- a/docs/source/mcp-apps-development.mdx
+++ b/docs/source/mcp-apps-development.mdx
@@ -1,0 +1,363 @@
+---
+title: MCP Apps Development
+subtitle: Key concepts when developing MCP Apps
+---
+
+This document explains the core concepts of developing MCP Apps using React, Apollo Client, and Apollo MCP Server.
+
+We recommend that you use the [Apollo AI Apps Template](https://github.com/apollographql/ai-apps-template) to get started developing your app. The template includes much of the code referenced in this document.
+
+## Apollo Client initialization
+
+If you have experience developing with Apollo Client, you already know most of what you need to know to develop with MCP Apps, which uses the `@apollo/client-ai-apps` package to manage GraphQL operations. However, this guide doesn't provide a comprehensive explanation of Apollo Client. To learn about it, go to the [Apollo Client documentation](../react).
+
+You initialize `ApolloClient` by using the `ApolloClient` class provided by `@apollo/client-ai-apps`. The `ApolloClient` constructor needs at least an instance of `ApolloCache` (typically `InMemoryCache`) and the [manifest file](./mcp-apps-reference#manifest-file).
+
+<Caution>
+
+Make sure you import `ApolloClient` from the `@apollo/client-ai-apps` package, not the core `@apollo/client` repository. The `ApolloClient` class from `@apollo/client-ai-apps` is for integrating with the MCP Apps environment.
+
+</Caution>
+
+```ts title=src/main.tsx
+import { InMemoryCache } from "@apollo/client";
+import { ApolloClient, ApplicationManifest } from "@apollo/client-ai-apps";
+// Note the manifest is written to the root of the app
+import manifest from "../.application-manifest.json";
+
+// You may also provide other options that the core ApolloClient accepts
+const client = new ApolloClient({
+  cache: new InMemoryCache(),
+  manifest: manifest as ApplicationManifest,
+});
+```
+
+<Note>
+
+Unlike the `ApolloClient` class from the core `@apollo/client` package, the `ApolloClient` class from `@apollo/client-ai-apps` doesn't require a configured `link`. By default, the class uses `ToolCallLink`, which executes GraphQL queries through MCP tools. <br/><br/>
+
+You can provide a custom link chain to grant additional capabilities to `ApolloClient`; however, your terminating link needs to be a `ToolCallLink`.
+
+</Note>
+
+### Provide your client to `ApolloProvider`
+
+After you've created your `client` instance, pass it to `ApolloProvider`. As with `ApolloClient`, you need to use the `ApolloProvider` component exported from the `@apollo/client-ai-apps` package.
+
+```tsx title=src/main.tsx
+import { ApolloProvider } from "@apollo/client-ai-apps/react";
+
+const client = new ApolloClient({
+  // ...
+});
+
+createRoot(document.getElementById("root")!).render(
+  <ApolloProvider client={client}>
+    <App />
+  </ApolloProvider>,
+);
+```
+
+The `ApolloProvider` implementation from `@apollo/client-ai-apps/react` initializes your application with Apollo MCP Server and provides the `client` instance in React context for use with Apollo Client's data-fetching hooks.
+
+#### Show a loading fallback during initialization
+
+The `ApolloProvider` component uses React's [Suspense](https://react.dev/reference/react/Suspense) functionality during initialization, which avoids rendering application code that relies on client initialization. The client is initialized when it receives the `ui/notifications/tool-result` notification from the host.
+
+If you want to display a loading fallback while the app initializes, wrap `ApolloProvider` with React's [`Suspense` component](https://react.dev/reference/react/Suspense).
+
+```tsx
+import { Suspense } from "react";
+
+createRoot(document.getElementById("root")!).render(
+  <Suspense fallback={<LoadingFallback />}>
+    <ApolloProvider client={client}>
+      <App />
+    </ApolloProvider>
+  </Suspense>,
+);
+```
+
+<Note>
+
+Using a `Suspense` component to display a loading fallback is optional. If you don't provide a `Suspense` component, the screen remains blank until the host provides the tool result.
+
+</Note>
+
+## Register an MCP tool
+
+To register MCP tools, combine your GraphQL queries with the `@tool` directive.
+
+For example, if you want to define a query that gets the highest-rated products in a marketplace application, define a query called `TopProductsQuery` in your React component, then use the `useQuery` hook from `@apollo/client/react` to read the query data. To register the query as its own tool, use the `@tool` directive.
+
+```tsx
+import { gql, TypedDocumentNode } from "@apollo/client";
+import { useQuery } from "@apollo/client/react";
+
+// Note: These types should be generated using a tool like GraphQL Codegen.
+// Avoid writing them by hand.
+const TOP_PRODUCTS_QUERY: TypedDocumentNode<
+  TopProductsQuery,
+  TopProductsQueryVariables
+> = gql`
+  query TopProductsQuery
+  @tool(
+    name: "TopProducts"
+    description: "Shows a list of the highest rated products."
+  ) {
+    topProducts {
+      id
+      title
+      rating
+      price
+      thumbnail
+    }
+  }
+`;
+
+function App() {
+  const { data, loading, error, dataState } = useQuery(TOP_PRODUCTS_QUERY);
+
+  if (loading) {
+    return <div>Loading...</div>;
+  }
+
+  if (error) {
+    return <div>Error! {error.message}</div>;
+  }
+
+  return (
+    <div>
+      {data.topProducts.map((product) => (
+        <Product key={product.id} product={product} />
+      ))}
+    </div>
+  );
+}
+```
+
+Now that the query is registered as a tool, you can run the app in ChatGPT or an MCP Apps-compatible host and prompt the LLM to show you the top products. They render using data fetched by MCP Server.
+
+## Platform-specific modules
+
+To create platform-specific modules, append the file extension that corresponds with the host you want to target.
+
+- `*.openai.ts(x)` - Code specific to a ChatGPT app
+- `*.mcp.ts(x)` - Code specific to an MCP app
+
+Those extensions allow you to write code only available to the targeted host.
+
+For example, the `@apollo/client-ai-apps/openai` entry point provides a `useWidgetState` hook that is only available to ChatGPT apps. To use that hook, create a component file with the `.openai.tsx` extension.
+
+```tsx title=MyComponent.openai.tsx
+import { useWidgetState } from "@apollo/client-ai-apps/openai";
+
+export function MyComponent() {
+  const [widgetState, setWidgetState] = useWidgetState();
+
+  return (
+    <div>
+      <div>{widgetState.foo.toString()}</div>
+      <button onClick={() => setWidgetState({ foo: true })}>Set foo</button>
+    </div>
+  );
+}
+```
+
+<Tip>
+
+For additional type safety when developing a platform-specific module, please reference the [TypeScript configuration](#typescript-configuration) section to update your configuration.
+
+</Tip>
+
+### How to import platform-specific modules
+
+To import platform-specific modules, use the base name. For example, if you create a platform-specific `Home` component for OpenAI (`Home.openai.tsx`) and MCP (`Home.mcp.tsx`), you import it like this:
+
+```tsx
+import { Home } from "./Home";
+```
+
+<Caution>
+
+Make sure each configured environment has a module it can import. If one doesn't exist, the application fails to build. For example, if you create a `Home.openai.tsx` file, you also need to create either a `Home.mcp.tsx` or `Home.tsx` file.
+
+</Caution>
+
+## TypeScript configuration
+
+`@apollo/client-ai-apps` provides extendable TypeScript configurations that make sure you only use code available to your application from [platform-specific modules](#platform-specific-modules).
+
+- `@apollo/client-ai-apps/tsconfig/core` — for shared modules that run in both environments. This config ensures that included files only use shared utilities available to all environments. Use this when type-checking shared files for all environments..
+- `@apollo/client-ai-apps/tsconfig/mcp` — For developing MCP-specific modules. This configuration allows you to access utilities exported by `@apollo/client-ai-apps/mcp`. Use this when type-checking your `.mcp.*` files.
+- `@apollo/client-ai-apps/tsconfig/openai` — For developing ChatGPT-specific modules. This configuration allows you to use utilities exported by `@apollo/client-ai-apps/openai`. Use this when type-checking your `.openai.*` files.
+
+### Recommended TypeScript configuration
+
+Use TypeScript [project references](https://www.typescriptlang.org/docs/handbook/project-references.html) to apply a TypeScript configuration to a subset of files. We recommend these configuration files:
+
+```
+.
+├── tsconfig.app.json
+├── tsconfig.base.json
+├── tsconfig.mcp.json
+├── tsconfig.openai.json
+└── tsconfig.json
+```
+
+**`tsconfig.base.json`**
+
+Shared configuration for all TypeScript files. This is be where you should configure most of your `compilerOptions`.
+
+```jsonc title=tsconfig.base.json
+{
+  "compilerOptions": {
+    "strict": true,
+    // ... all other preferred TypeScript settings
+  },
+}
+```
+
+**`tsconfig.app.json`**
+
+Configuration that targets all shared modules. Extend `@apollo/client-ai-apps/tsconfig/core` and exclude platform-specific modules (`.mcp.*` and `.openai.*` files).
+
+```jsonc title=tsconfig.app.json
+{
+  "extends": ["@apollo/client-ai-apps/tsconfig/core", "./tsconfig.base.json"],
+  "include": ["src"],
+  "exclude": ["src/**/*.mcp.*", "src/**/*.openai.*"],
+}
+```
+
+**`tsconfig.mcp.json`**
+
+Configuration that targets MCP app specific modules. Extend `@apollo/client-ai-apps/tsconfig/mcp` and include `.mcp.*` files.
+
+```jsonc title=tsconfig.mcp.json
+{
+  "extends": ["@apollo/client-ai-apps/tsconfig/mcp", "./tsconfig.base.json"],
+  "include": ["src/**/*.mcp.ts", "src/**/*.mcp.tsx"],
+}
+```
+
+<Note>
+
+If your app doesn't have any MCP Apps-specific files, don't include this file because TypeScript might be unable to match files in the `include` setting.
+
+</Note>
+
+**`tsconfig.openai.json`**
+
+Configuration that targets ChatGPT App-specific modules. Extend `@apollo/client-ai-apps/tsconfig/openai` and include `.openai.*` files. Include the `openai/globals` types in the `types` configuration, which provides types for `window.openai`.
+
+```jsonc title=tsconfig.openai.json
+{
+  "extends": ["@apollo/client-ai-apps/tsconfig/openai", "./tsconfig.base.json"],
+  "compilerOptions": {
+    "types": ["@apollo/client-ai-apps/openai/globals"],
+  },
+  "include": ["src/**/*.openai.ts", "src/**/*.openai.tsx"],
+}
+```
+
+<Note>
+
+If your app doesn't have any ChatGPT App-specific files, don't include this file because TypeScript might complain that it is unable to match files in the `include` setting.
+
+</Note>
+
+**`tsconfig.json`**
+
+The base configuration that should include all project references.
+
+```jsonc title=tsconfig.openai.json
+{
+  "files": [],
+  "references": [
+    { "path": "./tsconfig.app.json" },
+    // Omit if you don't create this file
+    { "path": "./tsconfig.mcp.json" },
+    // Omit if you don't create this file
+    { "path": "./tsconfig.openai.json" },
+  ],
+}
+```
+
+## Vite configuration
+
+The `@apollo/client-ai-apps` package includes a [Vite](https://vite.dev/) plugin that configures your application for developing ChatGPT and MCP apps.
+
+To use it, import `apolloClientAiApps` from `@apollo/client-ai-apps/vite` and provide it to the `plugins` option in your Vite configuration. Set the `targets` option to define what environments you are building for.
+
+- `mcp` - Creates a build artifact for MCP apps.
+- `openai` - Creates a build artifact for ChatGPT apps.
+
+```ts title=vite.config.ts
+import { defineConfig } from "vite";
+import { apolloClientAiApps } from "@apollo/client-ai-apps/vite";
+
+export default defineConfig({
+  plugins: [
+    apolloClientAiApps({
+      targets: ["mcp", "openai"],
+    }),
+    // ... other plugins
+  ],
+  // ... other config
+});
+```
+
+### Development mode
+
+You can only run one environment in development mode at a time. Set the `devTarget` option to define what environment you are developing for.
+
+```ts
+apolloClientAiApps({
+  devTarget: "openai",
+  // ...
+});
+```
+
+<Note>
+
+If the configured `targets` option only includes one value, you can omit the `devTarget` option because the plugin infers the `devTarget` as the value in `targets`.
+
+</Note>
+
+#### Use environment variables to set `devTarget`
+
+If you are developing both ChatGPT and MCP apps, switching the `devTarget` each time you want to develop for a different environment is cumbersome. We recommend that you use environment variables to configure those kinds of dynamic values so that you can easily change the environment without having to edit the build configuration.
+
+```ts
+apolloClientAiApps({
+  devTarget: process.env.DEV_TARGET,
+  // ...
+});
+```
+
+However, this results in a TypeScript error. `devTarget` only accepts the values `mcp` or `openai`, but `process.env.DEV_TARGET` is a `string` type. To fix this issue, use the `devTarget` helper exported from `@apollo/client-ai-apps/vite` which validates the input and returns the correctly typed value.
+
+```ts
+import { apolloClientAiApps, devTarget } from "@apollo/client-ai-apps/vite";
+
+apolloClientAiApps({
+  devTarget: devTarget(process.env.DEV_TARGET),
+  // ...
+});
+```
+
+<Tip>
+
+We recommend that you add separate development scripts in `package.json` that target each environment so you don't need to remember to set the environment variable.
+
+```json title=package.json
+{
+  "scripts": {
+    "dev:mcp": "DEV_TARGET=mcp vite",
+    "dev:openai": "DEV_TARGET=openai vite"
+  }
+}
+```
+
+</Tip>


docs/source/apps-sdk-architecture.mdx
```diff
@@ -0,0 +1,96 @@
+---
+title: MCP Apps Architecture
+subtitle: How MCP Apps work with Apollo MCP Server
+---
+
+MCP Apps combine GraphQL operations with custom UI resources to create interactive experiences inside ChatGPT. Apollo MCP Server bridges your GraphQL APIs to OpenAI's Apps SDK.
```
andrewmcgivery: Not just ChatGPT. ChatGPT is what we targeted originally but what we're releasing targets both ChatGPT via Apps SDK, as well as _any_ host that supports MCP Apps (the spec)... which ironically will also include ChatGPT, apparently. 😅

docs/source/apps-sdk-architecture.mdx
```diff
@@ -0,0 +1,96 @@
...
+    MCPServer -->|"4. Render in App UI"| App
+    App -->|"5. Display in ChatGPT"| ChatGPT
+    
+    style ChatGPT fill:#E3F2FD
+    style MCPServer fill:#1976D2,color:#fff
+    style GraphQL fill:#00ACC1,color:#fff
+    style App fill:#4CAF50,color:#fff
+```
+
+When ChatGPT invokes a tool, Apollo MCP Server executes the GraphQL operation, injects the results into your UI resource, and returns the rendered UI to ChatGPT.
```
andrewmcgivery: This, and the diagram above, isn't accurate.

The MCP Server tool executions do not return the UI resource, they only return the data. The host pre-fetches the UI resource as part of initial discovery of your app and on tool calls creates the iframe with the UI and injects the data from the MCP call into the iframe.

docs/source/apps-sdk-architecture.mdx
```diff
@@ -0,0 +1,96 @@
...
+
+## App components
+
+An MCP App consists of:
+
+- **Tools**: GraphQL operations exposed as MCP tools that ChatGPT can invoke
+- **Resources**: UI files (HTML, React apps, or remote URLs) that render the data
```
andrewmcgivery: This bit might be a bit confusing because with our approach, there is only one resource per app.

docs/source/apps-sdk-architecture.mdx
```diff
@@ -0,0 +1,96 @@
...
+
+The app manifest (`.application-manifest.json`) defines the mapping between tools, operations, and resources.
```
andrewmcgivery: I would maybe have this as "the resource" instead of "resources" for the same reason.

docs/source/apps-sdk-architecture.mdx
```diff
@@ -0,0 +1,96 @@
...
+    GraphQL -->|"Returns Data"| Resource
+    Resource -->|"Renders UI"| ChatGPT["ChatGPT Interface"]
+    
+    style App fill:#4CAF50,color:#fff
+    style Manifest fill:#FF9800,color:#fff
+    style GraphQL fill:#00ACC1,color:#fff
+    style ChatGPT fill:#E3F2FD
+```
+
+Apps are organized in directories under `apps/`, each containing a manifest file and UI resources. The manifest defines GraphQL operations, tool mappings, and resource paths.
```
andrewmcgivery: Again, I'd be careful with how we use the word "Resource" vs "resources"... Resource means something very specific in MCP-world.

docs/source/apps-sdk-architecture.mdx
```diff
@@ -0,0 +1,96 @@
...
+
+### Data access
```
andrewmcgivery: I probably would not include this section because

1. It's not accurate
2. It's different between Apps SDK vs MCP Apps
3. It's an implementation detail that we are purposefully abstracting away

docs/source/apps-sdk-architecture.mdx
```diff
@@ -0,0 +1,96 @@
...
+- Main result: `window.toolResult.result.data`
+- Prefetch results: `window.toolResult.prefetch.prefetchID`
+
+Prefetch operations run automatically when any tool in the app is invoked, useful for loading common data or user context.
+
+## Resource types
+
+Resources can be:
+- **Local HTML files**: Stored in your app directory
+- **React apps**: Built with Vite or other tools (recommended via [template](https://github.com/apollographql/ai-apps-template))
```
andrewmcgivery: I wouldn't differentiate between local html file and react app. The react app gets exposed as a local HTML file so these are really the same thing.

I would more document this as the React app can be a local index.html file or hosted remotely via a CDN and in either case, we recommend using our template.

docs/source/apps-sdk-architecture.mdx
```diff
@@ -0,0 +1,96 @@
...
+- **Remote URLs**: Hosted elsewhere and referenced by URL
+
+## App target selection
+
+Apollo MCP Server supports two app targets:
+- **AppsSDK** (default): For OpenAI ChatGPT apps
+- **MCPApps**: For MCP-compatible clients
```
andrewmcgivery: > For MCP-compatible clients

This is a little misleading. It isn't all MCP-compatible clients, but specifically MCP Apps compatible clients.

docs/source/_sidebar.yaml
```diff
@@ -40,6 +40,20 @@ items:
     href: "./licensing"
   - label: "Limitations"
     href: "./limitations"
+  - label: "MCP Apps"
+    children:
+      - label: "Overview"
+        href: "./apps-sdk"
```
andrewmcgivery: Bit of a meta note... let's not call this apps-sdk even in the URL because likely Apps-SDK is going to become the lesser of the 2 use case. I'd rather call it "mcp-apps".

docs/source/apps-sdk-prerequisites.mdx
```diff
@@ -0,0 +1,34 @@
+---
+title: MCP Apps Prerequisites
+subtitle: What you need to get started
+---
+
+Before building ChatGPT apps with Apollo MCP Server, ensure you have the following prerequisites.
```
andrewmcgivery: We support both Apps SDK and MCP Apps. We shouldn't make this super specific to Apps SDK.

docs/source/apps-sdk-prerequisites.mdx
```diff
@@ -0,0 +1,34 @@
...
+
+## Required prerequisites
+
+- Apollo MCP Server
+- GraphQL API endpoints that you want to expose through MCP Apps
+- ChatGPT Plus account with OpenAI Apps SDK access (currently in preview/beta)
```
andrewmcgivery: I would probably make this a note that you need this _if_ you want to test it end-to-end within the ChatGPT native interface... but you technically don't have to because you can use MCPJam which comes in our template.

docs/source/apps-sdk-prerequisites.mdx
```diff
@@ -0,0 +1,34 @@
...
+
+</Note>
+
+## Template-specific prerequisites
+
+If you're using the [Apollo AI Apps Template](https://github.com/apollographql/ai-apps-template), you'll also need:
+
+- Node.js v18 or later
+- npm or yarn
+- Tunneling tool (for example, [ngrok](https://ngrok.com/)) for local development with ChatGPT
```
andrewmcgivery: Only _if_ they want to test in the native ChatGPT interface. I would probably also make this a note because this isn't required for local development... and also isn't necessarily specific to our template so this might make more sense in the above section.

docs/source/apps-sdk-quickstart.mdx
```diff
@@ -0,0 +1,245 @@
...
+
+This guide walks you through creating your first ChatGPT app using the [Apollo AI Apps Template](https://github.com/apollographql/ai-apps-template). The template provides a complete setup with React, Vite, Apollo Client, and Apollo MCP Server integration.
+
+## Prerequisites
+
+Before starting, ensure you have all the [MCP Apps Prerequisites](/apollo-mcp-server/apps-sdk-prerequisites) in place, including:
+
+- Apollo MCP Server
+- A GraphQL API endpoint
+- ChatGPT with OpenAI Apps SDK access (Plus plan or higher)
```
andrewmcgivery: Optional, see previous comments

docs/source/apps-sdk-quickstart.mdx
```diff
@@ -0,0 +1,245 @@
...
+- Node.js v18 or later (for the template)
+- npm or yarn (for the template)
+- ngrok or similar tunneling tool (for accessing from ChatGPT)
```
andrewmcgivery: Optional, see previous comments

docs/source/apps-sdk-quickstart.mdx
```diff
@@ -0,0 +1,245 @@
...
+
+## Customizing your app
+
+### Modify the React app
+
+Edit files in `dev/the-store/` to customize your app's UI and behavior. The app uses:
+- **React** for the UI framework
+- **Vite** for building and bundling
+- **Apollo Client** for GraphQL integration
+- **Apollo MCP Server** for ChatGPT integration
```
andrewmcgivery: I would reword this to be more "provide tools and resources over mcp" than "for ChatGPT integration".

docs/source/apps-sdk-quickstart.mdx
```diff
@@ -0,0 +1,245 @@
...
+
+### Update GraphQL operations
+
+Modify the GraphQL operations in your app's manifest (located in `apps/`) to query different data or add new tools.
```
andrewmcgivery: The `/apps` folder is generated... and the user should never be manually editing the application manifest file.

The application manifest file is generated by Vite. If they want to update the operations, they update their React code.

docs/source/apps-sdk-reference.mdx
```diff
@@ -0,0 +1,75 @@
+---
+title: MCP Apps Reference
+subtitle: Key concepts and configuration
+---
+
+## App structure
+
+A ChatGPT app consists of a manifest file (`.application-manifest.json`), UI resources, and GraphQL operations. Apps are organized in directories under `apps/`, each containing a manifest and UI files.
```
andrewmcgivery: "A ChatGPT app"

This is incorrect/incomplete... we support Apps SDK or MCP Apps... which for now we're just calling generally "MCP Apps" for lack of a better overarching term.

docs/source/apps-sdk-reference.mdx
```diff
@@ -0,0 +1,75 @@
...
+
+## App manifest
```
andrewmcgivery: It may be worth documenting here (and possibly other pages) that this file is _generated_ and should not be manually updated.

docs/source/apps-sdk-reference.mdx
```diff
@@ -0,0 +1,75 @@
+---
+title: MCP Apps Reference
```
andrewmcgivery: Most of this page should be a series of tables, like we discussed in our meeting, to lay out what each element is, where it is mapped from (E.g. from an operation/directive or from package.json), and notes about what it maps to on the MCP/ChatGPT side.

docs/source/apps-sdk-reference.mdx
```diff
@@ -0,0 +1,75 @@
...
+The manifest defines:
+- GraphQL operations and their associated MCP tools
+- Resource path (local HTML, React app URL, or remote URL)
+- Configuration (CSP, widget settings, labels)
+
+Required fields: `format` (`"apollo-ai-app-manifest"`), `version` (`"1"`), `hash`, `resource`, and `operations`.
+
+## Resources
+
+Resources can be local HTML files, React apps (via the [template](https://github.com/apollographql/ai-apps-template)), or remote URLs.
```
andrewmcgivery: Similar note here as previous pages. This differentiation doesn't make sense.

docs/source/apps-sdk-reference.mdx
```diff
@@ -0,0 +1,75 @@
...
+A ChatGPT app consists of a manifest file (`.application-manifest.json`), UI resources, and GraphQL operations. Apps are organized in directories under `apps/`, each containing a manifest and UI files.
+
+## App manifest
+
+The manifest defines:
+- GraphQL operations and their associated MCP tools
+- Resource path (local HTML, React app URL, or remote URL)
+- Configuration (CSP, widget settings, labels)
+
+Required fields: `format` (`"apollo-ai-app-manifest"`), `version` (`"1"`), `hash`, `resource`, and `operations`.
```
andrewmcgivery: This isn't helpful, this is the kind of stuff that should be in tables.

docs/source/apps-sdk-reference.mdx
```diff
@@ -0,0 +1,75 @@
...
+
+## Resources
+
+Resources can be local HTML files, React apps (via the [template](https://github.com/apollographql/ai-apps-template)), or remote URLs.
+
+## Tools and operations
+
+Each app defines tools that ChatGPT can invoke. Tools are backed by GraphQL operations. Each operation includes:
```
andrewmcgivery: Not just ChatGPT

docs/source/apps-sdk-reference.mdx
```diff
@@ -0,0 +1,75 @@
...
+- `body`: The GraphQL query or mutation
+- `tools`: Array of tool definitions (name, description, optional extraInputs and labels)
+- `prefetchID` (optional): Runs as a prefetch for all tools
+
+## Accessing data in your UI
```
andrewmcgivery: I'd remove this section for similar reasons as the other page.

docs/source/apps-sdk-reference.mdx
```diff
@@ -0,0 +1,75 @@
...
+
+## App query parameter
+
+Use the `app` query parameter to specify which app to load: `https://your-server.com/mcp?app=my-app`. The app name corresponds to the directory name in `apps/`.
+
+## App target
+
+Apollo MCP Server supports two targets: **AppsSDK** (default, for ChatGPT) and **MCPApps** (for MCP-compatible clients). Specify via `appTarget` query parameter.
+
+## Content Security Policy (CSP)
```
andrewmcgivery: This section is incomplete and should be in one of the tables I mentioned above.

docs/source/apps-sdk-reference.mdx
```diff
@@ -0,0 +1,75 @@
...
+
+CSP settings control external resource access. Configure in manifest:
+- `connect_domains`: Network requests (fetch, XHR, WebSocket)
+- `resource_domains`: Resources (images, scripts, stylesheets)
+- `frame_domains`: Iframe embeds
+- `redirect_domains`: Redirects (AppsSDK only)
+
+## Widget settings
+
+Control app appearance in ChatGPT: `description`, `domain`, `prefersBorder`.
```
andrewmcgivery: table

docs/source/apps-sdk-reference.mdx
```diff
@@ -0,0 +1,75 @@
...
+
+## Tool invocation labels
+
+Customize user-facing messages during tool execution via `toolInvocation/invoking` and `toolInvocation/invoked` labels. Set at app or tool level.
```
andrewmcgivery: table

docs/source/apps-sdk-reference.mdx
```diff
@@ -0,0 +1,75 @@
...
+
+## Configuration
+
+Configure the apps directory path in `mcp-config.yaml`:
+
+```yaml
+apps:
+  path: ./my-apps
```
andrewmcgivery: This config doesn't exist 😅

docs/source/apps-sdk-reference.mdx
```diff
@@ -0,0 +1,75 @@
...
+```
+
+Apps automatically reload when manifest or resource files change. Multiple apps can coexist in the `apps/` directory, each accessible via its own `?app=app-name` parameter.
```
andrewmcgivery: This is a bit misleading. It's not that the apps reload, its that the MCP Server reloads (because of the script we have setup in the template repo).

Making changes to your app triggers vite... which re-creates the manifest file... which then triggers the shell script to restart the MCP server.

docs/source/apps-sdk-troubleshooting.mdx
```diff
@@ -0,0 +1,66 @@
+---
+title: MCP Apps Troubleshooting
+subtitle: Common issues and solutions
+---
+
+## App not appearing in ChatGPT
```
andrewmcgivery: Instead of ChatGPT here maybe say host?

docs/source/apps-sdk-troubleshooting.mdx
```diff
@@ -0,0 +1,66 @@
...
+
+**Common causes:**
+- Missing `?app=app-name` query parameter in URL
+- Apps feature not enabled in server build
```
andrewmcgivery: This won't be true when we launch. The app feature thing is because we were providing an experimental build with that internal feature flag.

docs/source/apps-sdk-troubleshooting.mdx
```diff
@@ -0,0 +1,66 @@
...
+- Services not running (GraphQL API, React app, MCP server)
+- Invalid manifest or missing required fields
```
andrewmcgivery: I wouldn't even include this. The user doesn't modify the manifest file, it's generated, so this should never happen.

docs/source/apps-sdk-troubleshooting.mdx
```diff
@@ -0,0 +1,66 @@
...
+
+**Solutions:**
+- Include app name in URL: `https://your-url/mcp?app=the-store`
+- Verify all services are running (if using template)
+- Check server logs for app loading errors
+- Ensure manifest is valid JSON with required fields (`format`, `version`, `hash`, `resource`, `operations`)
```
andrewmcgivery: Same here

docs/source/apps-sdk-troubleshooting.mdx
```diff
@@ -0,0 +1,66 @@
...
+
+## Manifest errors
```
andrewmcgivery: I wouldn't include this section for the same reason

docs/source/apps-sdk-troubleshooting.mdx
```diff
@@ -0,0 +1,66 @@
...
+
+**Symptoms**: Resources fail to load, network requests blocked, console CSP errors.
+
+**Solutions**: Configure CSP in manifest with required domains (`connect_domains`, `resource_domains`, `frame_domains`). Ensure domains include protocol (`https://`) and no trailing slashes. List all domains your app needs.
+
+## Tool execution problems
+
+**Tool executes but returns errors**: Verify GraphQL endpoint is correct and accessible, check authentication headers if needed, test operation directly against API, ensure operation matches schema.
+
+**Tool result not accessible**: Use correct paths: `window.toolResult.result.data` for main result, `window.toolResult.prefetch.prefetchID` for prefetch. Check for GraphQL errors in `window.toolResult.result.errors`.
```
andrewmcgivery: This doesn't make sense, these aren't correct, and even if they were, the user doesn't access these directly.

docs/source/apps-sdk-troubleshooting.mdx
```diff
@@ -0,0 +1,66 @@
...
+
+## Hot reload issues
+
+**Changes not reflected**: Verify file is saved, JSON is valid, server is watching apps directory. Restart server if hot reload isn't working.
```
andrewmcgivery: I'd be more specific here to stop and restart the start mcp script, if using our template.

docs/source/apps-sdk-troubleshooting.mdx
```diff
@@ -0,0 +1,66 @@
...
+
+- **Enable debug logging**: Set `logging.level: debug` in config
+- **Use MCP Inspector**: Test with `npx @modelcontextprotocol/inspector "http://127.0.0.1:8000/mcp?app=the-store" --transport http` (include app parameter)
+- **Check server logs**: Look for app loading messages and errors
+- **Validate manifest**: Use `jq` to validate JSON syntax and structure
+- **Test GraphQL operations**: Test operations directly against your API
+
+## Getting help
+
+If issues persist: check server logs, validate manifest, test with MCP Inspector, review the [MCP Apps Reference](/apollo-mcp-server/apps-sdk-reference), check [GitHub issues](https://github.com/apollographql/apollo-mcp-server/issues), or join the [Apollo community forums](https://community.apollographql.com/c/mcp-server/41).
```
andrewmcgivery: If you're going to link to the Apps SDK docs anywhere, you should also link to MCP Apps documentation.

docs/source/apps-sdk.mdx
```diff
@@ -0,0 +1,81 @@
+---
+title: MCP Apps Overview
+subtitle: Build custom, interactive experiences in ChatGPT
+---
+
+Apollo MCP Server supports [OpenAI's Apps SDK](https://developers.openai.com/apps-sdk) through MCP Apps, enabling you to build custom, interactive experiences directly inside ChatGPT. This integration allows you to control how results are displayed in ChatGPT, tailoring the look, feel, and interactivity of responses generated from your GraphQL APIs and agentic workflows.
```
andrewmcgivery: We support the Apps SDK AND MCP Apps. They are not the same thing, so this entire paragraph is misleading.

docs/source/apps-sdk.mdx
```diff
@@ -0,0 +1,81 @@
+---
+title: MCP Apps Overview
+subtitle: Build custom, interactive experiences in ChatGPT
```
andrewmcgivery: This entire page probably needs a review to not say specifically "ChatGPT" because again, thats only 50% of the story here.

docs/source/index.mdx
```diff
@@ -118,6 +118,8 @@ With Apollo MCP Server, these GraphQL advantages become immediately accessible t
 
 - **🚀 Connect AI to GraphQL in Minutes**. Developers can expose existing or new GraphQL API operations to AI clients without building complex custom integrations. By translating GraphQL functionalities into standardized MCP tools, Apollo MCP Server can significantly reduce the effort needed to connect AI to diverse data sources.
 
+- **🎨 Build Custom ChatGPT Experiences**. With MCP Apps support, you can create custom, interactive experiences directly inside ChatGPT. Control how your data appears with custom HTML/UI resources, enabling rich visualizations, interactive forms, and branded experiences powered by your GraphQL APIs.
```
andrewmcgivery: Same on this page... not just ChatGPT apps.

docs/source/mcp-apps-reference.mdx
```diff
@@ -0,0 +1,70 @@
+---
+title: MCP Apps Reference
+subtitle: Key concepts and configuration
+---
+
+The manifest file (`.application-manifest.json`) is automatically generated from your app code and `package.json`. Never edit the manifest directly. Instead, make changes to your app code and `package.json`, which will regenerate the manifest.
```
mabuyo: Try: Making "Never edit the manifest directly" as a new paragraph. If I'm skimming through, it's super easy for me to miss this.

docs/source/mcp-apps-reference.mdx
```diff
@@ -0,0 +1,75 @@
...
+| You define your app code and `package.json`, which generates the manifest | The host discovers your app via the MCP protocol and pre-fetches the UI resource |
+| You specify the resource URL (your UI) | The host pre-fetches the resource during discovery and creates an iframe when tools are invoked |
+| You define tools backed by GraphQL operations | The host invokes tools and receives data from the MCP server |
+| You access data via `window.toolResult.result.data` in your UI code | The host injects tool result data into the iframe via `window.toolResult` |
+
+## App-level configuration
+
+Configuration elements defined at the app level (from `package.json`):
+
+| Element | What it is | Mapped from | Maps to (MCP/ChatGPT) | Notes |
```
mabuyo: What is the reasoning behind the order of this table? Should we consider making it alphabetical?

docs/source/mcp-apps-reference.mdx
```diff
@@ -0,0 +1,75 @@
...
+| **Extra inputs** | Additional input parameters beyond GraphQL variables | Operation tool definition `extraInputs` array | Additional MCP tool input parameters | Must have unique names, cannot conflict with GraphQL variables |
+| **Tool-level labels** | Tool-specific invocation messages | Operation tool definition `labels` field | User-facing messages for this specific tool | Overrides app-level labels for this tool |
+
+## App query parameter
+
+Use the `app` query parameter to specify which app to load: `https://your-server.com/mcp?app=my-app`. The app name corresponds to the directory name in `apps/`.
+
+## App target
+
+Apollo MCP Server supports two targets: **AppsSDK** (default, for hosts supporting OpenAI's Apps SDK specification) and **MCPApps** (for MCP-compatible clients). Use the `appTarget` query parameter to specify the target.
```
mabuyo: Do: Add an example for what this would look like, similar to the `app` query parameter section above.

docs/source/mcp-apps-reference.mdx
```diff
@@ -0,0 +1,75 @@
...
+
+## Next steps
+
+- Follow the [Quickstart guide](/apollo-mcp-server/mcp-apps-quickstart) to build your first app
+- Check [Troubleshooting](/apollo-mcp-server/mcp-apps-troubleshooting) for common issues
+- Explore the [Apollo AI Apps Template](https://github.com/apollographql/ai-apps-template) for a complete setup
```
mabuyo: Personally I don't think a Reference page should need next steps 🤔

docs/source/mcp-apps-reference.mdx
```diff
@@ -0,0 +1,75 @@
+---
+title: MCP Apps Reference
+subtitle: Key concepts and configuration
```
mabuyo, line 3: Consider a more descriptive subtitle here, "key concepts" doesn't feel like it describes this particular reference page.

docs/source/mcp-apps-reference.mdx
```diff
@@ -0,0 +1,75 @@
...
+
+This reference guide provides a comprehensive overview of MCP Apps configuration elements. Use the tables below to understand what each element is, where it's configured in your app code or `package.json`, and how it maps to the MCP/ChatGPT side.
+
+## Manifest file
+
+The manifest file (`.application-manifest.json`) is automatically generated from your app code and `package.json`. Never edit the manifest directly. Instead, make changes to your app code and `package.json`, which will regenerate the manifest.
+
+## App vs Host
+
+This table shows what you work with in your app code versus what the host does:
```
mabuyo, line 141: Should the reader already understand what a "host" is referring to here?

docs/source/mcp-apps-reference.mdx
```diff
@@ -0,0 +1,75 @@
...
+| In Your App | In the Host |
+|-------------|-------------|
+| You define your app code and `package.json`, which generates the manifest | The host discovers your app via the MCP protocol and pre-fetches the UI resource |
+| You specify the resource URL (your UI) | The host pre-fetches the resource during discovery and creates an iframe when tools are invoked |
+| You define tools backed by GraphQL operations | The host invokes tools and receives data from the MCP server |
+| You access data via `window.toolResult.result.data` in your UI code | The host injects tool result data into the iframe via `window.toolResult` |
+
+## App-level configuration
+
+Configuration elements defined at the app level (from `package.json`):
```
mabuyo: If we preface with "from `package.json` here, do we need to have the column present in the table? The table is pretty wide already, so wondering if we can save some redundancy by removing the "Mapped from" column

docs/source/mcp-apps-reference.mdx
mabuyo: Same comment for other tables!

pragl: I was matching the hierarchical order of how a dev would come across these concepts as they're building. That said, for a reference document, I think you're right that alphabetical order makes more sense

pragl:
@andrewmcgivery, I think we're to a place where another round of technical review is warranted.

Please note:
- There's certainly more work I can do on the tech-writing side to polish everything, but I want to make sure that I'm working with correct content first.
- The diagrams are AI generated, and I could use guidance to get something really solid put together. (We can also just omit them from the initial publication and revisit later.)

docs/source/mcp-apps-architecture.mdx
```diff
@@ -0,0 +1,114 @@
...
+    
+    style Host1 fill:#E3F2FD
+    style Host2 fill:#E3F2FD
+    style MCPServer1 fill:#1976D2,color:#fff
+    style MCPServer2 fill:#1976D2,color:#fff
+    style GraphQL fill:#00ACC1,color:#fff
+    style Iframe fill:#4CAF50,color:#fff
+```
+
+When an MCP Apps-compatible host invokes a tool, Apollo MCP Server executes the GraphQL operation and returns only the data. The host then creates an iframe with the pre-fetched UI resource and injects the tool result data into it via `window.toolResult`.
```
andrewmcgivery: There are still references to `window.toolResult`. These should be completed removed as this does not exist.

docs/source/mcp-apps-architecture.mdx
```diff
@@ -0,0 +1,114 @@
...
+An MCP App consists of:
+
+- **Tools**: GraphQL operations exposed as MCP tools that MCP Apps-compatible hosts can invoke
+- **Resource**: The UI for your app, specified as a URL
+
+The app manifest (`.application-manifest.json`) is automatically generated from your app code and `package.json`. It defines the mapping between tools, operations, and the resource.
+
+## App structure
+
+```mermaid
```
andrewmcgivery: This diagram is really confusing to me... I'd probably just remove it tbh

docs/source/mcp-apps-architecture.mdx
```diff
@@ -0,0 +1,114 @@
...
+```
+
+Apps are organized in directories under `apps/`, each containing your app code, `package.json`, and the resource. The manifest file is automatically generated from these sources and defines GraphQL operations, tool mappings, and resource paths.
+
+
+
+
+## Resources
+
+The resource is the UI for your app, specified as a URL. This can point to a local file, a built React app, or a remote URL. We recommend using the [Apollo AI Apps Template](https://github.com/apollographql/ai-apps-template) for a complete setup.
```
andrewmcgivery: This is another instance of "a local file, a build react app, or a remote url"

A "build react app" IS a local file, so this is redundant. Anywhere we have this wording it should be just "a local built react app or a remotely hosted react app".

docs/source/mcp-apps-quickstart.mdx
```diff
@@ -0,0 +1,251 @@
...
+
+Edit files in `dev/the-store/` to customize your app's UI and behavior. The app uses:
+- **React** for the UI framework
+- **Vite** for building and bundling
+- **Apollo Client** for GraphQL integration
+- **Apollo MCP Server** for MCP Apps integration
+
+### Update GraphQL operations
+
+Modify the GraphQL operations in your app code or `package.json`. The manifest is automatically generated from those sources.
```
andrewmcgivery: > or `package.json`

This is incorrect. The operations do not come from the package.json

docs/source/mcp-apps-quickstart.mdx
```diff
@@ -0,0 +1,251 @@
...
+```
+
+## Development workflow
+
+### Local development
+
+For local development and testing:
+
+1. Run `npm run dev` in `dev/the-store/` (instead of `npm run dev:e2e`)
+2. Use a local emulator or test environment
```
andrewmcgivery, line 214: It's worth mentioning here that our template comes pre-bundled with MCP Jam for local emulation as to not confuse the reader into thinking they need to hook this up themselves.

docs/source/mcp-apps-reference.mdx
```diff
@@ -0,0 +1,80 @@
...
+An **app** is your MCP App—the code, configuration, and UI resource you build that combines GraphQL operations with custom UI. The **host** is the MCP Apps-compatible application (such as ChatGPT) that discovers your app, invokes tools, and displays your UI.
+
+This table shows what you work with in your app code versus what the host does:
+
+| In Your App | In the Host |
+|-------------|-------------|
+| You define your app code and `package.json`, which generates the manifest | The host discovers your app via the MCP protocol and pre-fetches the UI resource |
+| You specify the resource URL (your UI) | The host pre-fetches the resource during discovery and creates an iframe when tools are invoked |
+| You define tools backed by GraphQL operations | The host invokes tools and receives data from the MCP server |
+| You access data via `window.toolResult.result.data` in your UI code | The host injects tool result data into the iframe via `window.toolResult` |
```
andrewmcgivery: Again, these `window.*` references are incorrect and should be removed.

docs/source/mcp-apps-reference.mdx
```diff
@@ -0,0 +1,80 @@
...
+
+## App-level configuration
+
+Configuration elements defined at the app level (from `package.json`):
+
+| Element | What it is | Maps to (MCP/ChatGPT) | Notes |
```
andrewmcgivery: The `element` here should be the actual field names (dot noted) instead of englishy descriptions. Keep the englishy descriptions in the second column.

The "maps to" should also not be englishy. It should be the actual metadata name/path.

Overall these tables should be...

1. What is it?
2. Where does it come from?
3. Where does it map to?

docs/source/mcp-apps-reference.mdx
```diff
@@ -0,0 +1,80 @@
...
+| **CSP: redirect_domains** | Allowed domains for redirects | Content Security Policy for redirects (AppsSDK only) | AppsSDK target only |
+| **CSP: resource_domains** | Allowed domains for loading resources | Content Security Policy for images/scripts/stylesheets | Must include protocol (`https://`) |
+| **Labels: toolInvocation/invoked** | Message shown when tool completes | User-facing message after tool execution | Applies to all tools unless overridden |
+| **Labels: toolInvocation/invoking** | Message shown when tool starts | User-facing message during tool invocation | Applies to all tools unless overridden |
+| **Resource URL** | URL to your UI | UI resource that host pre-fetches during discovery | Can be local file path, built app URL, or remote URL |
+| **Widget: description** | Widget description | Widget metadata in host UI | AppsSDK target only |
+| **Widget: domain** | Domain restriction | Domain validation for widget | Restricts widget to specific domain |
+| **Widget: prefersBorder** | Border preference | Visual styling in host UI | Boolean flag for border display |
+
+## Operation-level configuration
```
andrewmcgivery: This doesn't make sense. There is no "operation configuration". The user defines operations using the `gql` tag and can then use `@prefetch` and `@tool` directives.

I'd highly suggest you try out our template to understand how this all fits together.

docs/source/mcp-apps-reference.mdx
```diff
@@ -0,0 +1,80 @@
...
+Configuration elements defined at the operation level (from GraphQL operations):
+
+| Element | What it is | Mapped from | Maps to (MCP/ChatGPT) | Notes |
+|---------|------------|-------------|----------------------|-------|
+| **Operation body** | GraphQL query or mutation | GraphQL operation definition | GraphQL operation executed by MCP server | Must be valid GraphQL matching your schema |
+| **Prefetch ID** | Identifier for prefetch data | Operation definition `prefetchID` field | Data available at `window.toolResult.prefetch.prefetchID` | Operation runs automatically when any tool in app is invoked |
+
+## Tool-level configuration
+
+Configuration elements defined at the tool level (from operation tool definitions):
```
andrewmcgivery: This only kind of makes sense... but it is a bit confusing cause it is written like the user is updating the manifest... and also like it's configuration, which it isn't.

It's `@tool` directives on an operation.

docs/source/mcp-apps.mdx
```diff
@@ -0,0 +1,82 @@
...
+
+## How MCP Apps work with Apollo MCP Server
+
+Apollo MCP Server bridges your GraphQL APIs to the MCP Apps specification, allowing you to:
+
+1. **Define GraphQL operations** that fetch data from your APIs
+2. **Create a UI resource** that controls how that data is displayed
+3. **Package them together** as apps that MCP Apps-compatible hosts can discover and use
+
+<Note>
```
jerelmiller: Would it make sense to make this a `<Tip />`?

docs/source/mcp-apps.mdx
```diff
@@ -0,0 +1,82 @@
+---
+title: MCP Apps Overview
+subtitle: Build custom, interactive experiences in MCP Apps-compatible hosts
+---
+
+Apollo MCP Server supports MCP Apps, enabling you to build custom, interactive experiences in any host that supports the MCP Apps specification (such as ChatGPT with OpenAI's Apps SDK). This integration allows you to control how results are displayed, tailoring the look, feel, and interactivity of responses generated from your GraphQL APIs and agentic workflows.
```
jerelmiller: Should we link to the specification doc?


```suggestion
Apollo MCP Server supports MCP Apps, enabling you to build custom, interactive experiences in any host that supports the [MCP Apps specification](https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/2026-01-26/apps.mdx) (such as ChatGPT with OpenAI's Apps SDK). This integration allows you to control how results are displayed, tailoring the look, feel, and interactivity of responses generated from your GraphQL APIs and agentic workflows.
```

docs/source/mcp-apps.mdx
jerelmiller, line 1: Do we want to add any sort of emphasis on the fact that it "feels" like building a GraphQL app with Apollo Client? I think that is a big selling point of this framework because it allows you to write MCP apps using the knowledge you already know building Apollo Client apps with some small modifications. Thoughts?

docs/source/_sidebar.yaml
```diff
@@ -40,6 +40,18 @@ items:
...
     href: "./limitations"
+  - label: "MCP Apps"
+    children:
+      - label: "Overview"
+        href: "./mcp-apps"
+      - label: "Prerequisites"
+        href: "./mcp-apps-prerequisites"
+      - label: "Architecture"
+        href: "./mcp-apps-architecture"
+      - label: "Quickstart"
```
jerelmiller, line 49: Quick note on ordering:
I'm used to seeing "Quickstart" as a 2nd or 3rd level item (3rd if there is an "installation" page). Here it feels so far down. Would it make sense to switch "Architecture" and "Quickstart" here? I think "Prerequisites" make sense before "Quickstart" so I'd leave that alone.

docs/source/mcp-apps-architecture.mdx
```diff
@@ -0,0 +1,84 @@
...
+When an MCP Apps-compatible host invokes a tool, Apollo MCP Server executes the GraphQL operation and returns only the data. The host then creates an iframe with the pre-fetched UI resource and injects the tool result data into it.
+
+## App components
+
+An MCP App consists of:
+
+- **Tools**: GraphQL operations exposed as MCP tools that MCP Apps-compatible hosts can invoke
+- **Resource**: The UI for your app, specified as a URL
+
+The app manifest (`.application-manifest.json`) is automatically generated from your app code and `package.json`. GraphQL operations come from your app code, while app-level configuration comes from `package.json`. The manifest defines the mapping between tools, operations, and the resource.
```
jerelmiller: As of the most recent release, this is now only partially true. Configuration _can_ come from `package.json`, but it's not the only place. Now that we are using `cosmiconfig`, we support the following places to define configuration:

- `.apollo-client-ai-apps.config.json`
- `apollo-client-ai-apps.config.json`
- `.apollo-client-ai-apps.config.yml`
- `apollo-client-ai-apps.config.yml`
- `.apollo-client-ai-apps.config.yaml`
- `apollo-client-ai-apps.config.yaml`
- `.apollo-client-ai-apps.config.js`
- `apollo-client-ai-apps.config.js`
- `.apollo-client-ai-apps.config.ts`
- `apollo-client-ai-apps.config.ts`
- `.apollo-client-ai-apps.config.cjs`
- `apollo-client-ai-apps.config.cjs`
- An `apollo-client-ai-apps` key in `package.json`

docs/source/mcp-apps-architecture.mdx
```diff
@@ -0,0 +1,84 @@
...
+
+MCP Apps work in two phases:
+
+**Discovery phase**: The host discovers your app and pre-fetches the UI resource.
+
+**Tool execution phase**: When a tool is invoked, the host receives data from the MCP server and injects it into a pre-loaded iframe.
+
+```mermaid
+graph TB
+    subgraph Discovery["Discovery Phase"]
```
jerelmiller, line 18: Unrelated to the content, I noticed this is unreadable in dark mode. Throwing this in here so I don't forget to ask about it.

Image

docs/source/mcp-apps-architecture.mdx
```diff
@@ -0,0 +1,84 @@
...
+- **Tools**: GraphQL operations exposed as MCP tools that MCP Apps-compatible hosts can invoke
+- **Resource**: The UI for your app, specified as a URL
+
+The app manifest (`.application-manifest.json`) is automatically generated from your app code and `package.json`. GraphQL operations come from your app code, while app-level configuration comes from `package.json`. The manifest defines the mapping between tools, operations, and the resource.
+
+## App structure
+
+
+
+Apps are organized in directories under `apps/`, each containing your app code, `package.json`, and the resource. The manifest file is automatically generated from these sources and defines GraphQL operations, tool mappings, and resource paths.
```
jerelmiller: We should probably distinguish that `apps/` is where the build artifacts are located. Development doesn't happen in `apps/`. This is a bit ambiguous, but a naive reader might assume that development is done in the `apps/` directory which is untrue.

Our AI template uses `dev/<app-name>` as the development directory, but that's not necessarily a requirement. 

@andrewmcgivery do we want to recommend `dev/<app-name>`, or leave it to the user to define that?

jerelmiller: @andrewmcgivery one other question... is the `apps/` directory a requirement of the MCP server? If so, we need to make that clear that it needs to be specified correctly in the `vite.config.ts` file to avoid issues.

docs/source/mcp-apps-architecture.mdx
```diff
@@ -0,0 +1,84 @@
...
+
+
+## Resources
+
+The resource is the UI for your app, specified as a URL. This can point to a local built React app or a remotely hosted React app. We recommend using the [Apollo AI Apps Template](https://github.com/apollographql/ai-apps-template) for a complete setup.
```
jerelmiller: qq: Are we talking about resources in the general sense as in the MCP "resource" type? If so, perhaps switch from `url` to `uri` (since an MCP server uses the `resourceUri` field, with a `ui://` scheme).

If you're talking about the resource from the standpoint of our apps SDK, we might want to tweak this a bit.

We have `resource` configuration specified in the manifest file which is defined either a) for you or b) set when you define an `entry` field in your configuration. That value can either be a URL or a path to the `.html` file.

Since this seems like a more general statement, perhaps we avoid mentioning "specified by a URL" here since thats only partly true in the apps sdk (and I would want to confuse anyone later). If anything, perhaps we mention the resource is the built `.html` file that contains the application code that will be loaded by the MCP host?

jerelmiller: @andrewmcgivery I welcome your thoughts on terminology as well

docs/source/mcp-apps-architecture.mdx
```diff
@@ -0,0 +1,84 @@
...
+
+## App target selection
+
+Apollo MCP Server supports two app targets:
+- **AppsSDK** (default): For hosts that support OpenAI's Apps SDK specification (for example, ChatGPT)
```
jerelmiller:
```suggestion
- **AppsSDK**: For hosts that support OpenAI's Apps SDK specification (for example, ChatGPT)
```

I don't believe we have a "default". You need to explicitly state which app target you're using.

docs/source/mcp-apps-architecture.mdx
```diff
@@ -0,0 +1,84 @@
...
+- **MCPApps**: MCP Apps-compatible clients
```
jerelmiller: qq: are `MCPApps` and `AppsSDK` meant to represent a value in our configuration or something else? I might provide a different recommendation here if thats the case, but I'd like to understand where these values come so I can give a proper suggestion if these need to change.

If we want code-specific values, we use `mcp` and `openai` as these two values. If we want to represent the environment, perhaps change this to "ChatGPT Apps" and "MCP Apps".

docs/source/mcp-apps-architecture.mdx
```diff
@@ -0,0 +1,84 @@
...
+
+The target determines MIME type, metadata format, and CSP format. For ChatGPT and other Apps SDK-compatible hosts, use the default AppsSDK target.
```
jerelmiller:
```suggestion
The target determines MIME type, metadata format, and CSP format.
```

Let's remove that last sentence since you need to specify the target and we have no default.

docs/source/mcp-apps-architecture.mdx
```diff
@@ -0,0 +1,84 @@
...
+
+## Next steps
+
+- Review the [MCP Apps Overview](/apollo-mcp-server/mcp-apps) for key concepts
```
jerelmiller: Do we expect the user to have read the overview at this point? If so, would it make sense to remove this?

docs/source/mcp-apps-quickstart.mdx
```diff
@@ -0,0 +1,254 @@
...
+## Prerequisites
+
+Before starting, ensure you have all the [MCP Apps Prerequisites](/apollo-mcp-server/mcp-apps-prerequisites) in place, including:
+
+- Apollo MCP Server
+- A GraphQL API endpoint
+- An MCP Apps-compatible host (this guide uses ChatGPT as an example)
+- Node.js v18 or later (for the template)
+- npm or yarn (for the template)
+- (Optional) ngrok or similar tunneling tool (for accessing from your host)
```
jerelmiller: How much do we want to lean on the AI apps template here? The template currently includes a GraphQL API as well as an install script to pull down the MCP server so the user doesn't really need these ahead of time.

That said, the GraphQL endpoint is for demo purposes and is definitely not something the user will want to keep long-term.

With that in mind, should we make it clear what the template includes? Otherwise someone might think they need to pull all these things together before they need to do anything.

For bare minimum running the demo app in the template, you need the following:

```suggestion
- An MCP Apps-compatible host (this guide uses ChatGPT as an example)
- Node.js v18 or later (for the template)
- npm or yarn (for the template)
- (Optional) ngrok or similar tunneling tool (for accessing from your host)
```

docs/source/mcp-apps-quickstart.mdx
```diff
@@ -0,0 +1,254 @@
...
+## Step 3: Understand the project structure
+
+The template creates a project with the following structure:
+
+```
+my-awesome-app/
+├── ecommerce-graph/          # Demo GraphQL API
+├── dev/
+│   └── the-store/            # Your React app (built with Vite)
+├── apps/                      # Apps directory for MCP Server
```
jerelmiller:
```suggestion
├── apps/                      # Output directory for the build artifacts
```

docs/source/mcp-apps-quickstart.mdx
```diff
@@ -0,0 +1,254 @@
...
+./start_mcp.sh
+```
+
+That command starts the Apollo MCP Server, which:
+- Loads your app from the `apps/` directory
+- Connects to your GraphQL API
+- Serves your React app as a resource
+- Exposes MCP tools to your host (for example, ChatGPT)
+
+## Step 5: Set up tunneling (for host access)
```
jerelmiller, line 106: We'll need one more thing here (which still needs updating in the template). There was a recent change in the MCP server that helps prevent DNS rebinding attacks, and this interferes with using `ngrok` with an MCP apps host (e.g. ChatGPT) because it will block requests. 

After the user starts `ngrok`, they will need to update `mcp-server.yaml` and set the ngrok domain as an allowed host. This is done under the `transport` option:

```yml
transport:
  type: streamable_http
  stateful_mode: false
  host_validation: 
    allowed_hosts:
      - example.ngrok.io # this needs to be set to the ngrok domain used when running ngrok

```

docs/source/mcp-apps-quickstart.mdx
```diff
@@ -0,0 +1,254 @@
...
+
+<Note>
+
+Make sure to note the HTTPS URL that ngrok provides. You'll need this in the next step.
+
+</Note>
+
+## Step 6: Connect to your host
+
+Follow your host's documentation to connect your MCP server. For this example using ChatGPT, follow the instructions in the [OpenAI documentation](https://developers.openai.com/apps-sdk).
```
jerelmiller: It might be best to use this link instead since it demonstrates how to connect your app to ChatGPT: https://developers.openai.com/apps-sdk/deploy/connect-chatgpt

The other doccumentation is more the entry point for general docs on building ChatGPT apps so it might take the user a while to figure out which doc to look at otherwise.

docs/source/mcp-apps-quickstart.mdx
```diff
@@ -0,0 +1,254 @@
...
+
+When adding your MCP server URL to your host (for example, ChatGPT) or MCP Inspector, use this format:
+
+```
+https://your-ngrok-url.ngrok-free.app/mcp?app=the-store
```
jerelmiller, line 152:
```suggestion
https://your-ngrok-url.ngrok-free.app/mcp?app=the-store&appTarget=openai
```

The `appTarget` is important so that our MCP server returns the right metadata 🙂

docs/source/mcp-apps-quickstart.mdx
```diff
@@ -0,0 +1,254 @@
...
+```
+
+**Important**: Include the `/mcp?app=the-store` query parameter. The `app` parameter tells the MCP server which app to load.
```
jerelmiller: We probably want to mention the `appTarget` here too

docs/source/mcp-apps-quickstart.mdx
jerelmiller, line 1: Argh, my bad, any comment I had regarding the default `openai` app target is correct here. I had thought it was required, but it is not. Ignore any comment I've made in that regard.

docs/source/mcp-apps-reference.mdx
```diff
@@ -0,0 +1,69 @@
+---
+title: MCP Apps Reference
+subtitle: Key concepts and configuration
+---
+
+This reference guide provides a comprehensive overview of MCP Apps configuration elements. Use the tables below to understand what each element is, where it's configured (GraphQL operations come from your app code, while app-level settings come from `package.json`), and how it maps to the MCP/ChatGPT side.
```
jerelmiller: Config now can be from other files, not just `package.json`. Perhaps we should list of the available locations on this page?

docs/source/mcp-apps-reference.mdx
```diff
@@ -0,0 +1,69 @@
...
+
+## Manifest file
+
+The manifest file (`.application-manifest.json`) is automatically generated from your app code and `package.json`. GraphQL operations come from your app code, while app-level configuration (like CSP settings, widget settings, and labels) comes from `package.json`.
+
+<warning>
+Don't edit the manifest directly. Instead, make changes to your app code (for GraphQL operations) and `package.json` (for app-level configuration). The manifest file generates automatically.
+</warning>
```
jerelmiller:
```suggestion
<Caution>
Don't edit the manifest directly. Instead, make changes to your app code (for GraphQL operations) and `package.json` (for app-level configuration). The manifest file generates automatically.
</Caution>
```

I think these might be `<Caution>` tags 🙂

docs/source/mcp-apps-reference.mdx
```diff
@@ -0,0 +1,69 @@
...
+
+## App-level configuration
+
+Configuration elements defined at the app level (from `package.json`):
+
+| Element | What it is | Where it comes from | Where it maps to | Notes |
+|---------|------------|---------------------|------------------|-------|
+| `name` | App identifier | `package.json` | MCP tool namespace | Defaults to directory name if not specified |
+| `description` | App description | `package.json` | Tool description prefix | Prefixed to tool descriptions |
+| `resource` | UI resource URL | `package.json` | Resource URI in manifest | Can be a local built React app or a remotely hosted React app |
```
jerelmiller:
```suggestion
| `entry` | UI resource URL | `package.json` | Resource URI in manifest | Can be a local built React app or a remotely hosted React app |
```

While we output the value as `resource` in the manifest file, we call it `entry` in the config.

docs/source/mcp-apps-reference.mdx
```diff
@@ -0,0 +1,69 @@
...
+This table shows what you work with in your app code versus what the host does:
+
+| In Your App | In the Host |
+|-------------|-------------|
+| You define your app code and `package.json`, which generates the manifest | The host discovers your app via the MCP protocol and pre-fetches the UI resource |
+| You specify the resource URL (your UI) | The host pre-fetches the resource during discovery and creates an iframe when tools are invoked |
+| You define tools backed by GraphQL operations | The host invokes tools and receives data from the MCP server |
+| You access data in your UI code | The host injects tool result data into the iframe |
+
+## App-level configuration
```
jerelmiller: Let's do a once over on this table. Here is what the config looks like as a TypeScript object with descriptions for what each field is (the same shape applies to the config if its in `package.json` as well):

```ts
const config: ApolloClientAiAppsConfig.Config = {
  // The name of your application. If not provided, this is set to the 'name'
  // property in package.json.
  name: "My App",

  // The description of your application. If not provided, this is set to the
  // "description" property in package.json
  description: "My app description",

  // The version of your application. This is reported as part of `appInfo`
  // during the 'ui/initialize' notification. If not provided, this set to the
  // "version" property in package.json.
  version: "1.0.0",

  // A map that configures the entry point for a mode. If not configured, the
  // entry points are configured for you. By default, 'development' and
  // 'production' are supported. Custom modes are supported by providing the
  // `--mode` option when running `vite`.
  //
  // An entrypoint can be configured either as a string or object with 'mcp' and
  // 'openai' keys. When a string is provided, that value is used for both the
  // 'mcp' and 'openai' environments.
  //
  // Defaults:
  // 'development' - Points to the url of the local vite development server
  // 'production' - The location of the `index.html` files created by the build
  entry: {
    // Entry point configuration for the 'development' mode (e.g. when running `vite`)
    development: "<url or location/of/index.html>",

    // Entry point configuration for the 'production' mode (e.g. when running `vite build`)
    production: {
      // Configures the 'mcp' environment entry point
      mcp: "...",
      // Configures the 'openai' environment entry point
      openai: "...",
    },

    // Custom entry points are allowed. Use the `--mode` option when running
    // `vite`.
    staging: "...",
  },

  // CSP settings for your application.
  csp: {
    // Approved origins for network requests (fetch/XHR/WebSocket). See MCP Apps
    // spec for detail.
    connectDomains: ["..."],

    // Approved origins for nested iframes (frame-src directive). See MCP Apps
    // spec for detail
    frameDomains: ["..."],

    // 'openai' environment only:
    // Enables ChatGPT a return link to the same conversation. ChatGPT will skip
    // the safe-link modal and append a redirectUrl query parameter to the
    // destination so you can route the user back into ChatGPT.
    redirectDomains: ["..."],

    // Approved origins for static resources (scripts, images, styles, fonts).
    resourceDomains: ["..."],
  },

  // Settings for the widget
  widgetSettings: {
    // visual boundary preference. See MCP Apps spec for detail
    prefersBorder: true,

    // 'openai' environment only
    // Let's the widget describe itself. See openai apps for reference:
    // https://developers.openai.com/apps-sdk/build/mcp-server
    description: "...",

    // Dedicated origin for the view. See MCP Apps spec for detail.
    domain: "...",
  },

  // 'openai' environment only
  // Configures custom text when the tool is invoking or invoked
  labels: {
    toolInvocation: {
      // Short status text while the tool runs
      invoking: "...",
      // Short status text after the tool completes
      invoked: "...",
    },
  },
};
```

docs/source/mcp-apps-reference.mdx
```diff
@@ -0,0 +1,69 @@
...
+| `csp.redirectDomains` | Allowed domains for redirects | `package.json` | `openai/widgetCSP.redirect_domains` (AppsSDK only) | AppsSDK target only |
+| `csp.baseUriDomains` | Allowed base URI domains | `package.json` | `_meta.ui.csp.baseUriDomains` (MCPApps only) | MCPApps target only |
+| `widgetSettings.description` | Widget description | `package.json` | `openai/widgetDescription` (AppsSDK only) | AppsSDK target only |
+| `widgetSettings.domain` | Domain restriction | `package.json` | `openai/widgetDomain` (AppsSDK)<br />`_meta.ui.domain` (MCPApps) | Restricts widget to specific domain |
+| `widgetSettings.prefersBorder` | Border preference | `package.json` | `openai/widgetPrefersBorder` (AppsSDK)<br />`_meta.ui.prefersBorder` (MCPApps) | Boolean flag for border display |
+| `labels.toolInvocation/invoking` | Message shown when tool starts | `package.json` | `openai/toolInvocation/invoking` (AppsSDK only) | Applies to all tools unless overridden |
+| `labels.toolInvocation/invoked` | Message shown when tool completes | `package.json` | `openai/toolInvocation/invoked` (AppsSDK only) | Applies to all tools unless overridden |
+
+## Tool-level configuration
```
jerelmiller, line 150: Let's also do a once-over on this table. Here are the available arguments available for the `@tool` directive:

```ts
type T = {
    name: string;
    description: string;
    extraInputs?: {
        name: string;
        description: string;
        type: "string" | "number" | "boolean";
    }[] | undefined;
    labels?: {
        toolInvocation?: {
            invoking?: string;
            invoked?: string;
        };
    } | undefined;
}
```

docs/source/mcp-apps-reference.mdx
```diff
@@ -0,0 +1,69 @@
...
+
+Tools are defined using the `@tool` directive on GraphQL operations in your app code. You can also use the `@prefetch` directive to mark operations that run automatically when any tool is invoked.
+
+| Element | What it is | Where it comes from | Where it maps to | Notes |
```
jerelmiller: I might suggest splitting this table up and instead:
- List the available directives as an unordered list
- Make the table the list of available arguments for `@tool`

See above for the reference on what arguments are available for `@tool`

PR #651, opened 2026-02-18
Support custom tool descriptions for manifest operations

### LINKED ISSUE #495: Question: How can we pass tool descriptions if tools are defined in manifest files?
Tool descriptions are extremely important for the agents. 
If there is no description on a tool, the agent will likely default to introspect or search instead of using a particular tool.
But i cannot see any way of passing descriptions if i define tools in a query manifest

DaleSeo:
Closes #495

Users loading operations from PQ manifests have no way to provide tool descriptions, so AI agents fall back to introspection or search to understand what each tool does. This PR adds a `descriptions` map to the `manifest` operation source config, letting users map operation names to tool descriptions. When a description is provided, it takes priority over auto-generated descriptions from schema types. Operations not listed in the map continue to use the existing auto-generation logic.

```yaml
operations:
  source: manifest
  path: ./persisted-queries-manifest.json
  descriptions:
    GetAlerts: "Get active weather alerts for a US state by its two-letter abbreviation"
    GetForecast: "Get a detailed weather forecast for a geographic coordinate"
```

## Testing

Tested with the following config file using MCP Inspector:

```yaml
transport:
  type: streamable_http
  stateful_mode: false
operations:
  source: manifest
  path: ./graphql/weather/persisted_queries/apollo.json
  descriptions:
    GetAlerts: "Test1 🥇"
    GetForecast: "Test2 🥈"
schema:
  source: local
  path: ./graphql/weather/api.graphql
```

2026-02-18 at 14 48 13

### DIFF
--- a/crates/apollo-mcp-server/src/operations/operation_source.rs
+++ b/crates/apollo-mcp-server/src/operations/operation_source.rs
@@ -227,14 +227,16 @@ impl From<ManifestSource> for OperationSource {
     }
 }
 
+#[allow(clippy::expect_used)]
+static OP_NAME_RE: std::sync::LazyLock<Regex> = std::sync::LazyLock::new(|| {
+    Regex::new(r"(?:query|mutation|subscription)\s+([A-Za-z_]\w*)")
+        .expect("OP_NAME_RE is a valid regex")
+});
+
 /// Extract the operation name from a GraphQL operation body using a lightweight
 /// regex, avoiding a full parse. Returns `None` for anonymous operations.
 fn extract_operation_name(source_text: &str) -> Option<&str> {
-    static OP_NAME_RE: std::sync::LazyLock<Option<Regex>> = std::sync::LazyLock::new(|| {
-        Regex::new(r"(?:query|mutation|subscription)\s+([A-Za-z_]\w*)").ok()
-    });
     OP_NAME_RE
-        .as_ref()?
         .captures(source_text)
         .and_then(|caps| caps.get(1))
         .map(|m| m.as_str())


DaleSeo:
I don't think any of AI review feedback needs addressing. I'm documenting my reasons for visibility here:

### 1. **[Should Fix]** LazyLock<Option<Regex>> is unnecessarily defensive

This project denies `clippy::expect_used` globally, so using `.expect()` would require an `#[expect(clippy::expect_used)]` override. I'd rather keep the `Option` pattern and stay consistent with the project's lint policy than add an exception for a one-liner.

https://github.com/apollographql/apollo-mcp-server/blob/498f23167b8c060c8d98c1b211d9eedc467fd4ea/Cargo.toml#L65

### 2. **[Consider]** HashMap cloning in closure

The `description` is moved into the closure via `move`, not cloned. `into_stream` consumes self, so there's just one ownership transfer. No `Arc` wrapping is needed.

### 3. **[Consider]** Test file naming complexity

This follows the same pattern as the existing tests in the same file. The timestamp helps prevent flaky failures during parallel test execution since cargo test runs tests concurrently by default.

docs/source/define-tools.mdx
```diff
@@ -115,6 +115,37 @@ operations:
   path: <PATH/TO/persisted-queries-manifest.json>
 ```
 
+#### Tool descriptions for manifest operations
+
+Tool descriptions help AI models understand when and how to use each tool. You can provide descriptions for operations in a persisted query manifest using either of the following methods.
+
+**Using GraphQL comments in the operation body** (recommended): Embed `#` comments before the operation definition in the `body` field. The MCP Server extracts leading comments and uses them as the tool description.
+
+```json title="Manifest with comment-based descriptions"
```
mabuyo:
```suggestion
```json title="Manifest with comment-based descriptions" {7}
```
Highlight line for better emphasis on where they should be focusing

docs/source/define-tools.mdx
```diff
@@ -115,6 +115,37 @@ operations:
...
+      "id": "f4d7c9e3...",
+      "body": "# Get active weather alerts for a US state\nquery GetAlerts($state: String!) { alerts(state: $state) { severity description instruction } }"
+    }
+  ]
+}
+```
+
+**Using config-level descriptions**: Add a `descriptions` map to your `operations` config that maps operation names to tool descriptions. These descriptions override auto-generated descriptions for the matching operations.
+
+```yaml title="Config with tool descriptions for manifest operations"
```
mabuyo:
```suggestion
```yaml title="Config with tool descriptions for manifest operations" {4-5}
```

crates/apollo-mcp-server/src/operations/operation_source.rs
```diff
@@ -38,17 +42,38 @@ pub enum OperationSource {
...
                     Event::OperationsUpdated(
-                        operations.into_iter().map(RawOperation::from).collect(),
+                        operations
+                            .into_iter()
+                            .map(|tuple| {
+                                let mut raw = RawOperation::from(tuple);
+                                if let Some(desc) = extract_operation_name(&raw.source_text)
+                                    .and_then(|name| descriptions.get(name))
+                                {
+                                    raw.description = Some(desc.clone());
```
andrewmcgivery, line 72: I agree with you that I think Claude got confused what this code is doing since what you're cloning here is the specific entry in `descriptions`, not the entire map itself.

So this clone feels fine to me!

crates/apollo-mcp-server/src/operations/operation_source.rs
```diff
@@ -195,10 +220,26 @@ impl OperationSource {
...
+        }
     }
 }
 
+/// Extract the operation name from a GraphQL operation body using a lightweight
+/// regex, avoiding a full parse. Returns `None` for anonymous operations.
+fn extract_operation_name(source_text: &str) -> Option<&str> {
+    static OP_NAME_RE: std::sync::LazyLock<Option<Regex>> = std::sync::LazyLock::new(|| {
+        Regex::new(r"(?:query|mutation|subscription)\s+([A-Za-z_]\w*)").ok()
+    });
```
andrewmcgivery: This is a bit of a tricky one. Regex is pretty rough in Rust.... ideally it would be compile-time checked but alas, here we are lol

I wanted to make sure we got this right so I got a second (non-AI) opinion on this. :) 

If you're going to use a `LazyLock`, it shouldn't be function scoped. It should be hoisted out of the function so that it is lazy evaluated one time, not every time the function is called.

That said, regarding the expect, this is a bit of an exception here. The intent of the expect lint is to prevent run-time panics, but as long as we cover this regex in tests, we can be sure the regex is valid and therefore will never panic. Because of that, suppressing the lint in this case would be appropriate.

If the regex was ever made invalid, debugging is going to be really difficult because we're going to have a `None` but no error message in tests.

So I'd suggest here that we should indeed use `expect` and suppress the lint for just this regex. :)

DaleSeo: Thanks for the thorough review, @andrewmcgivery! I've missed human feedback like this 😆 
Hoisted `OP_NAME_RE` to module level and switched to `expect` with a lint suppression. Agreed that swallowing the error with `.ok()` would make debugging painful.

andrewmcgivery (approved):

PR #657, opened 2026-02-20
Gate outputSchema and structuredContent on negotiated MCP protocol version

DaleSeo:
Fixes https://community.apollographql.com/t/enable-output-schema-breaks-claude-desktop-tool-execution/9721

When `enable_output_schema` is `true`, the server sends `outputSchema` in tool definitions and `structuredContent` in tool call responses. These features are from the 2025-06-18 version of MCP, but the server was advertising `2025-03-26`, which is the default for `rmcp`. Clients like Claude Desktop, which follow the negotiated protocol version, reject these fields as unexpected. This leads to an "Error occurred during tool execution" message for every tool call.

This PR updates the server to advertise `2025-06-18` when `enable_output_schema` is enabled. This way, clients that support the newer spec can negotiate it correctly. For clients using older protocol versions, `outputSchema` is removed from tool definitions in `list_tools`, and `structuredContent` is removed from responses in `call_tool`. Both checks are done per session based on the client's protocol version from `peer_info()`, allowing a single server to serve clients at different protocol levels correctly.

## Testing

With protocolVersion set to 2025-03-26, `outputSchema` does not exist from the response:

```sh
$ curl -s -X POST http://127.0.0.1:8000/mcp \
  -H 'Content-Type: application/json' \
  -H 'Accept: application/json, text/event-stream' \
  -H "Mcp-Session-Id: $SESSION" \
  -d '{"jsonrpc":"2.0","id":2,"method":"tools/list"}'
data: 
id: 0
retry: 3000

data: {"jsonrpc":"2.0","id":2,"result":{"tools":[{"name":"SearchUpcomingLaunches","description":"Fields searched - launch_designator, launch_service_provider__name, mission__name, name, pad__location__name, pad__name, rocket__configuration__manufacturer__abbrev, rocket__configuration__manufacturer__name, rocket__configuration__name, rocket__spacecraftflight__spacecraft__name. Codes are the best search terms to use. Single words are the next best alternative when you cannot use a code to search","inputSchema":{"type":"object","properties":{"query":{"type":"string"}},"required":["query"]},"annotations":{"readOnlyHint":true,"destructiveHint":false}},...
```

With protocolVersion set to 2025-06-18, `outputSchema` exists in the response:

```sh
$ curl -s -X POST http://127.0.0.1:8000/mcp \
  -H 'Content-Type: application/json' \
  -H 'Accept: application/json, text/event-stream' \
  -H "Mcp-Session-Id: $SESSION" \
  -d '{"jsonrpc":"2.0","id":2,"method":"tools/list"}'
data: 
id: 0
retry: 3000

data: {"jsonrpc":"2.0","id":2,"result":{"tools":[{"name":"SearchUpcomingLaunches","description":"Fields searched - launch_designator, launch_service_provider__name, mission__name, name, pad__location__name, pad__name, rocket__configuration__manufacturer__abbrev, rocket__configuration__manufacturer__name, rocket__configuration__name, rocket__spacecraftflight__spacecraft__name. Codes are the best search terms to use. Single words are the next best alternative when you cannot use a code to search","inputSchema":{"type":"object","properties":{"query":{"type":"string"}},"required":["query"]},"outputSchema":{"type":"object","properties":{"data":{"type":"object","properties":{"upcomingLaunches":{"oneOf":[{"type":"object","properties":{"pageInfo":{"oneOf":[{"type":"object","properties":{"count":{"oneOf":[{"type":"integer"},{"type":"null"}]}}},{"type":"null"}]},"results":{"oneOf":[{"type":"array","items":{"oneOf":[{"type":"object","properties":{"id":{"oneOf":[{"type":"string"},{"type":"integer"}]},"name":{"oneOf":[{"type":"string"},{"type":"null"}]},"weatherConcerns":{"oneOf":[{"type":"string"},{"type":"null"}]},"rocket":{"oneOf":[{"type":"object","properties":{"id":{"oneOf":[{"type":"string"},{"type":"integer"}]},"configuration":{"oneOf":[{"type":"object","properties":{"fullName":{"oneOf":[{"type":"string"},{"type":"null"}]}}},{"type":"null"}]}},"required":["id"]},{"type":"null"}]},"mission":{"oneOf":[{"type":"object","properties":{"name":{"oneOf":[{"type":"string"},{"type":"null"}]},"description":{"oneOf":[{"type":"string"},{"type":"null"}]}}},{"type":"null"}]},"webcastLive":{"oneOf":[{"type":"boolean"},{"type":"null"}]},"provider":{"oneOf":[{"type":"object","properties":{"name":{"oneOf":[{"type":"string"},{"type":"null"}]}}},{"type":"null"}]}},"required":["id"]},{"type":"null"}]}},{"type":"null"}]}}},{"type":"null"}]}}},"errors":{"type":"array","items":{"type":"object","properties":{"message":{"type":"string"},"locations":{"type":"array","items":{"type":"object","properties":{"line":{"type":"integer"},"column":{"type":"integer"}}}},"path":{"type":"array","items":{"oneOf":[{"type":"string"},{"type":"integer"}]}},"extensions":{"type":"object"}},"required":["message"]}},"extensions":{"type":"object"}}},"annotations":{"readOnlyHint":true,"destructiveHint":false}},...
```

### DIFF
--- a/crates/apollo-mcp-server/src/server/states/running.rs
+++ b/crates/apollo-mcp-server/src/server/states/running.rs
@@ -1812,6 +1812,244 @@ mod tests {
         }
     }
 
+    mod structured_content_gating {
+        use std::sync::Arc;
+
+        use axum::body::Body;
+        use http::{Request, StatusCode};
+        use http_body_util::BodyExt;
+        use rmcp::transport::streamable_http_server::session::local::LocalSessionManager;
+        use rmcp::transport::{StreamableHttpServerConfig, StreamableHttpService};
+        use serde_json::json;
+        use tokio::sync::RwLock;
+        use tower::ServiceExt;
+
+        use super::*;
+        use crate::operations::RawOperation;
+
+        fn create_running_with_mock_endpoint(endpoint: url::Url) -> Running {
+            let schema =
+                apollo_compiler::Schema::parse_and_validate("type Query { hello: String }", "test")
+                    .unwrap();
+
+            let raw_op: RawOperation = ("query Hello { hello }".to_string(), None).into();
+            let operation = raw_op
+                .into_operation(&schema, None, MutationMode::None, false, false, true)
+                .unwrap()
+                .expect("operation should be valid");
+
+            Running {
+                schema: Arc::new(RwLock::new(schema)),
+                operations: Arc::new(RwLock::new(vec![operation])),
+                apps: vec![],
+                headers: http::HeaderMap::new(),
+                forward_headers: vec![],
+                endpoint,
+                execute_tool: None,
+                introspect_tool: None,
+                search_tool: None,
+                explorer_tool: None,
+                validate_tool: None,
+                custom_scalar_map: None,
+                peers: Arc::new(RwLock::new(vec![])),
+                cancellation_token: CancellationToken::new(),
+                mutation_mode: MutationMode::None,
+                disable_type_description: false,
+                disable_schema_description: false,
+                enable_output_schema: true,
+                disable_auth_token_passthrough: false,
+                descriptions: HashMap::new(),
+                health_check: None,
+                server_info: Default::default(),
+            }
+        }
+
+        fn create_service(
+            running: Running,
+            session_manager: Arc<LocalSessionManager>,
+        ) -> StreamableHttpService<Running, LocalSessionManager> {
+            StreamableHttpService::new(
+                move || Ok(running.clone()),
+                session_manager,
+                StreamableHttpServerConfig {
+                    stateful_mode: true,
+                    ..Default::default()
+                },
+            )
+        }
+
+        fn build_initialize_request(protocol_version: &str) -> Request<Body> {
+            let body = json!({
+                "jsonrpc": "2.0",
+                "id": 1,
+                "method": "initialize",
+                "params": {
+                    "protocolVersion": protocol_version,
+                    "capabilities": {},
+                    "clientInfo": {
+                        "name": "test-client",
+                        "version": "1.0.0"
+                    }
+                }
+            });
+            Request::builder()
+                .method("POST")
+                .uri("/mcp")
+                .header("Content-Type", "application/json")
+                .header("Accept", "application/json, text/event-stream")
+                .body(Body::from(body.to_string()))
+                .unwrap()
+        }
+
+        fn build_notification_request(session_id: &str) -> Request<Body> {
+            let body = json!({
+                "jsonrpc": "2.0",
+                "method": "notifications/initialized"
+            });
+            Request::builder()
+                .method("POST")
+                .uri("/mcp")
+                .header("Content-Type", "application/json")
+                .header("Accept", "application/json, text/event-stream")
+                .header("Mcp-Session-Id", session_id)
+                .body(Body::from(body.to_string()))
+                .unwrap()
+        }
+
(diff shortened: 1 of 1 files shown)


crates/apollo-mcp-server/src/server/states/running.rs
```diff
@@ -491,8 +507,17 @@ impl ServerHandler for Running {
     ) -> Result<ListToolsResult, McpError> {
         let client_capabilities = context.peer.peer_info().map(|info| &info.capabilities);
 
-        self.list_tools_impl(context.extensions, client_capabilities)
-            .await
+        let mut result = self
+            .list_tools_impl(context.extensions, client_capabilities)
+            .await?;
+
+        if !self.client_supports_output_schema(&context.peer) {
```
andrewmcgivery: I'm not loving having this logic in this function given the rest of our list tools logic is in `list_tools_impl`.

If we follow the existing pattern, we could do something like...

```
let client_capabilities = context.peer.peer_info().map(|info| &info.capabilities);
let protocol_version = context.peer.peer_info().map(|info| &info.protocol_version);

self.list_tools_impl(context.extensions, client_capabilities, protocol_version)
            .await
```

However, it would mean having to refactor your `client_supports_output_schema` function to accept the `protocol_version` directly instead of the `peer`.

Your other call location would have to be something like...

```
if !self.client_supports_output_schema(&context.peer.peer_info().map(|info| &info.protocol_version))
            && let Ok(r) = &mut result
        {
            r.structured_content = None;
        }
```

So... bit of a tradeoff there but I tend to lean towards consistency in situations like this.

But, because of the tradeoff, I'm open to push back on this one. :D

crates/apollo-mcp-server/src/server/states/running.rs
```diff
@@ -480,6 +490,12 @@ impl ServerHandler for Running {
             .build()
             .add(1, &attributes);
 
+        if !self.client_supports_output_schema(&context.peer)
+            && let Ok(r) = &mut result
+        {
+            r.structured_content = None;
```
andrewmcgivery, line 491: This will fundamentally break MCP Apps cause they rely on `structured_content`. 😅 But I guess we're assuming those hosts will advertise a new-enough version of MCP protocol?

Remind me what app/host/etc we were applying this fix for?

crates/apollo-mcp-server/src/server/states/running.rs
DaleSeo: That makes sense! I updated the code so that `client_supports_output_schema` now takes `Option<&ProtocolVersion>` directly. The `outputSchema` stripping is now handled inside `list_tools_impl`. The `list_tools` function just grabs `client_capabilities` and `protocol_version` from a single `peer_info()` call and passes them through.

While I was at it, I also moved the `app_param` query-string extraction into a shared `extract_app_param` helper since it was duplicated in both `list_tools_impl` and `call_tool`.

crates/apollo-mcp-server/src/server/states/running.rs
DaleSeo: Ah, great catch, @andrewmcgivery! This fix was for Claude Desktop. It follows the negotiated protocol version and rejects `outputSchema` and `structuredContent` as unexpected when running against `2025-03-26`.

You're absolutely right that blanket-stripping `structured_content` would break MCP Apps. I added a guard so it only gets stripped when there's no `app` query parameter. Since MCP App hosts always go through the `?app=` path, they'll keep `structured_content` regardless of the protocol version.

PR #663, opened 2026-02-25
ci: update AI review summary comment in place

DaleSeo:
Every time you push changes, our AI review workflow adds a new comment. As seen in PR #657, this often clutters the PR with long blocks of text and makes it difficult to find comments from human reviewers that are relatively shorter. I suggest we adjust our AI workflow to check for existing comments and update them instead, like the way the `apollo-librarian` bot works in PR #636.

### DIFF
--- a/.claude/skills/review/SKILL.md
+++ b/.claude/skills/review/SKILL.md
@@ -197,13 +197,7 @@ After posting inline comments, add a summary comment with:
 - Final recommendation (Approve / Approve with suggestions / Request changes)
 - Sign-off line at the end: `_Reviewed by Claude Code {model}_` where `{model}` is the current model name (e.g., "Opus 4.5")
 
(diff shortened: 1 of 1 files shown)


.claude/skills/review/SKILL.md
```diff
@@ -191,7 +197,7 @@ After posting inline comments, add a summary comment with:
 - Final recommendation (Approve / Approve with suggestions / Request changes)
 - Sign-off line at the end: `_Reviewed by Claude Code {model}_` where `{model}` is the current model name (e.g., "Opus 4.5")
 
-If you have already posted a summary comment on this PR previously, any follow up summary comments should be concise and focused on what changed:
+**Important:** Always update the existing summary comment rather than creating a new one. Use the find-and-update pattern shown above. If this is a follow-up review (an existing summary comment was found), the updated content should be concise and focused on what changed:
```
andrewmcgivery:
```suggestion
**Important:** Always update the existing summary comment rather than creating a new one. Use the find-and-update pattern shown above.
```

Since we're having it update the existing comment, it can just be "latest review" instead of "what changed". :)

You can also remove the bulleted list just below this as well.

PR #664, opened 2026-02-25
fix: construct path-aware OAuth protected resource metadata URL

### LINKED ISSUE #533: Config for .well-known/oauth-protected-resource does not provide resource path for redirects
### Summary

Currently, Apollo MCP does **not provide a configuration option to customize the WWW-Authenticate resource-metadata URL**. This causes issues when Apollo MCP is deployed **behind a reverse proxy with path-based routing.**

While MCP Inspector has fallback logic that makes this usable, **other MCP clients such as VS Code and Claude strictly rely on the** _WWW-Authenticate resource-metadata value_, which breaks path-scoped MCP deployments.

### Background

MCP clients discover OAuth protected resource metadata using the WWW-Authenticate response header.

`WWW-Authenticate: Bearer resource_metadata="http://mcp.example.com/mcp/.well-known/oauth-protected-resource"
`
### Client Behavior Differences

1. MCP Inspector

Attempts discovery at:

`http://mcp.example.com/.well-known/oauth-protected-resource/mcp
`

 If not found, it falls back automatically to:

`http://mcp.example.com/.well-known/oauth-protected-resource
`

2. VS Code, Claude

Do not perform fallback

They route strictly based on the resource_metadata value returned in WWW-Authenticate

Because of this, correct configuration of resource-metadata is critical.

### Current Apollo MCP Behavior

Apollo MCP always constructs resource-metadata as:

`{resource_url}/.well-known/oauth-protected-resource`

The path component of the externally configured resource is ignored

There is no configuration option to override or customize this behavior

### Use Case: Apollo MCP Behind Reverse Proxy (Path-Based Routing)

**Architecture**

Client
  → Emissary
    → NGINX
      → Apollo MCP Server

**External Routes**
/first-service/mcp
/second-service/mcp

**Routing Flow**

1. Client calls:

`https://api.example.com/first-service/mcp`

2. Emissary:

Matches /first-service/mcp

Strips /first-service

3. NGINX forwards:

/mcp → Apollo MCP Server

From Apollo MCP’s perspective, it always receives requests at /mcp, even though the external resource is path-scoped.

**Expected Behavior**

When Apollo MCP is configured with:

resource: https://api.example.com/first-service/mcp

We expect OAuth metadata discovery at:

https://api.example.com/first-service/mcp/.well-known/oauth-protected-resource

And for Apollo MCP to:

Advertise this path-aware URL in WWW-Authenticate

Return consistent OAuth metadata responses

Allow multiple MCP servers to coexist under the same hostname

**Actual Behavior**

Apollo MCP (and downstream clients) fall back to:

https://api.example.com/.well-known/oauth-protected-resource

This happens because:

The path component (/first-service/mcp) is ignored

OAuth metadata is treated as host-scoped instead of path-scoped

**Impact**

This breaks path-based MCP deployments because:

❌ Multiple MCP servers cannot coexist under the same hostname

❌ OAuth protected resource discovery collides across products

❌ Authorization servers cannot distinguish resources correctly

❌ VS Code and Claude fail to authenticate

**Comparison**

FastMCP

Allows overriding OAuth metadata paths via middleware

Apollo MCP

No equivalent configuration or extension point

DaleSeo:
Fixes #533

The `resource_metadata` URL in the `WWW-Authenticate` header and the `.well-known` endpoint route were ignoring the path component of the configured `resource` URL. [Per RFC 9728 Section 3](https://datatracker.ietf.org/doc/html/rfc9728#name-obtaining-protected-resourc), the well-known URI must be formed by inserting `/.well-known/oauth-protected-resource` between the host and path components. This fixes OAuth metadata discovery for MCP servers deployed behind reverse proxies with path-based routing, where clients like VS Code and Claude could not authenticate.

### DIFF
--- a/crates/apollo-mcp-server/src/auth.rs
+++ b/crates/apollo-mcp-server/src/auth.rs
@@ -58,6 +58,8 @@ pub enum TlsConfigError {
     ClientBuild(#[from] reqwest::Error),
     #[error("Auth server URL at index {index} ({url}) has no host")]
     ServerUrlMissingHost { index: usize, url: String },
+    #[error("Resource URL has non-HTTP(S) scheme '{scheme}': {url}")]
+    ResourceUrlInvalidScheme { url: String, scheme: String },
 }
 
 impl TlsConfig {
@@ -215,6 +217,13 @@ impl Config {
             }
         }
 
+        let scheme = self.resource.scheme();
+        if scheme != "http" && scheme != "https" {
+            return Err(TlsConfigError::ResourceUrlInvalidScheme {
+                url: self.resource.to_string(),
+                scheme: scheme.to_string(),
+            });
+        }
         if self.allow_any_audience {
             warn!(
                 "allow_any_audience is enabled - audience validation is disabled. This reduces security."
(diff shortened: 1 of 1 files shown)


crates/apollo-mcp-server/src/auth.rs
```diff
@@ -256,10 +284,10 @@ async fn oauth_validate(
         (
             StatusCode::UNAUTHORIZED,
             TypedHeader(WwwAuthenticate::Bearer {
-                resource_metadata: resource_metadata_url(),
```
gocamille, line 259: Do you think we should add an integration test that triggers `401`, parses `WWW-Authenticate`, and then  asserts that `resource_metadata` is the exact RFC-style URL for a path-scoped resource?

crates/apollo-mcp-server/src/auth.rs
```diff
@@ -214,10 +251,7 @@ impl Config {
             .allow_methods([Method::GET])
             .allow_origin(Any);
         let auth_router = Router::new()
-            .route(
-                "/.well-known/oauth-protected-resource",
-                get(protected_resource),
-            )
```
gocamille, line 220: Would it be best to add a test that `GET`s the exact `resource_metadata` path that is in the header and expects `200`?

crates/apollo-mcp-server/src/auth.rs
```diff
@@ -237,13 +271,7 @@ async fn oauth_validate(
     next: Next,
 ) -> Result<Response, (StatusCode, TypedHeader<WwwAuthenticate>)> {
     let auth_config = &auth_state.config;
-
-    // Helper to construct the resource metadata URL
-    let resource_metadata_url = || {
-        let mut url = auth_config.resource.clone();
-        url.set_path("/.well-known/oauth-protected-resource");
-        url
-    };
```
gocamille, line 246: This makes me wonder about those who hard-code the metadata_url in a CI/CD flow or elsewhere, where this may break for them now. Should we add validation coverage for invalid resource values (i.e. non-HTTP(S) scheme or missing `host`) so startup fails fast?

crates/apollo-mcp-server/src/auth.rs
DaleSeo: Great idea! Added `unauthorized_response_contains_path_scoped_resource_metadata_url`.

crates/apollo-mcp-server/src/auth.rs
DaleSeo: Yep, good call! Added `get_resource_metadata_path_returns_ok` which GETs the path-scoped metadata URL and asserts `200 OK`.

crates/apollo-mcp-server/src/auth.rs
DaleSeo: Yeah, totally fair concern! Added a scheme check that rejects non-HTTP resource URLs at startup.

PR #667, opened 2026-02-26
feat: add camelCase-aware search tokenization

DaleSeo:
While working on https://github.com/apollographql/rover/pull/3020, I realized that the MCP server's schema index does not split camelCase identifiers when tokenizing, so searching for "post" fails to match types like `PostConnection`, `createPost`, or `CreatePostInput`.

### DIFF
--- a/crates/apollo-schema-index/src/lib.rs
+++ b/crates/apollo-schema-index/src/lib.rs
@@ -276,7 +276,7 @@ impl SchemaIndex {
                 description_field,
                 extended_type
                     .description()
-                    .map(|d| d.to_string())
+                    .map(|d| expand_identifiers(d))
                     .unwrap_or_default(),
             );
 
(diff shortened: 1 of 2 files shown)


crates/apollo-schema-index/src/lib.rs
```diff
@@ -261,7 +314,7 @@ impl SchemaIndex {
                 ),
                 _ => String::new(),
             };
```
gocamille, line 316: Hi @DaleSeo ! This one took me a little longer to parse as I'm new to this crate. But I noticed we’re running `expand_identifiers()` on query terms now (`token_stream(&expanded)`), and we also expand type names and the big `fields` string when indexing 👀 

But I believe descriptions are still indexed raw (`doc.add_text(description_field, ...`) without `expand_identifiers`). If a camelCase identifier only shows up in the description (like “Use `createPost` to…”), it might get indexed as one token (`createpost`) while the query becomes `create post`, so description-only matches could get missed.

Is it maybe worth also expanding descriptions at index time (or storing both the raw and expanded values)?

DaleSeo: Great catch, @gocamille! That was my oversight. I think expanding descriptions will help improve the search quality. I'll make sure to apply this to the Rover PR as well.

PR #669, opened 2026-02-27
fix: allow missing `aud` claim in access tokens for AWS Cognito compatibility

### LINKED ISSUE #640: Allow missing `aud` claim in access tokens for AWS Cognito compatibility
# Overview

I am trying to connect apollo-mcp-server 1.7.0 with our GraphQL API that is behind AWS Cognito auth.

However, I am getting the following error log during token validation: 
> 2026-02-13T14:35:12.927197Z  INFO mcp_server{method=GET uri=/mcp}:oauth_validate: Authorization server metadata discovered url=https://cognito-idp.eu-west-1.amazonaws.com/pool-id/.well-known/openid-configuration
> 2026-02-13T14:35:12.928028Z  WARN mcp_server{method=GET uri=/mcp}:oauth_validate: Token failed validation with error: JSON error: missing field `aud` at line 1 column 462
> 2026-02-13T14:35:12.928064Z  INFO mcp_server{method=GET uri=/mcp}:oauth_validate: Token did not pass validation

AWS Cognito is known for completely ommitting the `aud` claim in their access tokens unless a very specific managed login flow is used together with resource binding. Such a flow is not used in our case.

Would it be feasible to provide a way to configure apollo-mcp-server to skip the requirement for `aud` claim presence?

Another alternative to consider could be to skip validating the token all together (just its presence), since our GraphQL API validates it too anyway.

# References

- Previous issue that led to the introduction of `allow_any_audience` flag: https://github.com/apollographql/apollo-mcp-server/issues/438
- AWS Docs saying that `aud` claim is only returned when resource binding is used: https://docs.aws.amazon.com/cognito/latest/developerguide/amazon-cognito-user-pools-using-the-access-token.html#user-pool-access-token-payload
- AWS Docs saying that resource binding can only be used with a managed login flow: https://docs.aws.amazon.com/cognito/latest/developerguide/cognito-user-pools-define-resource-servers.html#cognito-user-pools-resource-binding

## `apollo-mcp-config.yml`
```yaml
endpoint: graphql_endpoint

schema:
  source: local
  path: ./schema.graphql

operations:
  source: introspect

introspection:
  introspect:
    enabled: true
  search:
    enabled: true
  execute:
    enabled: true

overrides:
  mutation_mode: all

transport:
  type: streamable_http
  port: 8456
  auth:
    servers:
      - https://cognito-idp.eu-west-1.amazonaws.com/pool-id
    scopes:
      - email 
      - openid
    scope_mode: disabled
    allow_any_audience: true
    resource: http://localhost:8456/mcp

forward_headers:
  - Authorization

logging:
  level: info
```

mayorandrew:
## Summary

- Fixes JWT validation failure when access tokens omit the `aud` claim entirely (e.g., AWS Cognito access tokens)
- When `allow_any_audience: true`, tokens without `aud` are now accepted instead of failing with `"missing field 'aud'"`
- When `allow_any_audience: false`, tokens without `aud` are still explicitly rejected to maintain security

## Details

AWS Cognito access tokens [omit the `aud` claim](https://docs.aws.amazon.com/cognito/latest/developerguide/amazon-cognito-user-pools-using-the-access-token.html#user-pool-access-token-payload) unless resource binding with managed login is used. Previously, the `Claims` struct required `aud` during deserialization, causing a serde error before validation logic could run — even with `allow_any_audience: true`.

Changes in `crates/apollo-mcp-server/src/auth/valid_token.rs`:
- Add `#[serde(default)]` to the `aud` field so missing `aud` deserializes to an empty vec
- Update `deserialize_audience` to handle `null`/missing values
- Add explicit post-decode check: reject empty `aud` when `allow_any_audience` is false (the `jsonwebtoken` crate skips its own audience check when the claim is absent from the raw JWT)
- Add two new tests covering both the accept and reject paths

## Test plan

- [x] `cargo test -p apollo-mcp-server valid_token` — all 16 tests pass
- [x] `cargo clippy --all-targets -- --deny warnings` — clean
- [x] `cargo fmt --check` — clean
- [ ] CI passes

Closes #640

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/apollo-mcp-server/src/auth/valid_token.rs
+++ b/crates/apollo-mcp-server/src/auth/valid_token.rs
@@ -141,7 +141,7 @@ pub(super) trait ValidateToken {
                     // so we enforce it here.
                     if !self.allow_any_audience() && token_data.claims.aud.is_empty() {
                         warn!("Token is missing the required `aud` claim");
-                        continue;
+                        break;
                     }
                     let scopes = extract_scopes(token_data.claims.scope.as_deref());
                     return Some(ValidToken { token, scopes });


crates/apollo-mcp-server/src/auth/valid_token.rs
```diff
@@ -458,6 +469,107 @@ mod test {
...
     }
 
+    #[tokio::test]
+    async fn it_validates_jwt_with_missing_audience_when_allow_any() {
+        use serde_json::json;
+
+        let key_id = "some-example-id".to_string();
+        let (encode_key, decode_key) = create_key("DEADBEEF");
+        let jwk = Jwk {
+            alg: KeyAlgorithm::HS512,
```
DaleSeo: We recently updated `jwks` to v0.5.3, and this field changed to `Option<KeyAlgorithm>`. Could you rebase the PR onto `main`? This will create a type error that you'll need to fix.

```suggestion
            alg: Some(KeyAlgorithm::HS512),
```

crates/apollo-mcp-server/src/auth/valid_token.rs
```diff
@@ -128,6 +131,14 @@ pub(super) trait ValidateToken {
 
             match decode::<Claims>(jwt, &jwk.decoding_key, &validation) {
                 Ok(token_data) => {
+                    // When audience validation is enabled, explicitly reject tokens
+                    // with a missing `aud` claim. The `jsonwebtoken` crate skips its
+                    // own audience check when the claim is absent from the raw JWT,
+                    // so we enforce it here.
+                    if !self.allow_any_audience() && token_data.claims.aud.is_empty() {
+                        warn!("Token is missing the required `aud` claim");
+                        continue;
```
DaleSeo: Would it be worth checking the remaining servers?

DaleSeo (commented):
Thanks for your contribution, @mayorandrew! This looks great. I have a couple of comments. Can you also add a changeset file? (e.g. https://github.com/apollographql/apollo-mcp-server/blob/main/.changeset/fix_path_aware_resource_metadata_url.md)

crates/apollo-mcp-server/src/auth/valid_token.rs
mayorandrew: Thanks for the heads up, I rebased on the latest main commit and fixed the type error.

crates/apollo-mcp-server/src/auth/valid_token.rs
mayorandrew: Good point, token won't change for other servers, so we can abort early. Replaced with `break`.

PR #672, opened 2026-03-02
feat: implements step-up-authorization-flow

DaleSeo:
This PR allows the MCP server to declare required OAuth scopes for each GraphQL operation. It implements [the step-up authorization flow](https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#step-up-authorization-flow) from the MCP Auth Spec 2025-11-25. Now, when a client calls a tool with a valid token that doesn't have the required scopes, the server responds with HTTP 403 and includes a `WWW-Authenticate: Bearer error="insufficient_scope", scope="..."` header. This way, the client knows exactly which scopes to request before trying again, instead of failing silently or needing broad permissions upfront.

The mechanism works through a `ScopeErrorCell` added to Axum request extensions by the `oauth_validate` middleware. If `call_tool` finds missing scopes, it writes to the cell and returns an MCP error. The middleware then converts that response to HTTP 403. This keeps the MCP layer separate from the HTTP transport layer. You can configure scope requirements under `overrides.required_scopes` in YAML, and they will be ignored if auth isn't set up or when running in stdio mode.

EDIT: Upgraded `rmcp` to v0.17 to get the auth-related fixes.

# Testing

Issue two tokens: one with the `launches:read` scope and one without.

2026-03-09 at 09 45 21

2026-03-09 at 09 48 05

Configure the MCP server to require `launches:read` the scope for the `SearchUpcomingLaunches` tool.

```yaml
endpoint: https://thespacedevs-production.up.railway.app/

transport:
  type: streamable_http
  auth:
    servers:
      - https://dev-yrrjv32vzhac3vp6.us.auth0.com
    audiences: 
      - http://localhost:8000/mcp-example
    resource: http://127.0.0.1:8000/mcp
    scopes: []

operations:
  source: local
  paths:
    - ./graphql/TheSpaceDevs/operations

schema:
  source: local
  path: ./graphql/TheSpaceDevs/api.graphql

overrides:
  mutation_mode: all
  required_scopes:
    SearchUpcomingLaunches:
      - launches:read   # only this tool requires a scope
```

1. Should be able to call `SearchUpcomingLaunches` tool using the token that doesn't have the `launches:read` scope.

1. 200 using the token that has the `launches:read` scope.

```sh
$ curl -s -D - -o /dev/null -X POST http://127.0.0.1:8000/mcp -I \
  -H "Content-Type: application/json" -H "Accept: application/json, text/event-stream" \
  -H "mcp-protocol-version: 2025-03-26" \
  -H "Authorization: Bearer <TOKEN_WITH_SCOPE>" \
  -d '{"jsonrpc":"2.0","id":1,"method":"tools/call","params":{"name":"SearchUpcomingLaunches","arguments":{"query":"SpaceX"}}}'
```

```sh
HTTP/1.1 200 OK
content-type: text/event-stream
cache-control: no-cache
vary: origin, access-control-request-method, access-control-request-headers
access-control-allow-origin: *
access-control-expose-headers: mcp-session-id,traceparent,tracestate
traceparent: 00-d5d469c6a56d014b7a2afa3e2deb5470-4363de51e9835ccd-01
tracestate:
transfer-encoding: chunked
date: Mon, 09 Mar 2026 13:36:59 GMT
```

2. Should receive 403 when using the token that doesn't have the `launches:read` scope.

```sh
$ curl -s -D - -o /dev/null -X POST http://127.0.0.1:8000/mcp \
  -H "Content-Type: application/json" -H "Accept: application/json, text/event-stream" \
  -H "mcp-protocol-version: 2025-03-26" \
  -H "Authorization: Bearer <TOKEN WITHOUT SCOPE>" \
  -d '{"jsonrpc":"2.0","id":1,"method":"tools/call","params":{"name":"SearchUpcomingLaunches","arguments":{"query":"SpaceX"}}}'
```

```sh
HTTP/1.1 403 Forbidden
www-authenticate: Bearer resource_metadata="http://127.0.0.1:8000/.well-known/oauth-protected-resource/mcp", error="insufficient_scope", scope="launches:read", scope_mode="disabled"
vary: origin, access-control-request-method, access-control-request-headers
access-control-allow-origin: *
access-control-expose-headers: mcp-session-id,traceparent,tracestate
traceparent: 00-1719e8f1c5c7b77b702669ea24495987-121217ba02767775-01
tracestate:
content-length: 0
date: Mon, 09 Mar 2026 13:40:06 GMT
```

1. Should be still able to call other tools using the token that doesn't have the `launches:read` scope.

```sh
$ curl -s -D - -o /dev/null -X POST http://127.0.0.1:8000/mcp \
  -H "Content-Type: application/json" -H "Accept: application/json, text/event-stream" \
  -H "mcp-protocol-version: 2025-03-26" \
  -H "Authorization: Bearer <TOKEN WITHOUT SCOPE>" \
  -d '{"jsonrpc":"2.0","id":2,"method":"tools/call","params":{"name":"GetAstronautsCurrentlyInSpace","arguments":{}}}'
```

```sh
HTTP/1.1 200 OK
content-type: text/event-stream
cache-control: no-cache
vary: origin, access-control-request-method, access-control-request-headers
access-control-allow-origin: *
access-control-expose-headers: mcp-session-id,traceparent,tracestate
traceparent: 00-fe9a89a3a84079820854144367faedc1-a844b31dc40fd3d2-01
tracestate:
transfer-encoding: chunked
date: Mon, 09 Mar 2026 13:42:51 GMT
```

### DIFF
--- a/crates/apollo-mcp-server/src/auth.rs
+++ b/crates/apollo-mcp-server/src/auth.rs
@@ -5,7 +5,6 @@ use std::time::Duration;
 
 use axum::{
     Json, Router,
-    body::Body,
     extract::{Request, State},
     http::StatusCode,
     middleware::Next,
@@ -216,10 +215,10 @@ struct AuthState {
 }
 
 impl Config {
-    /// Enable auth middleware with per-operation scope requirements.
+    /// Enable auth middleware on the router.
     ///
     /// Builds the HTTP client at startup to validate TLS configuration eagerly.
-    pub fn enable_middleware_with_scopes(
+    pub fn enable_middleware(
         &self,
         router: Router,
         required_scopes: HashMap<String, Vec<String>>,
@@ -298,10 +297,17 @@ const ANONYMOUS_DISCOVERY_METHODS: &[&str] = &["initialize", "tools/list", "reso
 /// headroom while preventing abuse from unauthenticated clients.
 const ANONYMOUS_PEEK_BODY_LIMIT: usize = 16 * 1024;
 
-/// Minimal struct for deserializing just the `method` field from a JSON-RPC request.
+/// Struct for deserializing the JSON-RPC `method` and optional `params.name`
+/// from a request body. Used by both anonymous discovery and per-operation scope checks.
 #[derive(Deserialize)]
 struct JsonRpcBodyPeek {
     method: String,
+    params: Option<JsonRpcParams>,
+}
+
+#[derive(Deserialize)]
+struct JsonRpcParams {
+    name: Option<String>,
 }
 
 async fn extract_body(request: &mut Request) -> Result<JsonRpcBodyPeek, StatusCode> {
@@ -325,32 +331,19 @@ async fn extract_body(request: &mut Request) -> Result<JsonRpcBodyPeek, StatusCo
     Ok(peek)
 }
 
-// Used to extract the operation name from a `tools/call` body for per-operation scope checks.
-#[derive(Deserialize)]
-struct JsonRpcToolsCall<'a> {
-    method: &'a str,
-    params: Option<JsonRpcToolsCallParams<'a>>,
-}
-
-#[derive(Deserialize)]
-struct JsonRpcToolsCallParams<'a> {
-    name: &'a str,
-}
-
-/// Checks if a `tools/call` request body is missing required scopes for the named operation.
+/// Checks if a `tools/call` request is missing required scopes for the named operation.
 ///
 /// Returns `Some(required_scopes)` if the token is missing scopes, `None` if the request
 /// is allowed to proceed (wrong method, unknown operation, or scopes satisfied).
 fn missing_scopes_for_operation<'a>(
-    body: &[u8],
+    peek: &JsonRpcBodyPeek,
     required_scopes: &'a HashMap<String, Vec<String>>,
     token_scopes: &[String],
 ) -> Option<&'a [String]> {
-    let req = serde_json::from_slice::<JsonRpcToolsCall>(body).ok()?;
-    if req.method != "tools/call" {
+    if peek.method != "tools/call" {
         return None;
     }
-    let op_name = req.params?.name;
+    let op_name = peek.params.as_ref()?.name.as_deref()?;
     let required = required_scopes.get(op_name)?;
     if required.iter().all(|s| token_scopes.contains(s)) {
         return None;
@@ -401,25 +394,34 @@ async fn oauth_validate(
         )
     };
 
-    // Anonymous discovery bypass: when enabled, unauthenticated POST requests
-    // whose body contains a discovery method (e.g. tools/list) skip auth entirely.
-    if auth_config.allow_anonymous_mcp_discovery
-        && token.is_none()
-        && request.method() == http::Method::POST
+    // Extract the body once if we need to inspect the JSON-RPC method for either
+    // anonymous discovery or per-operation scope checks.
+    let body_peek = if request.method() == http::Method::POST
+        && ((auth_config.allow_anonymous_mcp_discovery && token.is_none())
+            || !auth_state.required_scopes.is_empty())
     {
-        let json_rpc_body_peek = match extract_body(&mut request).await {
-            Ok(result) => result,
+        match extract_body(&mut request).await {
+            Ok(peek) => Some(peek),
             Err(status) => {
                 tracing::Span::current().record("status_code", status.as_u16());
                 return Ok(status.into_response());
             }
-        };
-
-        if ANONYMOUS_DISCOVERY_METHODS.contains(&json_rpc_body_peek.method.as_str()) {
-            let response = next.run(request).await;
-            tracing::Span::current().record("status_code", response.status().as_u16());
-            return Ok(response);
         }
+    } else {
+        None
+    };
+
+    // Anonymous discovery bypass: when enabled, unauthenticated POST requests
+    // whose body contains a discovery method (e.g. tools/list) skip auth entirely.
+    if auth_config.allow_anonymous_mcp_discovery
+        && token.is_none()
+        && body_peek
+            .as_ref()
+            .is_some_and(|peek| ANONYMOUS_DISCOVERY_METHODS.contains(&peek.method.as_str()))
+    {
+        let response = next.run(request).await;
+        tracing::Span::current().record("status_code", response.status().as_u16());
+        return Ok(response);
     }
 
     let discovery_timeout = auth_config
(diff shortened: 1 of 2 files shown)


crates/apollo-mcp-server/src/server/states/running.rs
```diff
@@ -582,9 +590,9 @@ impl ServerHandler for Running {
         };
 
         let protocol_version = if self.enable_output_schema {
-            ProtocolVersion::V_2025_06_18
-        } else {
             ProtocolVersion::default()
+        } else {
+            ProtocolVersion::V_2025_03_26
```
andrewmcgivery, line 590: Why are we flipping this if statement and changing the date? This feels unrelated to this PR's scope.

crates/apollo-mcp-server/src/auth.rs
```diff
@@ -200,13 +204,24 @@ struct AuthState {
...
+    /// Per-operation required scopes, keyed by operation name.
+    required_scopes: Arc<HashMap<String, Vec<String>>>,
 }
 
 impl Config {
     /// Enable auth middleware on the router.
     ///
     /// Builds the HTTP client at startup to validate TLS configuration eagerly.
     pub fn enable_middleware(&self, router: Router) -> Result<Router, TlsConfigError> {
+        self.enable_middleware_with_scopes(router, HashMap::new())
```
andrewmcgivery: Do we use both `enable_middleware` and `enable_middleware_with_scopes`? 🤔

crates/apollo-mcp-server/src/auth.rs
```diff
@@ -271,12 +287,52 @@ impl Config {
...
+}
+
+// Maximum body size buffered when checking per-operation scopes (1 MiB).
+const MAX_TOOLS_CALL_BODY: usize = 1024 * 1024;
+
+/// Checks if a `tools/call` request body is missing required scopes for the named operation.
+///
+/// Returns `Some(required_scopes)` if the token is missing scopes, `None` if the request
+/// is allowed to proceed (wrong method, unknown operation, or scopes satisfied).
+fn missing_scopes_for_operation<'a>(
```
andrewmcgivery, line 338: I'm struggling with the readability of this method and the logic that leads up to it in the code that calls this function. There's just so much going on and I am wondering if we can simplify it, especially given I have a PR that does something similar and I'd hate to have multiple implementations of "deconstruct the body and then re-construct it again. 😅 

First, I think we should have a general struct for `JsonRpcMethodPeek` and that struct should not be embedded within the method like you have it here. Having structs inside of functions is incredibly dificutl on the eyes/brain. 😅 

Next, Instead of doing a combination of streaming and string contains and etc, can the logic be something like...

1. Function that returns a copy of the de-serialized body as the `JsonRpcMethodPeek` struct
2. Apply any logic needed based on that struct
3. Convert back to bytes.

I'm hoping this will also clean up the need for all the lifetimes and etc cause those also are a bit messy right now.

crates/apollo-mcp-server/src/server/states/running.rs
DaleSeo: I encountered a session handling bug during testing and had to upgrade `rmcp` to v0.17, as mentioned in the PR description. This change is needed because `rmcp` changed the default protocol version from 2025-03-26 to 2025-06-18 in the new version.

crates/apollo-mcp-server/src/auth.rs
DaleSeo: No, I just wanted to make Claude happy. 😅 I didn't have this one-line abstraction ealier, but the Rust handbook suggested adding this to avoid passing the empty map repeatedly in the tests. Now I regret doing that, so I'll remove it.

crates/apollo-mcp-server/src/auth.rs
DaleSeo: Thanks for the feedback, @andrewmcgivery! I've moved `Params` and `Request` out of missing_scopes_for_operation to the module level.

On the prefix-peek complexity, I hear you that it's a lot going on, but my initial implementation didn't touch the body at all. I added that complexity while addressing this AI review feedback:
- https://github.com/apollographql/apollo-mcp-server/pull/672#discussion_r2905687698
- https://github.com/apollographql/apollo-mcp-server/pull/672#discussion_r2905904638

I think this is more of a micro-optimization that impacts readability, and I can revert it if you agree. Let me know!

gocamille:
Hi @DaleSeo thank you for this PR! I am still taking a look but a design element struck out to me, which is that the `required_scopes` are being written in two places, but only read in one: 

- `required_scopes` is added to the `RawOperation` constructor, and are cloned into the `Operation` struct; and
- `required_scopes` is added to the auth middleware (via the `enable_middleware_with_scopes()` function). 

And it seems it's being read in the auth middleware alone: 
- The middleware intercepts each HTTP request and checks whether the `body` of the request is a `tools/call` method;
- If so, it gets the operation name and checks whether the name is in the `required_scopes` HashMap passed to the auth middleware. 

It seems the `required_scopes` on the operation are never read outside of tests, and the real enforcement is done in the auth middleware path. Should we remove the `required_scopes` from the former, as it's only read in the middleware? Or is there maybe another design approach we are considering here?

(Oh, and I just realized that removing the logic from `operations` -- if that's the case we're going for -- would allow us to remove the additional logic from the `starting` state machine, as the auth middleware truly is doing all the work 🦾 !)

crates/apollo-mcp-server/src/server/states/running.rs
gocamille: Ah! I was wondering the same! @DaleSeo do we happen to have some migration / update story when we can move this `ProtocolVersion` tagged from a specific version back to `default()`?

crates/apollo-mcp-server/src/auth.rs
andrewmcgivery: It's definitely some premature optimization by our friend Claude. :D Lets air on the side of simplicity here since we don't expect super large request bodies!

andrewmcgivery: @DaleSeo I've got a new, cleaner, more generic implementation of body extraction in my PR: https://github.com/apollographql/apollo-mcp-server/pull/685 (see `extract_body` function)

You should be able to use this same implementation for your PR!

DaleSeo: Hey @andrewmcgivery, I'd prefer to wait for your PR to land first. This is a critical feature, so I want to rebase and simplify the implementation if still needed, and then manually test it again with Auth0 once everything is settled.

crates/apollo-mcp-server/src/server/states/running.rs
DaleSeo: @gocamille The `ProtocolVersion` branching was added to fix the issue where `enable_output_schema: true` caused Claude Desktop to fail. For more context, see https://community.apollographql.com/t/enable-output-schema-breaks-claude-desktop-tool-execution/9721

So, I think until all major MCP clients properly support `2025-06-18`, the explicit pin to `V_2025_03_26` is intentional, and we should maintain this branching. I have no idea when that will happen. 😅

DaleSeo:
That's great design feedback, @gocamille! I've dropped `required_scopes` from both `RawOperation` and `Operation`.  I also took it out of `Running` since it was only used in `Starting`.

crates/apollo-mcp-server/src/auth.rs
DaleSeo: @andrewmcgivery After rebasing on your `extract_body`, I replaced the streaming/prefix-peek logic with the same simple pattern. I also refactored `extract_body` slightly following the rust best practices. Let me know if this is easier on your eyes.

crates/apollo-mcp-server/src/auth.rs
```diff
@@ -297,25 +307,55 @@ struct JsonRpcBodyPeek {
...
 
     *request.body_mut() = axum::body::Body::from(bytes);
 
-    Ok(json_rpc_body_peek)
+    Ok(peek)
+}
+
+// Used to extract the operation name from a `tools/call` body for per-operation scope checks.
+#[derive(Deserialize)]
+struct JsonRpcToolsCall<'a> {
```
andrewmcgivery: You shouldn't need to duplicate this object, you should be able to just add `method` to `JsonRpcBodyPeek`.

crates/apollo-mcp-server/src/auth.rs
```diff
@@ -443,6 +483,34 @@ async fn oauth_validate(
         }
     }
 
+    // Per-operation scope check: buffer the body to inspect the JSON-RPC method and
+    // operation name. Skip entirely when no per-operation scopes are configured.
+    if !auth_state.required_scopes.is_empty() {
+        let body = std::mem::take(request.body_mut());
```
andrewmcgivery: We shoudn't be doing this body extraction logic twice in the same middleware.

If you move the call to `extract_body` on line 410 so that it is before the if statement on line 406, you can reuse the `json_rpc_body_peek` down here which removes a lot of duplication and should simplify your code as a result.

crates/apollo-mcp-server/src/auth.rs
```diff
@@ -208,13 +211,19 @@ struct AuthState {
...
+    required_scopes: Arc<HashMap<String, Vec<String>>>,
 }
 
 impl Config {
-    /// Enable auth middleware on the router.
+    /// Enable auth middleware with per-operation scope requirements.
     ///
     /// Builds the HTTP client at startup to validate TLS configuration eagerly.
-    pub fn enable_middleware(&self, router: Router) -> Result<Router, TlsConfigError> {
+    pub fn enable_middleware_with_scopes(
```
andrewmcgivery: I think we can just keep the original `enable_middleware` name here. :)

crates/apollo-mcp-server/src/auth.rs
DaleSeo: @andrewmcgivery I moved `extract_body` up front so the body is parsed once, and both the anonymous discovery check and per-operation scope check now use the same `body_peek`.

One thing I'm slightly concerned about after refactoring is that the token validation and global scope validation are separating these two checks pretty far. Do you have any suggestions for colocating them somehow, or do you think it reads fine as-is?

andrewmcgivery: I think this is fine for now. I need to explore some refactoring of this part of the code anyways to support per-tool auth for apps so this is likely all subject to change anyways lol

Appreciate the updates though, this is super clean now!

gocamille (approved):

PR #673, opened 2026-03-02
App resources will now only be listed when the "app" query param is present

andrewmcgivery:
Also makes sure to only list the resource relevant to that particular tool

### DIFF
--- a/crates/apollo-mcp-server/src/server/states/running.rs
+++ b/crates/apollo-mcp-server/src/server/states/running.rs
@@ -298,7 +298,7 @@ impl Running {
                 Some(app) => vec![attach_resource_mime_type(app.resource())],
                 None => {
                     return Err(McpError::new(
-                        ErrorCode::INVALID_REQUEST,
+                        ErrorCode::INVALID_PARAMS,
                         format!("App {app_name} not found"),
                         None,
                     ));


crates/apollo-mcp-server/src/server/states/running.rs
```diff
@@ -286,13 +286,30 @@ impl Running {
...
+    ) -> Result<ListResourcesResult, McpError> {
+        let app_param = extract_app_param(extensions);
+
+        let resources = if let Some(app_name) = app_param {
+            let app = self.apps.iter().find(|app| app.name == app_name);
+            match app {
+                Some(app) => vec![attach_resource_mime_type(app.resource())],
+                None => {
+                    return Err(McpError::new(
+                        ErrorCode::INVALID_REQUEST,
```
DaleSeo: Not sure if this is the right error code to use. According to [the JSON-RPC spec](https://www.jsonrpc.org/specification#error_object), it looks like `-32600 INVALID_REQUEST` means the JSON sent isn't a valid Request object. Shouldn't we use `-32602 INVALID_PARAMS` instead for an unrecognized parameter?

```suggestion
                        ErrorCode::INVALID_PARAMS,
```

crates/apollo-mcp-server/src/server/states/running.rs
```diff
@@ -310,13 +327,21 @@ impl Running {
...
-        })
+        if let Some(app_name) = app_param {
+            let resource =
+                get_app_resource(&self.apps, request, request_uri, &app_target, &app_name).await?;
+            Ok(ReadResourceResult {
+                contents: vec![resource],
+            })
+        } else {
+            Err(ErrorData::resource_not_found(
+                format!("Resource not found for URI: {}", request.uri),
```
DaleSeo, line 341: It would be helpful for the client to know why the resource wasn't found.

```suggestion
                "Resource cannot be read without an `app` query parameter".to_string(),
```

andrewmcgivery: The reason I left out this specific of a message was more to leave room for incase we have a reason for other kinds of resources in the future.

So the error shouldn't be understood as "you didn't give us an app param" but rather "we don't have a matching resource for what you requested".

We return the same error message from inside `get_app_resource` when things don't match up.

PR #677, opened 2026-03-03
Refactor call_tool into impl method, refactor tests

andrewmcgivery:
Refactor call_tool into impl method, refactor all tests into mods and to remove significant scaffolding

### DIFF
--- a/crates/apollo-mcp-server/src/server/states/running.rs
+++ b/crates/apollo-mcp-server/src/server/states/running.rs
@@ -2065,6 +2065,474 @@ mod tests {
             operation_mock.assert();
         }
     }
+}
+
+#[cfg(test)]
+mod integration_tests {
+    use super::*;
+
+    mod output_schema_gating {
+        use std::sync::Arc;
+
+        use axum::body::Body;
+        use http::{Request, StatusCode};
+        use http_body_util::BodyExt;
+        use rmcp::transport::streamable_http_server::session::local::LocalSessionManager;
+        use rmcp::transport::{StreamableHttpServerConfig, StreamableHttpService};
+        use serde_json::json;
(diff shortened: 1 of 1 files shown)


crates/apollo-mcp-server/src/server/states/running.rs
```diff
@@ -522,6 +431,106 @@ impl ServerHandler for Running {
...
+            .build()
+            .add(1, &attributes);
+        // TODO: how to remove these?
+        let mut peers = self.peers.write().await;
+        peers.push(context.peer);
+        Ok(self.get_info())
+    }
+
+    #[tracing::instrument(skip_all, parent = get_parent_span(&context), fields(apollo.mcp.tool_name = request.name.as_ref(), apollo.mcp.request_id = %context.id.clone()))]
+    async fn call_tool(
```
DaleSeo, line 522: The old tests in `output_schema_gating` and `structured_content_gating` checked the complete MCP HTTP flow. However, the new tests in `mod list_tools` and `mod call_tool` pass `protocol_version` directly to `call_tool_impl` and `list_tools_impl`, skipping this `call_tool` wrapper.

Could we consider at least keeping one integration test that follows the full `initialize` → `tools/call` flow?

andrewmcgivery: I added a new mod in the file called "integration_tests" and added those tests back in.

In the future it would probably be good to get our integration tests setup under the root `/tests` folder but thats a bit more time than I have right now to spend on this. 😅

DaleSeo: Yeah, that would definitely make for a better test organization!

PR #681, opened 2026-03-04
Introduce Rhai Extensibility MVP

andrewmcgivery:
Note: Docs will be in a separate PR!

### DIFF
--- a/crates/apollo-mcp-server/src/server/states/starting.rs
+++ b/crates/apollo-mcp-server/src/server/states/starting.rs
@@ -146,7 +146,6 @@ impl Starting {
             _ => None, // No health checks for Stdio or when disabled.
         };
 
-        // TODO: Load Rhai Scripts
         let mut engine = RhaiEngine::new();
         engine.load_from_path().map_err(|err| {
             error!("Error loading Rhai scripts: {err}");


crates/apollo-mcp-server/src/server/states/starting.rs
```diff
@@ -145,6 +146,21 @@ impl Starting {
             _ => None, // No health checks for Stdio or when disabled.
         };
 
+        // TODO: Load Rhai Scripts
```
gocamille: Initially I thought this `TODO: Load Rhai Scripts` was a forgotten comment, but in looking at what the `load_from_path()` function does, I noticed that the script was hardcoded to `main.rhai`. Does this mean there may be additional work to grab this rhai script from a field in the MCP server config in the future?

crates/apollo-mcp-server/src/server/states/starting.rs
```diff
@@ -145,6 +146,21 @@ impl Starting {
...
+        let mut engine = RhaiEngine::new();
```
gocamille, line 149: should this `load_from_path` be conditional and gated by the `experimental_rhai` feature? I'm realizing that right now someone could load a `main.rhai` script that is malicious, and since the `on_execute_graphql_function` in `execute.rs` is also not gated a default build will always run that rhai script. (I tested a mock malicious script to validate this assumption in an "attempt" to steal env vars)

crates/apollo-mcp-server/src/operations/execution.rs
```diff
@@ -16,21 +21,38 @@ pub(crate) async fn find_and_execute_operation(
...
 }
 
 pub(crate) async fn execute_operation(
     executable: &impl Executable,
     headers: &HeaderMap,
     arguments: Option<&JsonObject>,
     endpoint: &Url,
+    rhai_engine: &Arc<Mutex<RhaiEngine>>,
+    axum_parts: Option<&Parts>,
 ) -> Result<CallToolResult, McpError> {
```
gocamille, line 48: Bringing this up here too, as with the [note](https://github.com/apollographql/apollo-mcp-server/pull/681/changes#r2950419890) in `starting.rs` I'm wondering if we should gate this behind the `experimental_rhai` feature?

crates/apollo-mcp-server/src/server/states/starting.rs
andrewmcgivery: This was indeed an old TODO I forgot to take out 😅 

`main.rhai` is indeed hardcoded on purposes. In the future we may add configurability but I wanted to start with convention based as part of the MVP and layer on complexity as required. :)

crates/apollo-mcp-server/src/server/states/starting.rs
andrewmcgivery: As mentioned on slack, the feature flag isn't saying rhai itself is experimental, rather Its for feature flagging features of rhai that are experimental.

In this case, the `on_startup` checkpoint and `http` module are feature flagged because they aren't ready for prime time.

crates/apollo-mcp-server/src/operations/execution.rs
andrewmcgivery: See above comment!

PR #683, opened 2026-03-05
chore: update examples

DaleSeo:
The `graphql/` directory has been renamed to `examples/` to better reflect its purpose and match the Rust ecosystem's convention of keeping example material under `examples/`. I've updated all internal path references in the config files, VS Code settings, and documentation accordingly.

The README for TheSpaceDevs example was outdated with old command line flags that we removed a while ago. It now includes three config variants: `config.yaml` for streamable HTTP, `config-stdio.yaml` for stdio, and `config-manifest.yaml` for the persisted query manifest. This makes it easier to switch between different config sets.

Unlike the TheSpaceDevs example, the weather example doesn't have a hosted endpoint and doesn't require local infrastructure. So, I've added a README to it to clarify that it needs both a local Apollo Router and the MCP server. The intended way to run it is with `rover dev --supergraph-config supergraph.yaml --mcp config.yaml`, which spins up both in a single command.

I've also simplified the VS Code launch and task configurations by removing redundant entries and consolidating them into debug-only configs that point to the new paths.


examples/TheSpaceDevs/config-manifest.yaml
gocamille, line 1: ✨

PR #684, opened 2026-03-05
Support @private in operations to hide parts of a query result from LLMs

andrewmcgivery:
Adds support for `@private` directive coming from client operations. The use case is to have certain data be "private" from the LLM, but available to the caller (main use case here is MCP Apps).

`@private` causes MCP Server to put the "full" response in `_meta` and the "public" (private fields removed) into `structuredContent`. Note that `@private` is stripped from the query before it is sent downstream to the GraphQL endpoint.

Example Query:

```graphql
query ProductQuery @tool {
    products {
        sku
        name
    }
    userPreferences @private {
        ...
    }
}
```

Resulting response: (note `userPreferences` not present)
```json
"structuredContent": {
    "result": {
        "data": {
            "products": {
                ....
            }
        }
    }
}
```

Resulting meta: (note entire response present
```json
"_meta": {
    "structuredContent": {
        "result": {
            "data": {
                "products": {
                    ....
                }
                "userPreferences": {
                    ...
                }
            }
        }
    }
}
```

### DIFF
--- a/crates/apollo-mcp-server/src/operations/private_fields.rs
+++ b/crates/apollo-mcp-server/src/operations/private_fields.rs
@@ -1,4 +1,4 @@
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 
 use apollo_compiler::{
     Node,
@@ -31,7 +31,8 @@ pub(crate) fn collect_private_fields(
     named_fragments: &HashMap<String, Node<FragmentDefinition>>,
 ) -> PrivateFieldTree {
     let mut tree = PrivateFieldTree::default();
-    collect_from_selections(selection_set, named_fragments, &mut tree);
+    let mut visited = HashSet::new();
+    collect_from_selections(selection_set, named_fragments, &mut tree, &mut visited);
     tree
 }
 
@@ -39,6 +40,7 @@ fn collect_from_selections(
     selection_set: &[Selection],
     named_fragments: &HashMap<String, Node<FragmentDefinition>>,
     tree: &mut PrivateFieldTree,
+    visited: &mut HashSet<String>,
 ) {
     for selection in selection_set {
         match selection {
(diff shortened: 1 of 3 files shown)


.changeset/support_private_in_operations_to_hide_parts_of_a_query_result_from_llms.md
```diff
@@ -0,0 +1,22 @@
+---
+default: minor
+---
+
+# Support @private in operations to hide parts of a query result from LLMs
```
jerelmiller, line 5: Throwing this out there to bikeshed on this name a bit (gives us time to noodle).

Should we consider using "model" in the name somewhere? I fear `@private` might communicate that this value is hidden from the app itself, which obviously isn't the case. Maybe `@modelPrivate`? Or going on the idea of a `visibility` directive, we could also provide something like:

```gql
@visibility(model: false)
```

Its definitely more verbose which is the big downside here.

Nothing to decide right now, but figured I'd open the bikeshedding discussion sooner than later 🙂. I'll build the client POC using `@private` for now to test out the idea.

andrewmcgivery: Yea, I think we'll have to document this no matter what cause it's a bit of a novel concept of using the directive to rip apart a graphql request like this.

I can't think of any case where we would have visibility to to anything other than private from the LLM so I'm leaning towards sticking with `@private` for simplicity sake... and as you mentioned before, if in the future we come up with a use case for something else with visibility, we can deprecate it.

crates/apollo-mcp-server/src/apps/tool.rs
```diff
@@ -82,45 +82,113 @@ async fn execute_app_tool(
...
+/// If the operation had `@private` fields, the full result was stashed in
+/// `meta.structuredContent` by `execute()` — remove and return it.
+/// Otherwise, fall back to cloning `structured_content` (which *is* the full result
+/// when no fields were filtered).
+fn take_full_result(result: &mut CallToolResult) -> Option<Value> {
+    result
+        .meta
+        .as_mut()
+        .and_then(|meta| meta.remove("structuredContent"))
+        .or_else(|| result.structured_content.clone())
```
DaleSeo, line 96: Isn't `.or_else(|| result.structured_content.clone())` always `None`?

crates/apollo-mcp-server/src/operations/private_fields.rs
```diff
@@ -0,0 +1,458 @@
...
+                    collect_from_selections(
+                        &fragment.selection_set,
+                        named_fragments,
+                        tree,
+                        visited,
+                    );
+                }
+            }
+            Selection::InlineFragment(inline) => {
+                collect_from_selections(&inline.selection_set, named_fragments, tree, visited);
```
DaleSeo, line 89: `@private` on the inline fragment itself is not supported?

crates/apollo-mcp-server/src/apps/tool.rs
andrewmcgivery: Nope! The comment above this function explains that `result.structured_content` will contain the full result when no fields were filtered during execute.

crates/apollo-mcp-server/src/operations/private_fields.rs
andrewmcgivery: Right now, no, it should only be supported on individual fields.

This may change/expand in the future but the "MVP" right now is on fields.

I'm expecting that the typical use case for this is going to be adding `@private` to a field with a sub-selection in most cases, similar to the example in the PR description!

PR #685, opened 2026-03-06
Add `allow_anonymous_mcp_discovery` setting

andrewmcgivery:
Add `allow_anonymous_mcp_discovery` setting to allow unauthenticated access to MCP discovery methods (e.g. `tools/list`) when oauth is enabled.

Note that in a future PR we'll likely add another condition that does this "by default" when serving apps, but for now, this serves the purpose for those who need this escape hatch right now, and for non-apps use cases.

From the comments of this PR:

The two use cases we received directly from users:

> While reading the docs on the new OpenAI App sdk, they specifically callout allowing ["noauth" tool discovery ](https://developers.openai.com/apps-sdk/build/auth#pertool-authentication-with-securityschemes). I've noticed this when setting up in other flows as well, where it's very convenient to be able to list and select tools during setup, before you're doing a full flow that has a user's token involved.

Relevant section in OpenAI's docs: https://developers.openai.com/apps-sdk/build/auth#triggering-authentication-ui

> Implement an enhancement to Apollo MCP Server's authentication middleware that enables anonymous tool discovery while maintaining user-level authorization for tool execution. This change significantly improves integration with agent frameworks like Google ADK while enhancing security through per-user audit trails.

I suspect this is going to become more common with MCP moving from being originally a "run stuff locally" protocol to primarily being a "run stuff remotely" protocol, where registration/discovery can become a separate step from execution, and a server may have a combination of authed/nonauthed tools (this PR does not address this last use case... but something @gocamille and I discussed on a call).

Also worth mentioning that the developer may be registering a server with a platform, which triggers tool discovery, and then _later_ the user actually uses the MCP Server and auths at that point. Without this option, the developer would need to have a "fake" (or real) user account in production to be able to complete the registration.

So... acknowledged that this is technically against the spec but the use cases here are very reasonable and I suspect this is a case of spec lagging behind real-world usage.

### DIFF
--- a/crates/apollo-mcp-server/src/auth.rs
+++ b/crates/apollo-mcp-server/src/auth.rs
@@ -3,7 +3,6 @@ use std::time::Duration;
 
 use axum::{
     Json, Router,
-    body::Bytes,
     extract::{Request, State},
     http::StatusCode,
     middleware::Next,


crates/apollo-mcp-server/src/auth.rs
```diff
@@ -271,6 +279,46 @@ impl Config {
     }
 }
 
+/// MCP discovery methods that are allowed without authentication when
+/// `allow_anonymous_mcp_discovery` is enabled.
+const ANONYMOUS_DISCOVERY_METHODS: &[&str] = &["initialize", "tools/list", "resources/list"];
```
DaleSeo, line 284: I'm concerned about spec compliance because authorization **MUST** be included in every HTTP request from the client to the server according to [the MCP auth spec](https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#access-token-usage). This means that even `initialize`, `tools/list`, and `resources/list` need a valid Bearer token and should follow the protected resource discovery. 

Should we support MCP clients that don't follow the spec? We've been quite strict about adhering to the auth spec.

docs/source/auth.mdx
```diff
@@ -184,10 +185,31 @@ Per the MCP specification, authorization servers must provide at least one of:
...
+| -------- | ------------------------------------------------------------------------ | -------------------------------------- |
+| 1        | `https://auth.example.com/.well-known/oauth-authorization-server/tenant` | RFC 8414 (path-insertion)              |
+| 2        | `https://auth.example.com/.well-known/openid-configuration/tenant`       | OIDC Discovery (path-insertion)        |
+| 3        | `https://auth.example.com/tenant/.well-known/openid-configuration`       | OIDC Discovery (legacy path-appending) |
 
 The first URL that returns a valid metadata document is used. This means IdPs that only support OIDC Discovery (such as Auth0 and many cloud identity providers) work out of the box without any additional configuration.
+
+## Anonymous MCP discovery
+
+Some MCP clients need to call discovery methods before they have obtained an OAuth token. By default, these methods require authentication like every other request. You can allow unauthenticated access to these methods by setting `allow_anonymous_mcp_discovery` to `true`:
```
DaleSeo, line 198: What are the MCP clients needing this exception?

I think we should mention that enabling this option deviates from the MCP spec's requirement that authorization be present on every request. A brief callout would help users understand the security trade-off.

crates/apollo-mcp-server/src/auth.rs
andrewmcgivery: The two use cases we received directly from users:

> While reading the docs on the new OpenAI App sdk, they specifically callout allowing ["noauth" tool discovery ](https://developers.openai.com/apps-sdk/build/auth#pertool-authentication-with-securityschemes). I've noticed this when setting up in other flows as well, where it's very convenient to be able to list and select tools during setup, before you're doing a full flow that has a user's token involved.

Relevant section in OpenAI's docs: https://developers.openai.com/apps-sdk/build/auth#triggering-authentication-ui

> Implement an enhancement to Apollo MCP Server's authentication middleware that enables anonymous tool discovery while maintaining user-level authorization for tool execution. This change significantly improves integration with agent frameworks like Google ADK while enhancing security through per-user audit trails.

I suspect this is going to become more common with MCP moving from being originally a "run stuff locally" protocol to primarily being a "run stuff remotely" protocol, where registration/discovery can become a separate step from execution, and a server may have a combination of authed/nonauthed tools (this PR does not address this last use case... but something @gocamille and I discussed on a call).

Also worth mentioning that the developer may be registering a server with a platform, which triggers tool discovery, and then _later_ the user actually uses the MCP Server and auths at that point. Without this option, the developer would need to have a "fake" (or real) user account in production to be able to complete the registration.

So... acknowledged that this is technically against the spec but the use cases here are very reasonable and I suspect this is a case of spec lagging behind real-world usage.

If it would help, I think your note below is a very reasonable add to the docs:

> I think we should mention that enabling this option deviates from the MCP spec's requirement that authorization be present on every request. A brief callout would help users understand the security trade-off.

DaleSeo: Thanks for the context, @andrewmcgivery! The platform registration and agent framework use cases make sense. It would be helpful to include them in the PR description or Jira ticket for future reference.

Do you think those use cases need `initialize` on this list? It seem to go beyond discovery into session establishment. Correct me I'm mistaken.

andrewmcgivery: Yea, `initialize` is part of the discovery process since it needs to be called prior to `tool/list`.

gocamille: Correct me if I'm wrong! But I think `initialize` needs to be here because without it, the client can't get a session going and without a session, `tools/list` and other future calls wouldn't work either.

Looks like in the [Streamable HTTP spec](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#session-management) the server assigns a session ID (Mcp-Session-Id) during `initialize`, and requests without that session ID get a `400 Bad Request`. So even if `tools/list` is on the anonymous list, it'd fail without the session ID from `initialize`.

andrewmcgivery: Correct, `initialize` always precedes `tools/list` :)

DaleSeo: @gocamille @andrewmcgivery I see a contradiction here.

To let clients discover tools before completing OAuth, I was imagining a flow like this:

1. Call discovery methods anonymously
2. Complete OAuth and get token
3. Then establish a proper authenticated session

But Camile's argument seems to undermine the framing. If `initialize` must be anonymous because it's required before `tools/list`, then the client is establishing a full MCP session before authenticating, not just doing lightweight discovery.

Am I missing something, or should the docs be clearer about this?

andrewmcgivery: There are actually two different flows involved here that I'll name as: "Discovery time" and "User time".

Discovery Time establishes an MCP session for the purposes of reading meta data about the server.

> `initialize` (unauthed) -> `tools/list` (unauthed) -> DONE

User Time would look more like this:

> `initialize` (unauthed) -> `tools/call` (auth challenge!) -> etc

So yes, I think you are correct that the user would effectively be establishing an MCP session without authentication until they try to call a tool which then requires authentication.

DaleSeo: Thanks for for the distinction, @andrewmcgivery! Now I understand why `initialize` should be in the list.

PR #688, opened 2026-03-10
Disable existing code review workflow, install rust best practices skill directly, setup REVIEW.md

andrewmcgivery:


.agents/skills/rust-best-practices/references/chapter_01.md
DaleSeo, line 1: Should we really return this reference book to the repo? This feels like a step backward. 😅 Instead, can we instruct Claude to install the skill and use it in REVIEW.md before doing the review?

andrewmcgivery: This isn't a copy like last time. This is installed using the skills `npx` command so we can more easily keep it up to date.

I also think having the skill installed on the repo is the right way to use skills anyways because we can also use it for our own local development.

DaleSeo: Hmm... 🤔 but this feels like checking in `node_modules` into version control. You usually only check in `package-lock.json`, right?

DaleSeo: I just noticed `npx skills experimental_install` which indicates that Vercel is working to introduce the same flow as `npm install` for installing skills from the lock file. OK, this should be fine for now.

ochoav:
Hey! I plugged this MCP server into our GraphQL API server at work. It's incredible!! Great job guys. 

Now I'm peeking around the repo and was surprised to see this REVIEW.md was just 3 hours old. Really sick to see team like you guys leaning into the agentic review and agentic flows. I'm curious about this rust best practices book you link to. Was that hand-written? Or a combo of the LLM'ing alongside the Official API guidelines you link to there?

PR #692, opened 2026-03-11
test: do tool calls improve with openai input schema style?

phryneas:
@jerelmiller has told me that a number of tools just don't get called right by some agents, so my suspicion is that these tools first try to "just generate the tool call", and if that doesn't work apply a step similar to the Vercel SDK's `experimental_repairToolCall`, in which they probably try to call the model's variant of `generateObject` with the `inputSchema`.
In the case of OpenAI models, the JSON Schema specification for that kind of call is *very* restricted, see [their documentation](https://developers.openai.com/api/docs/guides/structured-outputs/).

Most importantly: there are no optional fields - every field has to be `required`, and `additionalProperties` always needs to be `false`.

Our schemas currently don't fit these requirements - so let's just try if that improves the situation.

Obviously, this is not a fully functional PR, and tons of tests are still failing, but it should be enough for Jerel to build this locally and try out if the quality of tool calls improves.

Note: there are a lot of additional subtleties, so please inspect the generated schema if it really matches the OpenAI requirements before drawing any conclusions from this 😅

### DIFF
--- a/crates/apollo-mcp-server/src/json_schema.rs
+++ b/crates/apollo-mcp-server/src/json_schema.rs
@@ -16,14 +16,19 @@ macro_rules! schema_from_type {
 
 #[cfg(test)]
 mod tests {
-    use schemars::JsonSchema;
+    use schemars::{JsonSchema, Schema, SchemaGenerator};
     use serde::Deserialize;
     use serde_json::Value;
 
     #[derive(JsonSchema, Deserialize)]
     struct TestInput {
         #[allow(dead_code)]
-        field: String,
+        #[schemars(schema_with = "option_schema")]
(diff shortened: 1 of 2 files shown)


gocamille:
Hey @phryneas! Thanks for exploring this 🙏 After digging back into it, I see this changes argument validation after a tool is selected, whereas the problem was with the tool being called in the first place. It also seems to apply OpenAI-specific strict schema rules globally and changes GraphQL’s distinction between omitted values and `null`, which can break defaults.

Since we don’t have a reproduction showing this helps, I’m going to close it for now but please feel free to follow up if you have a repro and we can revisit it! ✨

PR #693, opened 2026-03-11
Support `extraOutputs` on tools in manifest file

andrewmcgivery:
Tools can now provide an `extraOutputs` argument. The content of this property is written as-is to the `extra` property under `structuredContent` on tool responses.


crates/apollo-mcp-server/src/apps/manifest.rs
```diff
@@ -140,6 +142,7 @@ pub(crate) fn load_from_path(
                     operation: operation.clone(),
                     labels,
                     tool,
+                    extra_outputs,
```
DaleSeo: It seems there's no good reason to assign this to `extra_outputs` above before using it here.

```suggestion
                    extra_outputs: tool.extra_outputs,
```

crates/apollo-mcp-server/src/apps/tool.rs
```diff
@@ -176,6 +178,10 @@ fn nest_app_tool_result(
         full_m.insert("toolName".into(), Value::String(tool_name.to_string()));
     }
 
+    if let Some(extra) = extra_outputs {
+        restricted_map.insert("extra".into(), extra.clone());
+    }
```
DaleSeo, line 205: The `toolName` field is written into both `restricted_map` and `full_m`, but `extra_outputs` is only written into `restricted_map`. Is this intentional?

crates/apollo-mcp-server/src/apps/manifest.rs
andrewmcgivery: This was being done because the `tool` variable was being overwritten with something totally different.

I've updated the name of the other variable so that this no longer happens and we can save the variable allocation. :)

crates/apollo-mcp-server/src/apps/tool.rs
andrewmcgivery: Yes! We only want `extra_outputs` on `structured_content` because it is meant to be shown to the LLM. It's not a meta property.

`toolName` _is_ a meta property... but we are writing it to both as a temporary workaround, per the comment on line 175:

```
// This is a temporary workaround because some MCP hosts don't properly expose _meta so we need the tool name to be available here as a backup
```

PR #698, opened 2026-03-13
fix: allow aarch64 musl installer fallback

DaleSeo:
The install script used to block installation on `aarch64` Linux systems with musl libc, showing an explicit error. However, we now provide a musl binary for that target. This change removes the block, allowing ARM64 musl systems to get the musl binary just like `x86_64` musl systems.

To make the installer easier to test, the main execution block is now protected by a `BASH_SOURCE` check. This way, the script can be sourced without causing side effects. We also added a new `test-install.sh` script that tests the architecture detection and download URL logic across different OS, CPU, and libc combinations. Additionally, there's a corresponding `installer` CI job that runs this script on every push.

### DIFF
--- a/.changeset/fix_aarch64_musl_installer_fallback.md
+++ b/.changeset/fix_aarch64_musl_installer_fallback.md
@@ -1,0 +1,9 @@
+---
+default: patch
+---
+
+# Fix aarch64 musl installer fallback
+
+The install script previously blocked installation on `aarch64` Linux systems with older glibc versions, even though a musl binary is now available for that target. This change removes the block, allowing ARM64 musl systems to fall back to the musl binary just like `x86_64` systems.
+
+The installer script is now testable via `source` without side effects, and a new `test-install.sh` validates architecture detection and download URL logic across OS, CPU, and libc combinations.


scripts/nix/install.sh
```diff
@@ -242,4 +236,6 @@ downloader() {
     fi
 }
 
-download_binary_and_run_installer "$@" || exit 1
+if [[ -z "${BASH_SOURCE[0]:-}" ]] || [[ "${BASH_SOURCE[0]}" == "$0" ]]; then
```
andrewmcgivery, line 239: Can you help me understand the purpose of this new if statement? Bash scripts are kinda hard to read sometimes 😅

DaleSeo: This guard lets the script be **sourced** by the test file (`test-install.sh`) without actually running the installer.

Here's how it works across the three ways this script can be invoked:

1. **`curl ... | bash` (piped):** `BASH_SOURCE[0]` is empty/unset, so the `-z` check is true → the installer **runs**. (The `:-` default value syntax prevents a crash from `set -u` on line 11.)
2. **`bash install.sh` or `./install.sh` (direct execution):** `BASH_SOURCE[0]` equals `$0` (both point to `install.sh`) → the installer **runs**.
3. **`source install.sh` (sourced from another script):** `BASH_SOURCE[0]` is `install.sh` but `$0` is the *sourcing* script (e.g. `test-install.sh`), so they differ and `-z` is false → the installer is **skipped**, but all the functions become available to the sourcing script.

So case 3 is the key motivation: `test-install.sh` can `source install.sh` to call individual functions like `get_architecture` in isolation without triggering the full download-and-install side effect.

---
_Generated by [Claude Code](https://claude.ai/code)_

DaleSeo: I just asked Claude Code locally, and it added this comment without my permission! 🙈
Does it help though, @andrewmcgivery?

Issue #699, opened 2026-03-16
Feature Request: Support MCP Prompts for GraphQL workflow guidance


wakita181009:
## Problem

Apollo MCP Server supports Tools and Resources but not MCP Prompts. `overrides.descriptions` works for individual tool guidance, but there's no way to express **cross-operation workflows** — e.g., "to get a user's order history, call GetUserByEmail → GetUserOrders (cursor-paginated) → GetOrderDetails in sequence."

This kind of orchestration knowledge exists in every GraphQL API but can't be captured in schema descriptions or per-tool overrides alone.

## Proposal

Support MCP Prompts via the config YAML:

```yaml
prompts:
  - name: check_order_history
    description: "Look up a user's recent orders"
    arguments:
      - name: email
        required: true
    messages:
      - role: user
        content: |
          1. GetUserByEmail(email={{email}}) → get userId
          2. GetUserOrders(userId, first=10) → cursor-paginated
          3. GetOrderDetails(orderId) for each order if needed
```

## Context

Claude Code already surfaces MCP Prompts as slash commands (`/mcp__server__promptname`). Adoption is growing across other clients too. This would complement the existing Tools + Resources support as the third MCP capability.

DaleSeo:
Thanks for the detailed proposal, @wakita181009! We've tracked this as internally. The workflow orchestration use case you described sounds compelling. We'll follow up here as we make progress.

wakita181009:
Thanks @DaleSeo. I went ahead and took a stab at implementing this in [#706](https://github.com/apollographql/apollo-mcp-server/pull/706). The PR adds a `prompts` config section that covers the proposed YAML format. Would love any feedback on the approach!

PR #706, opened 2026-03-21
feat: add MCP prompts support via YAML configuration 

wakita181009:
## Summary                                                                                                                                                                 

Issue [#699](https://github.com/apollographql/apollo-mcp-server/issues/699)
                                                                                                                                                                                                                                                                                 
- Add support for [MCP Prompts](https://modelcontextprotocol.io/docs/concepts/prompts) as the third MCP capability alongside Tools and Resources                           
- Prompts are defined in the YAML config under a new `prompts` key, with support for template arguments (`{{arg}}`), multiple content types (text, image, resource,
resource\_link), and role-based messages (user/assistant)                                                                                                                  
- Server advertises `PromptsCapability` and handles `list_prompts` / `get_prompt` requests when prompts are configured
                                                                                                                                                                           
## Example config
                                                                                                                                                                           
```yaml         
prompts:
  - name: check_order_history                                                                                                                                              
    description: "Look up a user's recent orders"
    arguments:                                                                                                                                                             
      - name: email
        required: true                                                                                                                                                     
    messages:   
      - role: user                                                                                                                                                         
        content:                                                                                                                                                           
          type: text                                                                                                                                                       
          text: |                                                                                                                                                          
            1. GetUserByEmail(email={{email}}) → get userId                                                                                                                
            2. GetUserOrders(userId, first=10) → cursor-paginated                                                                                                          
            3. GetOrderDetails(orderId) for each order if needed                                                                                                           
```
                                                                                                                                                      
Changes                                                                                                                                                                    
                                                                                                                                                                           
- prompts/config.rs — YAML-backed config types (PromptConfig, PromptMessageConfig, PromptContentConfig, etc.) with serde + JSON Schema support                             
- prompts/render.rs — Single-pass {{arg}} template engine with injection prevention (argument values are never re-evaluated)
- prompts/store.rs — Prompts store with validation (empty/duplicate name, required argument enforcement) and conversion to MCP protocol types                              
- errors.rs — PromptError variants for validation failures                                                                                                                 
- server/states/running.rs — list_prompts and get_prompt handler implementations; PromptsCapability advertised conditionally                                               
- runtime/config.rs — prompts field on Config with #[serde(default)]

### DIFF
--- a/crates/apollo-mcp-server/src/errors.rs
+++ b/crates/apollo-mcp-server/src/errors.rs
@@ -41,6 +41,25 @@ pub enum OperationError {
     Collection(CollectionError),
 }
 
+/// An error in prompt configuration
+#[derive(Debug, thiserror::Error)]
+pub enum PromptError {
+    #[error("Prompt definition has an empty name")]
+    EmptyName,
+
+    #[error("Duplicate prompt name '{0}'")]
+    DuplicateName(String),
+
+    #[error("Prompt not found: '{0}'")]
+    NotFound(String),
+
+    #[error("Missing required argument '{argument}' for prompt '{prompt_name}'")]
+    MissingRequiredArgument {
+        prompt_name: String,
+        argument: String,
+    },
+}
+
 /// An error in server initialization
 #[derive(Debug, thiserror::Error)]
 pub enum ServerError {
@@ -112,6 +131,9 @@ pub enum ServerError {
 
     #[error("There was a problem parsing Rhai scripts on startup.")]
     RhaiError,
+
+    #[error("Invalid prompt configuration: {0}")]
+    Prompt(#[from] PromptError),
 }
 
 /// An MCP tool error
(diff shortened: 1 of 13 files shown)


DaleSeo:
@wakita181009 Thank you for taking the time to implement MCP Prompts support!

After discussing this internally, we'd like to take a different approach. Instead of defining prompts inline in the YAML config, we're leaning toward what we already do with apps. In this model, prompts would be individual Markdown files in a `/prompts` directory.

The goal is to keep content out of the config. As the prompts list grows, embedding full message text inside YAML can become hard to manage. A file-per-prompt approach would look like this:

```markdown
<!-- /prompts/check_order_history.md -->
---
name: check_order_history
description: "Look up a user's recent orders"
arguments:
  - name: email
    required: true
---

1. GetUserByEmail(email={{email}}) → get userId
2. GetUserOrders(userId, first=10) → cursor-paginated
3. GetOrderDetails(orderId) for each order if needed
```

We also want to think carefully about scope. Looking at the use case, the primary need is a single, text-based message. So, we're considering whether we should initially limit support to that, rather than supporting multi-message prompts with image or resource content types from the start.

We have an internal ticket tracking this work. We want to ensure we provide you with clear feedback instead of leaving this open-ended. We're going to close this PR for now so we can pursue the file-based approach, but we genuinely appreciate you taking the initiative here. If you'd like to continue collaborating on this, we'd love your input as we design the new approach. 🙏

PR #732, opened 2026-04-22
feat: infer JWK alg from discovery metadata

### LINKED ISSUE #730: JWK "alg" required during auth token validation - Azure AD B2C
During auth token validation Apollo MCP Server seems to require the `alg` in the JWK to know what algorithm to use, which makes sense, however providers such as Azure AD B2C do not provide `alg` for each JWK in their `/discovery/v2.0/keys` endpoint.

I want to know if there is anyway Apollo MCP Server can support an optional `alg` in the JWK so the auth token validation can work with providers which don't supply the `alg` field.

Maybe the algorithm can be specified in the config `.yaml` file as an override or fallback value to use in these scenarios?

Thanks in advance for any help/advice going forward!

DaleSeo:
Fixes #730

Some identity providers (Microsoft Entra ID / Azure AD B2C, AWS Cognito, Ping Identity) deliberately omit the `alg` field on JSON Web Keys. [RFC 7517](https://datatracker.ietf.org/doc/html/rfc7517#section-4.4) permits this, but until now Apollo MCP Server rejected tokens signed by such keys because it had no way to pick a verification algorithm. This change resolves the algorithm automatically by reading `id_token_signing_alg_values_supported` from the authorization server's discovery document when a JWK omits `alg`.

### DIFF
--- a/crates/apollo-mcp-server/src/auth/networked_token_validator.rs
+++ b/crates/apollo-mcp-server/src/auth/networked_token_validator.rs
@@ -195,6 +195,11 @@ impl ValidateToken for NetworkedTokenValidator<'_> {
         self.upstreams
     }
 
+    /// `discovery_timeout` bounds each network stage (metadata fetch and JWKS
+    /// fetch) independently, so a cold-cache lookup can take up to 2×
+    /// `discovery_timeout` on the happy path. The JWKS fetch does not fall
+    /// back to alternate discovery URLs on failure; real providers advertise
+    /// the same `jwks_uri` from every well-known path.
     async fn get_key(&self, server: &Url, key_id: &str) -> Option<Jwk> {
         let metadata = discover_metadata(self.client, server, self.discovery_timeout).await?;
         let mut jwks = fetch_jwks(self.client, &metadata.jwks_uri, self.discovery_timeout).await?;
--- a/crates/apollo-mcp-server/src/auth/networked_token_validator.rs
+++ b/crates/apollo-mcp-server/src/auth/networked_token_validator.rs
@@ -1,7 +1,10 @@
+use std::str::FromStr;
 use std::time::Duration;
 
+use jsonwebtoken::jwk::KeyAlgorithm;
 use jwks::{Jwk, Jwks};
-use tracing::{info, trace, warn};
+use serde::Deserialize;
+use tracing::{error, info, trace, warn};
 use url::Url;
 
 use super::valid_token::ValidateToken;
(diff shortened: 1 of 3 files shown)


crates/apollo-mcp-server/src/auth/networked_token_validator.rs
```diff
@@ -160,9 +195,56 @@ impl ValidateToken for NetworkedTokenValidator<'_> {
...
+        .iter()
+        .filter_map(|s| {
+            KeyAlgorithm::from_str(s)
+                .inspect_err(
+                    |_| trace!(alg = %s, "Ignoring unrecognized signing algorithm from discovery"),
+                )
+                .ok()
+        })
+        .collect();
+
```
gocamille: I'm wondering if this current filter means `["RS256", "none"]` or `["RS256", "<unknown>"]` slips through as `RS256`, which could be a security concern. Do you feel it may be safer to gate on `advertised.len()` and let an unrecognized entry count as ambiguity?

crates/apollo-mcp-server/src/auth/networked_token_validator.rs
```diff
@@ -160,9 +195,56 @@ impl ValidateToken for NetworkedTokenValidator<'_> {
...
+    /// `discovery_timeout` on the happy path. The JWKS fetch does not fall
+    /// back to alternate discovery URLs on failure; real providers advertise
+    /// the same `jwks_uri` from every well-known path.
     async fn get_key(&self, server: &Url, key_id: &str) -> Option<Jwk> {
-        let jwks = discover_jwks(self.client, server, self.discovery_timeout).await?;
-        jwks.keys.get(key_id).cloned()
+        let metadata = discover_metadata(self.client, server, self.discovery_timeout).await?;
+        let mut jwks = fetch_jwks(self.client, &metadata.jwks_uri, self.discovery_timeout).await?;
+        let mut jwk = jwks.keys.remove(key_id)?;
+        if jwk.alg.is_none() {
```
gocamille, line 212: I am learning a lot about key signing right now! And one investigation makes me wonder about exposure in the `resolve_alg` function: if a discovery doc advertised just `["HS256"]` or something similar, we'd stamp `HMAC` onto an RSA key, which i think opens up the public-key-as-HMAC-secret attack? It might be worth restricting to the asymmetric algorithms (RS*/PS*/ES*/EdDSA) just to be safe, but let me know if i'm overthinking it!

crates/apollo-mcp-server/src/auth/networked_token_validator.rs
```diff
@@ -112,27 +115,44 @@ fn build_discovery_urls(issuer: &Url) -> Result<Vec<Url>, DiscoveryUrlError> {
...
-    let urls = match build_discovery_urls(issuer) {
-        Ok(urls) => urls,
-        Err(e) => {
-            warn!(error = %e, "Failed to build discovery URLs");
-            return None;
-        }
+/// Subset of the OIDC/OAuth discovery document that we consume.
+///
+/// `jwks_uri` locates the public keys, and `id_token_signing_alg_values_supported`
+/// lets us infer the signing algorithm when a JWK omits `alg` (RFC 7517 §4.4).
```
gocamille: Should we be explicit here about what we're reusing? 

```suggestion
/// Subset of the OIDC/OAuth discovery document that we consume.                                                                                
///                                                             
// `jwks_uri` locates the public keys. `id_token_signing_alg_values_supported`                                                                 
/// is strictly the algorithm list for OIDC ID tokens, but MCP validates OAuth 
/// access tokens — no standard discovery field advertises the access-token                                                                     
/// signing algorithm. In practice, the providers this fallback targets                                                                         
/// (Microsoft Entra ID, Azure AD B2C, AWS Cognito, Ping Identity) sign access                                                                  
/// tokens with the same algorithm they advertise here, so we reuse this field                                                                  
/// as a proxy to fill in `alg` when a JWK omits it (RFC 7517 §4.4).

crates/apollo-mcp-server/src/auth/networked_token_validator.rs
DaleSeo: Thanks for the feedback, @gocamille! You made me dig into [RFC 7517](https://datatracker.ietf.org/doc/html/rfc7517#section-4.1) which I appreciate. The spec prevents this because each JWK has the `kty` parameter which identifies the key's mathematical family. Any spec-compliant validator derives the verification primitive from `kty`, so an RSA key has no path into an HMAC routine regardless of what `alg` claims. I confirmed `jsonwebtoken` refuses forged tokens proceeding with the `InvalidKeyFormat` error. Given that, I'd prefer not to add an asymmetric whitelist, and I'd rather keep `resolve_alg` focused on the ambiguity check we actually need.

crates/apollo-mcp-server/src/auth/networked_token_validator.rs
DaleSeo: Good catch, @gocamille! I've updated `resolve_arg` to destructure on the raw slice.

crates/apollo-mcp-server/src/auth/networked_token_validator.rs
gocamille: Oooh this is good and more :til: -- I agree with this approach 👍

gocamille (approved):
:shipit:

PR #736, opened 2026-04-23
ci: run clippy on Windows to catch cfg-gated warnings

DaleSeo:
PR CI runs only on `ubuntu-latest`, so Windows-specific compile errors slip through review and only surface when the release-bins workflow tries to compile for Windows targets. This has now caused two broken release cuts in #719, #734 so it's worth closing the loop.

This PR fixes the latent Windows-only compile error found in the v1.13.0 release and adds a Windows clippy job to CI to detect this kind of issues earlier.

### DIFF
--- a/.github/workflows/release-bins.yml
+++ b/.github/workflows/release-bins.yml
@@ -52,6 +52,7 @@ jobs:
   build-artifacts:
     needs: [get-version]
     strategy:
+      fail-fast: false
       matrix:
         include:
           - target: aarch64-apple-darwin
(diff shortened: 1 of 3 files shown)


.github/workflows/release-bins.yml
```diff
@@ -52,6 +52,7 @@ jobs:
   build-artifacts:
     needs: [get-version]
     strategy:
+      fail-fast: false
```
DaleSeo, line 55: To only require rebuilding the broken target.

gocamille (approved):

PR #739, opened 2026-04-27
refactor: remove dead persisted_query support

DaleSeo:
Cleans up persisted-query infrastructure that was disabled but still wired through the codebase. The persisted-query branch in `graphql::execute()` was unreachable for both `Operation` and the introspection `Execute` tool. The companion `RawOperation::persisted_query_id` field was set during construction but its value never reached execution.


jerelmiller (approved):
🎉

PR #741, opened 2026-04-28
refactor: remove description from RawOperation

DaleSeo:
The `description` field on `RawOperation` was just a temporary step between `apply_description_override`, which looked it up by operation name in the config map, and `Operation::from_document`, which preferred it over the auto-generated description. No data source ever filled it, so it was really just a lookup from the config pretending to be part of the operation's state. This PR removes the field and directly integrates the descriptions map into `from_raw`, which we renamed from `from_document` since it now takes a `RawOperation` instead of a parsed document. This way, the lookup is done against the operation name derived from the AST.


gocamille (approved):
:shipit:

PR #742, opened 2026-04-29
fix: address rmcp DNS rebinding advisory

DaleSeo:
Bumps `rmcp` from 1.2 to 1.6 to pull in the fix for [GHSA-89vp-x53w-74fx](https://github.com/modelcontextprotocol/rust-sdk/security/advisories/GHSA-89vp-x53w-74fx) (CVE-2026-42559). With 1.6, host validation is enforced inside the Streamable HTTP transport, so the custom axum middleware we used to layer on top is no longer needed.

### DIFF
--- a/crates/apollo-mcp-server/Cargo.toml
+++ b/crates/apollo-mcp-server/Cargo.toml
@@ -30,7 +30,7 @@ http = "1.3.1"
 humantime-serde = "1.1.1"
 jsonschema = "0.42.0"
 jsonwebtoken = { version = "10.3", features = ["rust_crypto"] }
-jwks = { git = "https://github.com/chenhunghan/jwks", tag = "v0.5.3" }
+jwks = "0.5.3"
 lz-str = "0.2.1"
 opentelemetry = "0.31.0"
 opentelemetry-otlp = { version = "0.31.0", features = [
--- a/docs/source/config-file.mdx
+++ b/docs/source/config-file.mdx
@@ -285,7 +285,7 @@ These fields are under the `host_validation` key within the `transport` configur
 
 <Note>
 
-Host validation is only available when using the `streamable_http` transport. Localhost addresses (`localhost`, `127.0.0.1`, `::1`, `0.0.0.0`) are always allowed when validation is enabled.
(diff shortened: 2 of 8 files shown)


docs/source/config-file.mdx
```diff
@@ -285,7 +285,7 @@ These fields are under the `host_validation` key within the `transport` configur
 
 <Note>
 
-Host validation is only available when using the `streamable_http` transport. Localhost addresses (`localhost`, `127.0.0.1`, `::1`, `0.0.0.0`) are always allowed when validation is enabled.
+Host validation is only available when using the `streamable_http` transport. Loopback addresses (`localhost`, `127.0.0.1`, `::1`) are always allowed when validation is enabled.
```
DaleSeo, line 288: I dropped `0.0.0.0` to match the Rust SDK implementation. It only serves as a bind sentinel, meaning it listens on every interface, not as a destination. When a real client connects to a server bound to `0.0.0.0`, it resolves through a real interface and sends that in the Host header. The previous in-tree validator's check for `is_loopback() || is_unspecified()` mixed up the bind-side and destination-side semantics. The TypeScript and Go MCP SDKs also follows the same implementation.

DaleSeo:
I'm converting this to a draft. I noticed that `rcmp` doesn't emit tracing events. I put out a fix upstream in https://github.com/modelcontextprotocol/rust-sdk/pull/826. We'll upgrade it to v1.6 instead.

crates/apollo-mcp-server/Cargo.toml
```diff
@@ -30,7 +30,7 @@ http = "1.3.1"
 humantime-serde = "1.1.1"
 jsonschema = "0.42.0"
 jsonwebtoken = { version = "10.3", features = ["rust_crypto"] }
-jwks = { git = "https://github.com/chenhunghan/jwks", tag = "v0.5.3" }
+jwks = "0.5.3"
```
DaleSeo, line 33: JWKS v0.5.3 was published on crates.io so this git workaround is no longer needed.

andrewmcgivery (approved):

PR #744, opened 2026-05-01
docs: add docs page on error handling

DaleSeo:
Documents how Apollo MCP Server handles errors and clarifies cloud load balancer names in the deployment docs.


gocamille (approved):
I so appreciate the additional more in-depth context (and learned some things along the way)

PR #745, opened 2026-05-01
feat: expose trace_id to rhai graphql hook

### LINKED ISSUE #737: feat: Expose trace_id as a read-only property on the Rhai lifecycle hook context
## Summary

Add a `trace_id` read-only property to the `on_execute_graphql_operation` Rhai context object, allowing scripts to access the current OpenTelemetry trace ID for custom structured logging.

## Motivation

With PR #728 (released in 1.13.0), `trace_id` is now automatically prepended to server log lines via `TraceIdFormat`. However, there is no way to access this value **from within Rhai scripts**.

Our team uses Rhai's `on_execute_graphql_operation` hook to emit custom structured log lines with fields like `app_code`, `tool`, JWT claims, client identity, etc. We need `trace_id` as a field inside this structured log so our log aggregation platform (e.g., Splunk, ELK) can parse it as a discrete field for correlation -- not just as an unstructured prefix.

**Current workaround:** Parsing the `traceparent` header from `ctx.incoming_request.headers["traceparent"]`. This only works when the MCP client sends the W3C header, which is not guaranteed.

## Proposed API

```rhai
fn on_execute_graphql_operation(ctx) {
    let trace_id = ctx.trace_id;  // read-only, 32-hex string or "" if no span
    print(`[MCP Request] trace_id=${trace_id} tool=${ctx.tool_name}`);
}

DaleSeo:
Closes #737

The `on_execute_graphql_operation` Rhai hook now has a read-only `ctx.trace_id` property. This lets scripts access the current OpenTelemetry trace ID for custom structured logging. When an OpenTelemetry trace context is active, the value is a 32-character lowercase hex string. If it's not active, the value will be an empty string. This matches the format used for the `trace_id=<hex>` prefix in server log lines. Now, you can emit log lines from Rhai with `trace_id` as a separate field that log aggregators like Splunk and ELK can index for correlation with distributed traces.

### DIFF
--- a/.changeset/expose_trace_id_in_rhai_on_execute_graphql_operation.md
+++ b/.changeset/expose_trace_id_in_rhai_on_execute_graphql_operation.md
@@ -1,0 +1,7 @@
+---
+default: minor
+---
+
+# Expose `trace_id` to the `on_execute_graphql_operation` Rhai hook
+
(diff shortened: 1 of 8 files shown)


DaleSeo:
@mabuyo @pragl Could I get a review for the new docs?

jerelmiller (approved):

PR #746, opened 2026-05-04
fix(registry): coalesce file watch notifications instead of blocking watcher thread

### LINKED ISSUE #743: File watcher enters self-sustaining 20 Hz `no available capacity` log loop when consumer falls behind (no recovery without process restart)
## Summary

`crates/apollo-mcp-registry/src/files.rs` watches local schema/operations paths via `notify::PollWatcher` and signals changes on a `tokio::sync::mpsc::channel(1)`. The send-side callback runs a synchronous `loop { try_send → on Full, log warn + thread::sleep(50ms) }`. If the consumer side ever fails to drain a single signal — for example because the consumer task is parked on a slow GraphQL parse, deadlocked, was killed by a Tokio panic, or simply lost a scheduling race — the polling thread becomes permanently stuck in this retry loop, emitting `WARN could not process file watch notification. no available capacity` at exactly **20 Hz**, and never recovers without a process restart.

We hit this twice in production over a recent 8-day window. Both incidents required a manual `kubectl rollout restart` to clear; the bug is otherwise fully absorbing.

## Code reference

[`crates/apollo-mcp-registry/src/files.rs`](https://github.com/apollographql/apollo-mcp-server/blob/v1.7.0/crates/apollo-mcp-registry/src/files.rs#L41) (verified on `v1.7.0`, also unchanged on `main` at the time of writing):

```rust
let (watch_sender, watch_receiver) = mpsc::channel(1);   // ← capacity 1
…
let mut watcher = PollWatcher::new(
    move |res: Result<notify::Event, notify::Error>| match res {
        Ok(event) => {
            if matches!(event.kind, EventKind::Modify(…) | EventKind::Create(_) | EventKind::Remove(_)) {
                if event.paths.contains(&watched_path)
                    || (is_dir && event.paths.iter().any(|p| p.starts_with(&watched_path)))
                {
                    loop {
                        match watch_sender.try_send(()) {
                            Ok(_) => break,
                            Err(err) => {
                                tracing::warn!(
                                    "could not process file watch notification. {}",
                                    err.to_string()
                                );
                                if matches!(err, TrySendError::Full(_)) {
                                    std::thread::sleep(Duration::from_millis(50));   // ← 20 Hz retry
                                } else {
                                    panic!("event channel failed: {err}");
                                }
                            }
                        }
                    }
                }
            }
        }
        Err(e) => tracing::error!("event error: {:?}", e),
    },
    config,
);
```

## Why this is pathological

The `()` payload on this channel is a **wake-up notification**, not a content delivery: the consumer always re-reads the entire watched path when it receives one. The signal is therefore **idempotent** — if the channel already has an unread `()`, queuing another adds nothing, because the consumer will pick up the latest state on disk when it eventually drains.

The current implementation does the opposite: it **blocks the polling thread** waiting for the channel to drain, retrying every 50 ms. While this loop is active:

1. The polling thread cannot return to `PollWatcher`'s internal scheduler to run the next 3 s poll, so the FS-change detection itself stops making forward progress.
2. Each iteration emits a `WARN`. At 50 ms cadence that produces exactly **20 entries per second** — empirically what we observed (≈1198 entries/minute, sustained, for the whole duration of both incidents).
3. If the consumer is genuinely stuck (panicked Tokio task, mutex deadlock during schema/tool reload, runtime starvation), `try_send` returns `Full` forever and the polling thread spins indefinitely.

The state is **absorbing** because there is no path out of the loop except `try_send → Ok`, which requires the consumer to drain — and the consumer is precisely the thing that has stopped working.

## Reproduction

We were not able to construct a deterministic minimal repro on a laptop — the trigger appears to be a transient consumer slowdown (CPU contention from a noisy neighbour, a long GraphQL parse during schema reload, GC or scheduling pause in the upstream service we proxy to) that causes the consumer task to fall behind by **at least one event** during a moment of file change. Once that happens, the absorbing state forms and persists.

What we observe consistently in production:

1. The pod runs cleanly for hours/days; auth/tool traces show normal operation.
2. At some point, the WARN rate steps up to ~20/s within a single second and stays there.
3. While the WARN flood is active, our `oauth_validate` traces (on requests with `transport.auth` enabled) drop to zero — actual MCP requests are not being processed.
4. `kubectl rollout restart` immediately clears the state. The new pod runs cleanly for hours/days again, until the trigger condition recurs.

A synthetic repro should be reachable along the lines of:

- Configure schema source as a local file, with a supergraph SDL of non-trivial size (federated graphs in the multiple-MB range are realistic).
- Have an external process re-write the schema file every 100 ms, alternating between two valid versions so `with_compare_contents(true)` actually fires events.
- Constrain the process CPU to ~30 % of a core via cgroups so schema parsing in the consumer is slow.
- Watch the rate of the `no available capacity` WARN and the time until the watcher stops processing legitimate FS changes.

I haven't built this synthetic repro myself — the production fingerprint is already specific enough that I'm confident in the diagnosis, and I think the static analysis of the code (below) makes the pathology clear regardless.

## Suggested fix

The retry-and-block logic is treating the `()` signal as if delivery were required. It isn't: the signal is a coalescable wake-up, and `mpsc::channel(1)` is the right primitive **iff** the producer drops on full instead of retrying. Roughly:

```rust
match watch_sender.try_send(()) {
    Ok(_) => {}
    Err(TrySendError::Full(_)) => {
        // The consumer already has an unread wake-up signal queued; coalesce.
        // The signal is idempotent — losing it is fine, because the next
        // poll cycle (3 s by default) will re-detect the same content if
        // the consumer hasn't caught up yet.
        tracing::trace!("watch signal coalesced");
    }
    Err(TrySendError::Closed(_)) => {
        // Same panic semantics as today: this only happens once the
        // receiver stream has been dropped, which is a programming error.
        panic!("event channel closed: receiver gone");
    }
}
```

This:

- Removes the `loop` entirely.
- Removes the synchronous `thread::sleep` (which arguably never belonged inside a `notify` callback anyway, since it blocks the watcher's dispatch thread).
- Reclassifies what is now `WARN` to `TRACE` since coalescing is a normal occurrence, not an error condition.
- Preserves the panic on `Closed` for consistency with today's invariants.
- Is a pure ~10-line change in one function.

It does **not** paper over a stuck consumer — that remains a real concern (and probably deserves its own follow-up: e.g. a watchdog that logs at WARN if the consumer hasn't drained in N seconds, since today the only signal is the broken loop). But it stops a transient stuck consumer from cascading into a permanent 20 Hz log flood + dead watcher.

## Impact

In our production deployment we hit this twice in 8 days. The first incident lasted ~40 h before being noticed (only because end-user-facing symptoms were reported externally — the WARN flood by itself didn't page anyone until we set up a dedicated alert). The second hit a freshly-spawned pod within ~3 hours of starting up. We've since added a defensive 4-hour rolling-restart CronJob as a workaround, but the underlying state-machine bug is the right thing to fix.

We're posting this as the only public signal we could find for the symptom (zero matches on `"no available capacity"` across this repo and `apollographql/router`, where the comment in `files.rs` notes the code originated). Other production users running with `with_compare_contents(true)`, a non-trivial schema, and CPU constraints are probably hitting it too — they may just be restarting often enough that the absorbing state never has time to form.

## Offer

Per `CONTRIBUTING.md` I'm raising this as an issue first to get alignment on the approach before opening a PR. Happy to submit a small PR with:

1. The change above.
2. A unit test that simulates a "slow consumer" — e.g. a receiver that sleeps before draining — and asserts the producer's call site returns immediately on `Full` instead of entering a retry loop. (Existing tests in `files.rs` already use `tokio::test` and a temp file; this would extend that pattern.)
3. A short changelog/CHANGELOG.md note (the repo uses `knope` per `knope.toml`, so I'd follow whatever convention is in place there — happy to be guided.)

Open questions where I'd appreciate maintainer input:

- **WARN → TRACE downgrade**: would you prefer to keep the log at WARN (with a note about expected behaviour during slowdowns), or are you OK with TRACE? The current implementation emits this so often when triggered that even at normal `info` log levels it dominates output volume; downgrading seems clearly correct, but it's a UX/observability call I want to confirm.
- **Watchdog for stuck consumer**: should I include a basic watchdog in the same PR (e.g. log WARN once per 30 s if the channel has been Full for the entire window), or keep that as a follow-up?

Thanks for the great library — happy to iterate on the patch shape.

felipearomani:
## Summary

Fix the file-watcher self-sustaining log loop and watcher-thread starvation reported in #743.

`apollo-mcp-registry::files::watch_inner` retried `mpsc::Sender::try_send` in a busy loop with `std::thread::sleep(50ms)` when the consumer had not yet drained the previous event. Because the retry executes on the `notify::PollWatcher` callback thread, a slow consumer can pin that thread in the sleep/log loop and produce a continuous ~20 Hz stream of `could not process file watch notification. no available capacity` warnings indefinitely.

We've observed this in production: a single pod accumulated **~14k WARN/24h** of this message, sustained until manual restart, with the watcher effectively blocked from processing further events.

## Fix

The channel has capacity 1 and the consumer only needs to know *something changed*, not how many times — duplicate notifications are redundant. Replace the retry loop with a non-blocking `try_send` that:

- Drops queued duplicates at trace level (Full)
- Logs the unrecoverable case at error level instead of panicking from the watcher thread (Closed)

```rust
match watch_sender.try_send(()) {
    Ok(_) => {}
    Err(TrySendError::Full(_)) => {
        tracing::trace!("file watch notification coalesced (consumer falling behind)");
    }
    Err(err @ TrySendError::Closed(_)) => {
        tracing::error!("file watch channel closed: {err}");
    }
}
```

Net diff: +14/-13 lines (mostly comment).

## Test plan

- [x] `cargo test -p apollo-mcp-registry --lib files` — `basic_watch` and `recursive_watch_detects_subdir_changes` pass
- [x] `cargo clippy -p apollo-mcp-registry --all-targets -- -D warnings` — clean
- [x] `cargo fmt --check -p apollo-mcp-registry` — clean
- [ ] (No new test added: the bug is timing-sensitive and a deterministic regression test would be flaky. Existing tests cover the happy path. Behavior on `Full` is now explicitly defined: drop. Open to writing one if maintainers prefer.)

Closes #743

### DIFF
--- a/.changeset/fix_file_watcher_deadlock.md
+++ b/.changeset/fix_file_watcher_deadlock.md
@@ -1,0 +1,17 @@
+---
+default: patch
+---
+
+# Fix file-watcher self-sustaining log loop and watcher-thread starvation
+
+The file-change notifier in `apollo-mcp-registry` retried `mpsc::Sender::try_send`
+in a busy loop with `std::thread::sleep(50ms)` whenever the receiver had not yet
+drained the previous event. Because the retry runs on the `notify::PollWatcher`
+callback thread, a slow consumer can cause the callback to absorb itself in the
+sleep/log loop and emit a continuous ~20 Hz stream of `could not process file
+watch notification. no available capacity` warnings (issue #743).
+
+The channel has capacity 1 and the consumer only needs to know "something
+changed" — duplicate notifications are redundant. Replace the retry loop with a
+non-blocking `try_send` that drops a queued duplicate at trace level, and log
+the unrecoverable `Closed` case rather than panicking from the watcher thread.
--- a/crates/apollo-mcp-registry/src/files.rs
+++ b/crates/apollo-mcp-registry/src/files.rs
@@ -79,20 +79,21 @@ fn watch_inner(
                             watched_path
                         );
                     } else {
-                        loop {
-                            match watch_sender.try_send(()) {
-                                Ok(_) => break,
-                                Err(err) => {
-                                    tracing::warn!(
-                                        "could not process file watch notification. {}",
-                                        err.to_string()
-                                    );
-                                    if matches!(err, TrySendError::Full(_)) {
-                                        std::thread::sleep(Duration::from_millis(50));
-                                    } else {
-                                        panic!("event channel failed: {err}");
-                                    }
-                                }
+                        // Coalesce notifications: the channel has capacity 1 and the
+                        // consumer only needs to know "something changed", not how many
+                        // times. If the channel is full a notification is already queued,
+                        // so a duplicate is safe to drop. Looping with `std::thread::sleep`
+                        // here from inside the notify watcher thread could absorb the
+                        // thread under burst (issue #743).
+                        match watch_sender.try_send(()) {
+                            Ok(_) => {}
+                            Err(TrySendError::Full(_)) => {
+                                tracing::trace!(
+                                    "file watch notification coalesced (consumer falling behind)"
+                                );
+                            }
+                            Err(err @ TrySendError::Closed(_)) => {
+                                tracing::error!("file watch channel closed: {err}");
                             }
                         }
                     }


DaleSeo:
@claude review

crates/apollo-mcp-registry/src/files.rs
```diff
@@ -79,20 +79,21 @@ fn watch_inner(
...
+                        // thread under burst (issue #743).
+                        match watch_sender.try_send(()) {
+                            Ok(_) => {}
+                            Err(TrySendError::Full(_)) => {
+                                tracing::trace!(
+                                    "file watch notification coalesced (consumer falling behind)"
+                                );
+                            }
+                            Err(err @ TrySendError::Closed(_)) => {
+                                tracing::error!("file watch channel closed: {err}");
```
DaleSeo, line 96: I'm a bit confused about the intended handling `TrySendError::Closed` because the actual implementation in the PR is not aligned with the suggested fix in the issue. Since `Closed` here genuinely is a programming error, we might want to keeping failing fast.

Image

crates/apollo-mcp-registry/src/files.rs
```diff
@@ -79,20 +79,21 @@ fn watch_inner(
...
+                        // times. If the channel is full a notification is already queued,
+                        // so a duplicate is safe to drop. Looping with `std::thread::sleep`
+                        // here from inside the notify watcher thread could absorb the
+                        // thread under burst (issue #743).
+                        match watch_sender.try_send(()) {
+                            Ok(_) => {}
+                            Err(TrySendError::Full(_)) => {
+                                tracing::trace!(
+                                    "file watch notification coalesced (consumer falling behind)"
+                                );
```
DaleSeo, line 93: When this trace fires for an operator with multiple watchers active, the current message doesn't clarify which one is coalescing.

```suggestion
                                tracing::trace!(
                                    path = ?watched_path,
                                    "file watch notification coalesced (consumer falling behind)"
                                );
```

crates/apollo-mcp-registry/src/files.rs
```diff
@@ -79,20 +79,21 @@ fn watch_inner(
...
-                                    } else {
-                                        panic!("event channel failed: {err}");
-                                    }
-                                }
+                        // Coalesce notifications: the channel has capacity 1 and the
+                        // consumer only needs to know "something changed", not how many
+                        // times. If the channel is full a notification is already queued,
+                        // so a duplicate is safe to drop. Looping with `std::thread::sleep`
+                        // here from inside the notify watcher thread could absorb the
+                        // thread under burst (issue #743).
```
DaleSeo, line 87: Can we simplify this comment and move it above the `Err(TrySendError::Full(_))` arm? We usually don't include ticket numbers in the comments.

.changeset/fix_file_watcher_deadlock.md
```diff
@@ -0,0 +1,17 @@
...
+in a busy loop with `std::thread::sleep(50ms)` whenever the receiver had not yet
+drained the previous event. Because the retry runs on the `notify::PollWatcher`
+callback thread, a slow consumer can cause the callback to absorb itself in the
+sleep/log loop and emit a continuous ~20 Hz stream of `could not process file
+watch notification. no available capacity` warnings (issue #743).
+
+The channel has capacity 1 and the consumer only needs to know "something
+changed" — duplicate notifications are redundant. Replace the retry loop with a
+non-blocking `try_send` that drops a queued duplicate at trace level, and log
+the unrecoverable `Closed` case rather than panicking from the watcher thread.
```
DaleSeo, line 17: nit: Line breaks aren't needed in the changeset file.

DaleSeo (commented):
Thanks for your contribution, @felipearomani. I have a few comments.

PR #751, opened 2026-05-19
fix: index schema by field instead of by type

tninesling:
The schema index previously created one Tantivy doc per type with all field names concatenated, so length normalization buried specific operations (e.g. `slack_userByEmail`) under unrelated types whose own text matched more of the constituent tokens. Re-anchor the index on fields: one doc per object/interface field and per enum value, carrying the parent type, field name, arg names, return type, and description.

The `search` API (`Vec<Scored<PathNode>>`) is unchanged — paths still start at a root operation type. Path construction now anchors at the hit's parent type and walks up through the in-memory type-reference graph to a root, so existing consumers keep working.

Field-name term matches get a 3× boost since they are the most direct signal for the agent's typical "find an operation" workload. Adds a regression test using a synthetic noise schema that buries the target operation behind unrelated types under the old indexing.


crates/apollo-schema-index/src/lib.rs
```diff
@@ -47,11 +54,62 @@ pub mod error;
...
+
+/// An edge in the type-reference graph: "type X is referenced by parent_type's field_name
+/// (with these field_args)".
+#[derive(Clone, Debug)]
+struct ReferencingEdge {
+    parent_type: String,
+    /// `None` when the type is reached without a field name (e.g., union member, interface
+    /// implementer).
+    field_name: Option<Name>,
+    field_args: Vec<NamedType>,
```
DaleSeo: It looks like `field_args` is storing argument types instead of argument names or full definitions. So, using `field_arg_types` would make the path-building and tree-shaking intent clearer.

```suggestion
    field_arg_types: Vec<NamedType>,
```

crates/apollo-schema-index/src/lib.rs
```diff
@@ -47,11 +54,62 @@ pub mod error;
...
+    score: f32,
+}
+
+/// One indexable field on a type — produced by [`SchemaIndex::field_records`] and consumed
+/// by [`SchemaIndex::write_field_doc`].
+struct FieldRecord<'a> {
+    parent_type: &'a str,
+    field_name: &'a str,
+    arg_names: Vec<&'a str>,
+    arg_types: Vec<String>,
```
DaleSeo: Is there a reason why only this field is owned unlike the other fields? Since `FieldRecord` only exists long enough to write a Tantivy document, could `arg_types` possibly borrow from the schema?

crates/apollo-schema-index/src/lib.rs
```diff
@@ -187,35 +246,42 @@ impl SchemaIndex {
             .filter(Stemmer::new(Language::English))
             .build();
 
-        // Create the schema builder and add fields with the custom analyzer
+        let text_indexing = || TextFieldIndexing::default().set_tokenizer("en_stem");
```
DaleSeo, line 320: Could we lift `text_indexing` to the same abstraction level as the raw helper below? A `text_options` helper returning `TextOptions` would make the searchable and raw field setup more parallel.

```suggestion
        let text_options = || {
            TextOptions::default()
                .set_indexing_options(TextFieldIndexing::default().set_tokenizer("en_stem"))
        };
  ```

File: crates/apollo-schema-index/src/lib.rs [hunk 4]
```diff
@@ -47,11 +54,62 @@ pub mod error;
...
+struct DocFields {
+    parent_type_name: Field,
+    field_name: Field,
+    arg_names: Field,
+    return_type_name: Field,
+    description: Field,
+    parent_type_name_raw: Field,
+    field_name_raw: Field,
+    return_type_name_raw: Field,
+    field_args_raw: Field,
```
DaleSeo, line 179: `arg_names` above omits the `field_` prefix while `field_args_raw` here includes it.  Should both use or not use the same prefix unless this is intentional?

crates/apollo-schema-index/src/lib.rs
```diff
@@ -83,18 +141,14 @@ impl From<OperationType> for AstOperationType {
...
-    /// The maximum number of paths to root to include for each matching schema type
+    /// The maximum number of paths to root to include for each matching field hit
     pub max_paths_per_type: usize,
 
     /// The boost factor applied to shorter paths to root (0.0 for no boost, 1.0 for 100% boost)
     pub short_path_boost_factor: f32,
-
-    /// The percentage of the score of each parent type added to the overall score of the path
-    /// to root 0.0 for 0%, 1.0 for 100%)
-    pub parent_match_boost_factor: f32,
```
DaleSeo, line 97: Removing this option makes sense with field-anchored hits because scores are no longer keyed by type. Do we have coverage for queries where relevance is distributed across the path rather than concentrated in the matched field name? The new regression covers direct operation-name lookup, but it might be worth adding a test for a broader schema-navigation query to make sure removing parent path scoring does not regress those results.

crates/apollo-schema-index/src/lib.rs
```diff
@@ -255,117 +311,170 @@ impl SchemaIndex {
...
+            field_args_raw: field_args_raw_field,
+        };
+        let mut index_writer = index.writer(index_memory_bytes)?;
+        let mut field_count = 0usize;
+        for type_name in type_references.keys() {
             let type_name = NamedType::new_unchecked(type_name.as_str());
-            let extended_type = if let Some(extended_type) = schema.types.get(&type_name) {
-                extended_type
-            } else {
-                // This can never really happen since we got the type name from the schema above
```
DaleSeo, line 264: Could we keep this comment here? Without the comment, the silent continue reads like a normal expected case.

crates/apollo-schema-index/src/lib.rs
```diff
@@ -255,117 +311,170 @@ impl SchemaIndex {
...
+                    arg_types: Vec::new(),
+                    return_type: None,
+                    parent_description,
+                    field_description: value.description.as_ref().map(|n| n.as_str()).unwrap_or(""),
+                })
+                .collect(),
+            // Scalar and union types produce no documents. Their names remain partially
+            // recoverable via the `return_type_name` text on fields that reference them, but
+            // descriptions on custom scalars/unions are not searchable. Union member types
+            // surface in their own right as object fields/types.
```
DaleSeo: Is dropping standalone scalar and union docs an intentional tradeoff? Their names are still searchable through `return_type_name` on referencing fields, but the descriptions for custom scalars and unions don't seem to be searchable anymore.  If operation-building is the priority this may be fine, but I wanted to confirm we are comfortable losing the ability to find matches that only appear in scalar/union descriptions.

crates/apollo-schema-index/src/lib.rs
```diff
@@ -376,112 +485,64 @@ impl SchemaIndex {
...
-                                field_args,
-                                NamedType::new_unchecked(&type_name),
-                            ));
-                        }
-                    }
-                }
+        // For each top hit, anchor a path at the hit field and walk up to root operation
+        // types. Tantivy's BM25 score for the hit is the path's score; path-shape preferences
+        // (shorter paths preferred) are applied below.
+        for hit in hits.iter().take(options.max_type_matches) {
```
DaleSeo: Since the search unit is now a field hit rather than a type hit, should we rename this option accordingly?

```suggestion
        for hit in hits.iter().take(options.max_field_matches) {
```

crates/apollo-schema-index/src/lib.rs
```diff
@@ -376,112 +485,64 @@ impl SchemaIndex {
...
+            let leaf_path = self.build_leaf_path(hit);
+            for path in
+                self.walk_up_to_roots(&hit.parent_type, leaf_path, options.max_paths_per_type)
```
DaleSeo: I'd rename this option for the same reason.

```suggestion
                self.walk_up_to_roots(&hit.parent_type, leaf_path, options.max_paths_per_field_match)
```

crates/apollo-schema-index/src/lib.rs
```diff
@@ -255,117 +311,170 @@ impl SchemaIndex {
...
+                    parent_description,
+                    field_description: field.description.as_ref().map(|n| n.as_str()).unwrap_or(""),
+                })
+                .collect(),
+            ExtendedType::Enum(en) => en
+                .values
+                .iter()
+                .map(|(name, value)| FieldRecord {
+                    parent_type: en.name.as_str(),
+                    field_name: name.as_str(),
```
DaleSeo: This feels a bit awkward because enum values are being stored in `field_name`, even though they are not GraphQL fields.

crates/apollo-schema-index/src/lib.rs
```diff
@@ -496,6 +557,77 @@ impl SchemaIndex {
             .collect::<Vec<_>>())
     }
 
+    /// Build the leaf path segment for a field hit. For regular fields, the segment is
+    /// `parent_type --field_name--> return_type`. For enum values, the segment is
+    /// `parent_type -> value_name` with no field-name arrow.
+    fn build_leaf_path(&self, hit: &FieldHit) -> PathNode {
```
DaleSeo: Just like `field_records` and `write_field_doc`, this can be an associated function.

```suggestion
    fn build_leaf_path(hit: &FieldHit) -> PathNode {
```

crates/apollo-schema-index/src/lib.rs
```diff
@@ -496,6 +557,77 @@ impl SchemaIndex {
...
+                // `max_paths`).
+                out.push(current_path);
+                continue;
+            }
+            // Mark this intermediate type visited so we don't re-expand its edges in cycles
+            // through the type graph (e.g., `User.friend: User`). Roots are already handled
+            // above, so they never participate in cycle prevention.
+            if !visited.insert(current_type.clone()) {
+                continue;
+            }
```
DaleSeo: I think the global `visited` set can drop valid distinct paths. It tracks only the current type name, not the path that reached it, so two different paths that converge on the same intermediate type are collapsed.

For example:

```rs
const CONVERGING_PATH_SCHEMA: &str = r#"
    type Query {
        account: Account
    }

    type Account {
        primaryProfile: Profile
        secondaryProfile: Profile
    }

    type Profile {
        avatar: Image
    }

    type Image {
        url: String
    }
"#;

#[test]
fn search_preserves_distinct_paths_that_converge_on_an_intermediate_type() {
    let schema = Schema::parse(CONVERGING_PATH_SCHEMA, "converging_path.graphql")
        .expect("Failed to parse converging path schema")
        .validate()
        .expect("Failed to validate converging path schema");

    let index = SchemaIndex::new(&schema, OperationType::Query.into(), 15_000_000).unwrap();

    let results = index
        .search(
            vec!["avatar".to_string()],
            Options {
                max_paths_per_type: 10,
                ..Options::default()
            },
        )
        .unwrap();

    let paths: Vec<String> = results.iter().map(ToString::to_string).collect();

    assert!(
        paths
            .iter()
            .any(|path| path.contains("primaryProfile") && path.contains("avatar")),
        "Expected a path through primaryProfile.\nFound paths:\n{}",
        paths.join("\n")
    );
    assert!(
        paths
            .iter()
            .any(|path| path.contains("secondaryProfile") && path.contains("avatar")),
        "Expected a path through secondaryProfile.\nFound paths:\n{}",
        paths.join("\n")
    );
}
```

PR #757, opened 2026-05-29
feat: add issuer (iss) validation for OAuth tokens

Staticsubh:
## Summary

Apollo MCP Server validates a JWT's signature, expiry, and audience, but it never checks the **issuer (`iss`) claim** — unlike [Apollo Router's JWT authentication](https://www.apollographql.com/docs/router/configuration/authn-jwt), which supports issuer validation. This means that in a deployment delegating to multiple upstream OAuth servers, MCP cannot restrict which issuer a token must come from.

This PR adds **per-server** issuer validation: each configured authorization server can declare the `iss` values it is allowed to mint, and validation is bound to the server whose JWKS actually verified the signature. This brings the MCP Server to parity with the Router, which configures `issuers` per-JWKS entry.

> **Note:** This PR was originally written with a single global `transport.auth.issuers` list. Following [review feedback](https://github.com/apollographql/apollo-mcp-server/pull/757#discussion_r3343757827), it was reworked to bind issuers to each server. A global list applied the same allowlist to every server, so with multiple servers a token signed by server A could pass while claiming server B's issuer — and it did not match Router's per-JWKS semantics. The per-server design closes both gaps.

## Behavior

A `servers` entry is now **either** a bare URL string (unchanged behavior — no issuer check) **or** an object with a `url` and an optional per-server `issuers` list:

```yaml
transport:
  type: streamable_http
  auth:
    servers:
      - https://auth.example.com          # bare form — issuer validation skipped
      - url: https://auth.other.example.com
        issuers:
          - https://auth.other.example.com
```

- **Server entry has no `issuers` (bare string, or object without the field):** issuer validation is skipped for that server — existing configurations are completely unaffected.
- **Server entry lists `issuers`:** a token whose signature is verified by that server's keys must carry an `iss` claim matching one of that server's listed issuers, or the request is rejected with `401`.
- A token **missing the `iss` claim** while its signing server requires issuers is also rejected.
- Issuer validation is **bound to the signing server**: a token signed by one configured server cannot be accepted by claiming a different configured server's issuer.

## Implementation

- A new `AuthServer { url: String, issuers: Vec<String> }` type with a custom `Deserialize` that accepts either a bare URL string or `{ url, issuers }`. `auth::Config.servers` is now `Vec<AuthServer>`; the previous global `issuers` field is removed. `url` is kept as a `String` (not `Url`) for the same reason as before: `Url::parse` appends a trailing `/` to bare-authority inputs, which would break exact `iss` matching and alter the value advertised in the protected-resource metadata.
- `ValidateToken::get_issuers_for(server)` replaces the global `get_issuers()`; the `validate()` loop applies `Validation::set_issuer(...)` (and the missing-`iss` guard) using **only the current server's** issuers.
- Threaded through `NetworkedTokenValidator` (which now owns a parsed, index-aligned view of the configured servers).
- Ordering is preserved: signature and expiry are verified before the issuer check.

## Tests

- `iss` matches the signing server's configured issuer → valid
- `iss` matches **any** of a server's multiple configured issuers → valid
- `iss` does not match → rejected (`InvalidIssuer`)
- `iss` missing while the signing server requires issuers → rejected
- server has no issuers configured → token without `iss` still valid (backward-compat)
- **two configured servers, token signed by server A but claiming server B's issuer (shared `kid`) → rejected** (the cross-server regression test from review; verified to fail against the old global behavior)
- **two configured servers, each token validated against its own server's issuer → valid**
- config deserialization: bare-string form, object form with/without issuers, mixed list, invalid URL rejected

## Docs & changeset

- `docs/source/auth.mdx`: "Configuring allowed issuers" section, per-server shape.
- `docs/source/config-file.mdx`: `servers` table rows (`url` / `issuers`) + example.
- Changeset (`minor` — new backward-compatible feature).

`cargo fmt --check`, `cargo clippy --all-targets -- --deny warnings`, and the auth test suite all pass locally.

### DIFF
--- a/.changeset/add_issuer_validation.md
+++ b/.changeset/add_issuer_validation.md
@@ -4,8 +4,13 @@ default: minor
 
 # Add issuer validation for OAuth tokens
 
-Apollo MCP Server can now validate the `iss` (issuer) claim of incoming JWTs. Set
-`transport.auth.issuers` to a list of accepted issuer URLs; a token's `iss` claim must
-match one of them or the request is rejected. When `issuers` is empty (the default),
-issuer validation is skipped, so existing configurations are unaffected. This brings the
-MCP Server to parity with Apollo Router's JWT issuer validation.
+Apollo MCP Server can now validate the `iss` (issuer) claim of incoming JWTs. Issuer
+validation is configured per authorization server: a `transport.auth.servers` entry may
+now be written as an object with a `url` and an optional `issuers` list, in addition to
+the existing bare URL string form. When a server lists `issuers`, a token whose signature
+is verified by that server's keys must carry an `iss` claim matching one of the listed
+values, or the request is rejected. Binding issuers to the signing server (rather than a
+single global list) mirrors Apollo Router's per-JWKS `issuers` semantics and prevents a
+token signed by one configured server from being accepted while claiming a different
+configured server's issuer. Server entries written as bare URL strings skip issuer
+validation, so existing configurations are unaffected.
--- a/.changeset/add_issuer_validation.md
+++ b/.changeset/add_issuer_validation.md
@@ -4,13 +4,11 @@ default: minor
 
 # Add issuer validation for OAuth tokens
 
-Apollo MCP Server can now validate the `iss` (issuer) claim of incoming JWTs. Issuer
-validation is configured per authorization server: a `transport.auth.servers` entry may
-now be written as an object with a `url` and an optional `issuers` list, in addition to
-the existing bare URL string form. When a server lists `issuers`, a token whose signature
-is verified by that server's keys must carry an `iss` claim matching one of the listed
-values, or the request is rejected. Binding issuers to the signing server (rather than a
-single global list) mirrors Apollo Router's per-JWKS `issuers` semantics and prevents a
-token signed by one configured server from being accepted while claiming a different
-configured server's issuer. Server entries written as bare URL strings skip issuer
-validation, so existing configurations are unaffected.
+Apollo MCP Server can now validate the `iss` (issuer) claim of incoming JWTs. Set
+`transport.auth.issuers` to a list of accepted issuer values; a token's `iss` claim must
+match one of them or the request is rejected. Issuer validation is also bound to the
+authorization server that signed the token: the `iss` claim must equal the issuer that
+server advertises in its discovery metadata, so a token signed by one configured server
+cannot be accepted while it claims a different configured server's issuer. When `issuers`
+is empty (the default), issuer validation is skipped, so existing configurations are
+unaffected.
--- a/.changeset/add_issuer_validation.md
+++ b/.changeset/add_issuer_validation.md
@@ -1,0 +1,11 @@
+---
+default: minor
+---
+
+# Add issuer validation for OAuth tokens
+
+Apollo MCP Server can now validate the `iss` (issuer) claim of incoming JWTs. Set
+`transport.auth.issuers` to a list of accepted issuer URLs; a token's `iss` claim must
+match one of them or the request is rejected. When `issuers` is empty (the default),
+issuer validation is skipped, so existing configurations are unaffected. This brings the
+MCP Server to parity with Apollo Router's JWT issuer validation.
--- a/crates/apollo-mcp-server/src/auth.rs
+++ b/crates/apollo-mcp-server/src/auth.rs
@@ -146,6 +146,15 @@ pub struct Config {
     #[serde(default)]
     pub audiences: Vec<String>,
 
+    /// List of accepted token issuers (the JWT `iss` claim).
+    ///
+    /// When non-empty, a token's `iss` claim must exactly match one of these
+    /// values or the token is rejected. When empty (default), issuer validation
+    /// is skipped. Not `Vec<Url>`: like `servers`, `Url::parse` appends `/` to
+    /// bare-authority inputs, which would break exact `iss` string matching.
+    #[serde(default)]
+    pub issuers: Vec<String>,
+
     /// Allow any audience (skip validation) - use with caution
     #[serde(default)]
     pub allow_any_audience: bool,
@@ -484,6 +493,7 @@ async fn oauth_validate(
         .collect();
     let validator = NetworkedTokenValidator::new(
         &auth_config.audiences,
+        &auth_config.issuers,
         auth_config.allow_any_audience,
         &auth_servers,
         &auth_state.client,
@@ -580,6 +590,7 @@ mod tests {
         Config {
             servers: vec!["http://localhost:1234".to_string()],
             audiences: vec!["test-audience".to_string()],
+            issuers: vec![],
             allow_any_audience: false,
             resource: Url::parse("http://localhost:4000").unwrap(),
             resource_documentation: None,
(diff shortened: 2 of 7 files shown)


Staticsubh:
### Status update & CI note

Thanks for the automated checks. A couple of clarifications, plus a process correction on my side.

**Process:** I opened this PR before reading `CONTRIBUTING.md`, which asks contributors to open an issue and get maintainer agreement on the approach before submitting a feature PR. Apologies for the ordering — I've now opened **#759** describing the issuer-validation feature and the Apollo Router parity rationale. Happy to align there first and revise this PR (or close it) based on the team's guidance.

**On the two failing checks** — both are fork-PR infrastructure limitations rather than problems with the change itself:

- **`Verify`** — the changeset check itself passes (`Verify changeset included` ✅). The job fails only on the subsequent *"Add or update changeset comment on success"* step, which tries to `POST` a comment and gets `403 Resource not accessible by integration` because a fork PR's `GITHUB_TOKEN` is read-only (`issues:write`/`pull_requests:write` aren't granted to forks).

- **`End-to-End Tests`** — the Docker image builds successfully from the current code; the job then fails inside `google-github-actions/auth` with *"GitHub Actions did not inject `$ACTIONS_ID_TOKEN_REQUEST_TOKEN` … this job is being run from a fork."* The e2e smoke test needs an OIDC token to pull a private GCP Artifact Registry image, and fork PRs don't receive that token/secret.

All code-quality and test gates that *can* run on a fork are green: **Run Tests** (`cargo test`), **Build**, **Run Clippy checks** (and Windows), **Run Cargo Format**, **Run Cargo Doc**, **codecov/patch**, **codecov/project**, and **AI Style Review**. The CLA is signed.

I've also verified locally that the feature works end-to-end: running the server with `transport.auth.issuers` set, a token whose `iss` matches authenticates and opens an MCP session, while an otherwise-identical token with a mismatched `iss` is rejected with `401` (`InvalidIssuer`).

If it's helpful, a maintainer re-running these two jobs from within the repo (where secrets/OIDC are available) should clear them. Let me know how you'd like to proceed.

crates/apollo-mcp-server/src/auth/valid_token.rs
```diff
@@ -149,6 +157,10 @@ pub(super) trait ValidateToken {
                     val.set_audience(self.get_audiences());
                 }
 
+                if !self.get_issuers().is_empty() {
+                    val.set_issuer(self.get_issuers());
+                }
+
```
DaleSeo, line 181: This applies the same global issuer list to every configured auth server. With multiple servers, a token can be signed by server A's JWKS but claim `iss = "https://server-b.example.com"` and still pass if server B is in `transport.auth.issuers`.

Can we bind issuer validation to the JWKS/discovery metadata that verified the signature instead? A regression test with two configured servers and a mismatched signing server vs. `iss` would make this behavior clear.

DaleSeo (commented):
Thanks for adding issuer validation, @Staticsubh. This seems like a good feature to tighten security. I have one concern though. Let me know if it makes sense.

.changeset/add_issuer_validation.md
```diff
@@ -0,0 +1,11 @@
...
+default: minor
+---
+
+# Add issuer validation for OAuth tokens
+
+Apollo MCP Server can now validate the `iss` (issuer) claim of incoming JWTs. Set
+`transport.auth.issuers` to a list of accepted issuer URLs; a token's `iss` claim must
+match one of them or the request is rejected. When `issuers` is empty (the default),
+issuer validation is skipped, so existing configurations are unaffected. This brings the
+MCP Server to parity with Apollo Router's JWT issuer validation.
```
DaleSeo: I noticed one discrepancy with Router's current `issuers` behavior. Router validates `iss` only when the claim is present, while this PR rejects missing `iss` when issuers are configured. I'll follow up with the Router team to confirm the intended behavior.

crates/apollo-mcp-server/src/auth/valid_token.rs
Staticsubh: Hi @DaleSeo,
You are absolutely right, thanks for catching this. The global list was getting applied at every server in the loop, so `set_issuer` was never tied to the JWKS that actually verified the signature — with two upstreams sharing a `kid`, a token signed by server A could sneak through claiming `iss = server-B`.

While fixing it I also realised the global shape was missing the Router parity itself: Router keeps `issuers` per-JWKS entry (`authentication.router.jwt.jwks[].issuers`), not one flat list. So binding the issuers to the signing server gives us both the security fix and proper parity.

I have reworked it accordingly:

- `issuers` now sits on each server. A `servers` entry is either a bare URL string (no change, no issuer check) or an object `{ url, issuers: [...] }`. The bare-string form still parses, so existing configs are untouched.
- Dropped the top-level `transport.auth.issuers` field — it was only added in this PR, so no compatibility concern, and keeping it would only bring back the same ambiguity.
- In `validate()`, only the current server's issuers are applied, so the check stays scoped to the JWKS that verified the signature.
- Added the two-server regression test you asked for (signed by A, claiming B's issuer with a shared `kid` → rejected), plus a positive two-server test. Do note, I confirmed the rejection test actually fails against the old global behaviour, so it is genuinely guarding the fix.
- Updated `auth.mdx`, `config-file.mdx` and the changeset for the per-server shape.

Kindly have a look whenever you get a chance.

crates/apollo-mcp-server/src/auth.rs
```diff
@@ -118,29 +118,89 @@ where
...
+    ///
+    /// Each entry is either a bare URL string or an object with `url` and an
+    /// optional per-server `issuers` list (the accepted JWT `iss` values for
+    /// tokens that server signs). Issuer validation is bound to the server
+    /// whose JWKS verifies a token's signature, matching Apollo Router's
+    /// per-JWKS `issuers` semantics.
     #[serde(deserialize_with = "deserialize_auth_servers")]
-    #[schemars(with = "Vec<Url>")]
-    pub servers: Vec<String>,
+    pub servers: Vec<AuthServer>,
```
DaleSeo: @Staticsubh Thanks for the rework and the regression test. That's exactly the coverage I wanted.

I think we should stick with the optional flat `issuers` field you proposed initially. Reshaping the public config like this would involve the docs team, require updates to our educational materials, and need a major version bump, which we can't justify for just one feature. Our config has never mirrored Router's so I'd keep `servers` as it is.

My original suggestion was to link issuer validation to the discovery metadata that verifies the signature. We already fetch it in `get_key` and currently drop the `issuer` field. Matching `iss` against it gives us the same cross-server guarantee without affecting `servers`, and your test should pass as it is.

Please let me know if this direction makes sense.

Staticsubh:
Resolved — implemented per your suggestion. See commit 1c84572 on the PR branch.

crates/apollo-mcp-server/src/auth/valid_token.rs
Staticsubh: Done — addressed in commit 1c84572 (discovery binding + regression test).

crates/apollo-mcp-server/src/auth.rs
Staticsubh: Hi @DaleSeo ,
  
   That makes sense, thanks — keeping servers as it is and avoiding a major
   version bump for a single feature is a fair call. I have reworked it along
   the lines you suggested.
  
   I confirmed your point about the discovery metadata: we already fetch the
   document in get_key, but DiscoveryMetadata only deserialised jwks_uri and
   id_token_signing_alg_values_supported, so the issuer field was being
   dropped. I have captured it and now bind the token's iss to the issuer
   advertised by the server whose JWKS verified the signature. RFC 8414 makes
   that issuer required and identical to the server's issuer identifier, and a
   compliant iss matches it, so the cross-server guarantee holds without
   touching servers.

crates/apollo-mcp-server/src/auth/valid_token.rs
```diff
@@ -162,6 +193,46 @@ pub(super) trait ValidateToken {
...
+                        // discovered issuer identifier. This prevents a token signed
+                        // by one configured server from being accepted while claiming
+                        // a different configured server's issuer. Fail closed if the
+                        // server's issuer could not be determined.
+                        // Bind the issuer to the server whose JWKS verified the
+                        // signature: the token's `iss` must equal that server's
+                        // discovered issuer identifier. This prevents a token signed
+                        // by one configured server from being accepted while claiming
+                        // a different configured server's issuer. Fail closed if the
+                        // server's issuer could not be determined.
```
DaleSeo: @Staticsubh Can we avoid repeating the same comment block?

crates/apollo-mcp-server/src/auth/valid_token.rs
```diff
@@ -162,6 +193,46 @@ pub(super) trait ValidateToken {
...
+                                );
+                                break;
+                            }
+                            None => {
+                                warn!(
+                                    "Cannot verify token `iss`: the signing server did not advertise an issuer"
+                                );
+                                break;
+                            }
+                        }
```
DaleSeo: `DiscoveryMetadata.issuer` is required, while `VerificationKey.issuer` is optional. Making the field a `String` instead of `Option<String>` will eliminate the dead branch and collapses the `match` to a single comparison.

```suggestion
                        if discovered_issuer != token_issuer {
                            warn!(
                                token_issuer = %token_issuer,
                                server_issuer = %discovered_issuer,
                                "Token `iss` does not match the issuer of the server that signed it"
                            );
                            break;
                        }
```

crates/apollo-mcp-server/src/auth/valid_token.rs
Staticsubh: Done — removed the duplicate block in 1db4db2.

crates/apollo-mcp-server/src/auth/valid_token.rs
Staticsubh: Good catch — since `DiscoveryMetadata.issuer` is required, `VerificationKey.issuer` as `Option<String>` was unnecessary. Changed to `String` and collapsed the match to a single `if` comparison in 1db4db2.

DaleSeo (approved):

PR #760, opened 2026-06-09
refactor: restructure auth token validation internals

DaleSeo:
This PR splits the old `ValidateToken` trait into a one-method `KeyResolver`, which is all the tests need to substitute for the network. There's also a plain `TokenValidator` struct that holds the validation data and contains the single `validate` implementation. This change removes the duplicated accessor boilerplate that both the production and test validators had, making it easier to add future validation inputs as struct fields instead of modifying a trait and two implementations.

Additionally, it now parses the configured server URLs once at startup and shares `Config` via `Arc`. This means an authenticated request no longer deep-clones the config or re-parses URLs during the hot path.

### DIFF
--- a/crates/apollo-mcp-server/src/auth.rs
+++ b/crates/apollo-mcp-server/src/auth.rs
@@ -16,7 +16,7 @@ use axum_extra::{
     headers::{Authorization, authorization::Bearer},
 };
 use http::Method;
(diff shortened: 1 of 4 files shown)


jerelmiller (approved):
💯

DaleSeo:
@claude What is the state of this PR?

PR #762, opened 2026-06-09
fix: retain tool catalog on transient Uplink manifest fetch failure

### LINKED ISSUE #761: bug: transient Uplink persisted-query manifest fetch failure silently clears tool catalog while /health reports UP
## Problem Statement

When `operations.source: uplink` is configured, a transient Uplink persisted-query manifest fetch failure (network timeout, DNS failure, HTTP 5xx, retryable `retry_later` response) causes the server to replace its active tool catalog with an empty list. The server continues to report `{"status": "UP"}` from `/health`, sends a `tools/list_changed` notification to all connected clients, and those clients receive `{"tools": []}` — with no error, no degraded signal, and no indication that a failure occurred.

This means the failure is **invisible to all standard monitoring** (health checks, orchestrator readiness probes, alerting) while silently breaking all AI agent workflows that depend on MCP tools from this server.

## Root Cause

`manifest_poller.rs` line 44–47: any `Err` from the Uplink manifest stream is converted directly into `Event::UpdateManifest(vec![])` — an authoritative empty manifest update:

```rust
Err(e) => {
    tracing::error!("error from manifest stream: {}", e);
    Event::UpdateManifest(vec![])  // transient error → empty catalog
}
```

This propagates as `OperationsUpdated([])` to the running server, which unconditionally replaces the catalog (`running.rs:187`: `*operations_lock = updated_operations`). The server stays healthy, sends `tools/list_changed` to clients, and those clients get `{"tools": []}`.

Sources of transient `Err` that trigger this path (all in `uplink.rs`):
- `Error::FetchFailedSingle` / `FetchFailedMultiple` — all Uplink endpoints unreachable
- `UplinkResponse::Error { retry_later: true }` — explicitly retryable, yet still converted to empty catalog on the current tick
- Failed signed-URL chunk fetch after all fallback URLs are exhausted

**The asymmetry:** `CollectionError` already has keep-last-good behavior in `states.rs:205–212`:
```rust
ServerEvent::CollectionError(e) => match state {
    State::Running(running) => {
        tracing::error!("Collection error while running, keeping existing operations: {e}");
        running.into()  // catalog is NOT replaced
    }
    ...
}
```
The Uplink/manifest path has no equivalent — errors are never surfaced as an error variant because they are converted to an empty `UpdateManifest` before reaching the state machine.

## Exact Source Locations

| Location | Issue |
|---|---|
| `crates/apollo-mcp-registry/src/uplink/persisted_queries/manifest_poller.rs:44–47` | `Err` → `UpdateManifest(vec![])` — the core bug |
| `crates/apollo-mcp-registry/src/uplink/persisted_queries/manifest_poller.rs:199–205` | `Ok(None)` → `PersistedQueryManifest::default()` (empty) — secondary concern |
| `crates/apollo-mcp-server/src/server/states/running.rs:187` | Unconditional full replacement: `*operations_lock = updated_operations` |
| `crates/apollo-mcp-server/src/server/states.rs:196–201` | No guard against empty `OperationsUpdated` while Running |

## Impact

- **Silent availability degradation**: AI agents calling `tools/list` receive `{"tools": []}` and cannot distinguish "intentionally empty" from "fetch failed." Workflows silently no-op or fail upstream with no trace back to the MCP server.
- **Health endpoint is misleading**: `/health` returns `UP` during the degradation window. Operators relying on health-based alerting or K8s readiness probes receive no signal.
- **`tools/list_changed` actively misleads clients**: The server sends a notification telling clients to re-fetch — they do, and receive an empty list with no error context.
- **No recovery signal**: When Uplink recovers and tools reappear there is no log entry marking the degradation window, making post-incident analysis difficult.

## Expected Behavior

| Uplink response | Expected server behavior |
|---|---|
| HTTP 200, non-empty manifest | Apply new catalog (current — correct) |
| HTTP 200, empty manifest | Clear catalog (intentional operator action — correct) |
| Network error / timeout | Retain last known good catalog; log structured error |
| HTTP 5xx from Uplink | Retain last known good catalog; log structured error |
| `retry_later: true` response | Retain catalog; schedule retry without modifying catalog |

## Proposed Fix

Mirror the existing `CollectionError` keep-last-good pattern:

1. Add a `ManifestError(BoxError)` variant to the manifest event enum.
2. In `manifest_poller.rs`, emit `ManifestError(e)` on `Err` instead of `UpdateManifest(vec![])`.
3. Propagate `ManifestError` as a new `ServerEvent::ManifestError` (alongside the existing `CollectionError`).
4. In the state machine, handle `ManifestError` while `Running` by retaining the existing catalog and logging the error — exactly as `CollectionError` is handled. Treat it as fatal during startup (before the first successful load).

This fix distinguishes transient infrastructure failure from an authoritative "operator cleared the collection" signal (HTTP 200 with a valid empty manifest), which should continue to clear the catalog.

## Regression Test

1. Seed a running server with a non-empty manifest via a successful Uplink mock.
2. Simulate a subsequent fetch failure (timeout / 5xx).
3. Assert `tools/list` still returns the previously loaded tools.
4. Simulate recovery (successful fetch with a new manifest).
5. Assert `tools/list` returns the new tool set.

Contrast test: HTTP 200 with an empty manifest should still clear the catalog (intentional empty).

## Configuration

- Affected: `operations.source: uplink`
- Transport: all (stdio and streamable_http)
- Version: confirmed in v1.14.0

## Not the Root Cause (Ruled Out)

- **Multiple Okta token scopes**: `scp` array and `scope` string forms both parse correctly; no failure path exists in `valid_token.rs`. `tools/list` is not scope-filtered — scope checks only gate `tools/call`.
- **Schema reload race**: possible but distinct scenario.
- **Mutation mode dropping tools**: possible but distinct — would affect all sources, not just Uplink.

Staticsubh:
## Summary

Closes #761.

When `operations.source: uplink` is configured, a transient Uplink persisted-query manifest fetch failure (network timeout, DNS failure, HTTP 5xx, retryable `retry_later` response) previously caused the server to replace its active tool catalog with an empty list. The server continued reporting `/health` UP, sent `tools/list_changed` to connected clients, and those clients received `{"tools": []}` with no error signal.

This PR fixes the bug by mirroring the existing `CollectionError` keep-last-good pattern for manifest fetch failures.

## Root Cause

`manifest_poller.rs:44–47` converted any `Err` from the Uplink stream into `Event::UpdateManifest(vec![])` — treating a transient infrastructure failure as an authoritative "empty collection" signal:

```rust
// Before (bug): error → silently wipes catalog
Err(e) => {
    tracing::error!("error from manifest stream: {}", e);
    Event::UpdateManifest(vec![])
}
```

## Changes

- **`apollo-mcp-registry`** — Added `ManifestError(BoxError)` variant to the manifest event enum. `manifest_poller.rs` now emits `ManifestError(e)` on `Err` instead of `UpdateManifest(vec![])`.
- **`apollo-mcp-server`** — Added `ManifestError(BoxError)` to `ServerEvent` and `OperationError`. `operation_source.rs` forwards `ManifestError` through the event stream. State machine handles `ManifestError` while `Running` by retaining the existing catalog and logging the error (exactly as `CollectionError` is handled); treats it as fatal during startup before the first successful load.
- **Tests** — Two new regression tests in `states.rs`: one confirming the server stays alive with its catalog intact on `ManifestError` while Running; one confirming it is fatal during startup.

## Behavior

| Uplink response | Before | After |
|---|---|---|
| HTTP 200, non-empty manifest | Apply catalog ✅ | Apply catalog ✅ |
| HTTP 200, empty manifest | Clear catalog ✅ | Clear catalog ✅ |
| Network error / timeout | **Silent empty catalog** ❌ | Retain catalog, log error ✅ |
| HTTP 5xx / `retry_later` | **Silent empty catalog** ❌ | Retain catalog, log error ✅ |

## Test Plan

- [ ] `cargo test -p apollo-mcp-server manifest_error` — both new regression tests pass
- [ ] `cargo test -p apollo-mcp-server collection_error` — existing tests still pass
- [ ] `cargo clippy --all-targets -- --deny warnings` — no new warnings
- [ ] Manual: start server with `operations.source: uplink`, simulate Uplink unreachable, confirm `tools/list` returns previous tools and `/health` logs a structured error

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/apollo-mcp-registry/src/uplink/persisted_queries/manifest_poller.rs
+++ b/crates/apollo-mcp-registry/src/uplink/persisted_queries/manifest_poller.rs
@@ -41,14 +41,7 @@ impl ManifestSource {
                             .map(|(k, v)| (k.operation_id.clone(), v.clone()))
                             .collect(),
                     ),
-                    Err(e) => {
-                        tracing::error!(
-                            "transient error fetching persisted query manifest; \
-                             retaining previous catalog: {}",
-                            e
-                        );
-                        Event::ManifestError(e)
-                    }
+                    Err(e) => Event::ManifestError(e),
                 })
                 .boxed(),
             Err(e) => {
--- a/crates/apollo-mcp-registry/src/uplink/persisted_queries/manifest_poller.rs
+++ b/crates/apollo-mcp-registry/src/uplink/persisted_queries/manifest_poller.rs
@@ -42,8 +42,12 @@ impl ManifestSource {
                             .collect(),
                     ),
                     Err(e) => {
-                        tracing::error!("error from manifest stream: {}", e);
-                        Event::UpdateManifest(vec![])
+                        tracing::error!(
+                            "transient error fetching persisted query manifest; \
+                             retaining previous catalog: {}",
+                            e
+                        );
+                        Event::ManifestError(e)
                     }
                 })
                 .boxed(),
--- a/crates/apollo-mcp-server/src/errors.rs
+++ b/crates/apollo-mcp-server/src/errors.rs
@@ -44,6 +44,20 @@ pub enum OperationError {
     Manifest(Box<dyn std::error::Error + Send + Sync + 'static>),
 }
 
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn operation_error_manifest_display() {
+        let err = OperationError::Manifest("connection refused".into());
+        assert_eq!(
+            err.to_string(),
+            "Error fetching Uplink persisted-query manifest: connection refused"
+        );
+    }
+}
+
 /// An error in server initialization
 #[derive(Debug, thiserror::Error)]
 pub enum ServerError {
--- a/crates/apollo-mcp-server/src/errors.rs
+++ b/crates/apollo-mcp-server/src/errors.rs
@@ -5,6 +5,7 @@ use apollo_mcp_registry::platform_api::operation_collections::error::CollectionE
 use reqwest::header::{InvalidHeaderName, InvalidHeaderValue};
 use rmcp::serde_json;
 use tokio::task::JoinError;
+use tower::BoxError;
 use url::ParseError;
 
 /// An error in operation parsing
@@ -39,6 +40,9 @@ pub enum OperationError {
 
     #[error("Error loading collection: {0}")]
     Collection(CollectionError),
+
+    #[error("Error fetching Uplink persisted-query manifest: {0}")]
+    Manifest(BoxError),
 }
 
 /// An error in server initialization
--- a/crates/apollo-mcp-server/src/errors.rs
+++ b/crates/apollo-mcp-server/src/errors.rs
@@ -5,7 +5,6 @@ use apollo_mcp_registry::platform_api::operation_collections::error::CollectionE
 use reqwest::header::{InvalidHeaderName, InvalidHeaderValue};
 use rmcp::serde_json;
 use tokio::task::JoinError;
-use tower::BoxError;
 use url::ParseError;
 
 /// An error in operation parsing
@@ -42,7 +41,7 @@ pub enum OperationError {
     Collection(CollectionError),
 
     #[error("Error fetching Uplink persisted-query manifest: {0}")]
-    Manifest(BoxError),
+    Manifest(Box<dyn std::error::Error + Send + Sync + 'static>),
 }
 
 /// An error in server initialization
--- a/crates/apollo-mcp-server/src/errors.rs
+++ b/crates/apollo-mcp-server/src/errors.rs
@@ -40,7 +40,7 @@ pub enum OperationError {
     #[error("Error loading collection: {0}")]
     Collection(CollectionError),
 
-    #[error("Error fetching Uplink persisted-query manifest: {0}")]
+    #[error("Error fetching persisted-query manifest: {0}")]
     Manifest(Box<dyn std::error::Error + Send + Sync + 'static>),
 }
 
@@ -53,7 +53,7 @@ mod tests {
         let err = OperationError::Manifest("connection refused".into());
         assert_eq!(
             err.to_string(),
-            "Error fetching Uplink persisted-query manifest: connection refused"
+            "Error fetching persisted-query manifest: connection refused"
         );
     }
 }
--- a/crates/apollo-mcp-server/src/server/states.rs
+++ b/crates/apollo-mcp-server/src/server/states.rs
@@ -213,8 +213,8 @@ impl StateMachine {
             },
             ServerEvent::ManifestError(e) => match state {
                 State::Running(running) => {
-                    tracing::error!(
-                        "Transient Uplink manifest fetch error while running, \
+                    tracing::warn!(
+                        "Transient manifest fetch error while running, \
                          keeping existing operations: {e}"
                     );
                     running.into()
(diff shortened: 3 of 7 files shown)


crates/apollo-mcp-server/src/server/states.rs
```diff
@@ -557,6 +567,42 @@ mod tests {
         );
     }
 
+    // A ManifestError (transient Uplink fetch failure) while Running should NOT kill the server
+    // and must NOT clear the existing tool catalog.
+    #[tokio::test]
+    async fn manifest_error_keeps_running_server_alive_and_retains_catalog() {
```
DaleSeo, line 573: This test just checks that the server stays `Running`. A regression that keeps the server `Running` while clearing operations would still pass.

crates/apollo-mcp-server/src/server/states.rs
```diff
@@ -211,6 +211,16 @@ impl StateMachine {
                 }
                 _ => State::Error(ServerError::Operation(OperationError::Collection(e))),
             },
+            ServerEvent::ManifestError(e) => match state {
+                State::Running(running) => {
+                    tracing::error!(
```
DaleSeo: Could we log this as a warning because this branch is the keep-last-good path? The server stays up, continues serving its existing operations, and the Uplink poll loop will recover on the next tick.

```suggestion
                    tracing::warn!(
```

crates/apollo-mcp-registry/src/uplink/persisted_queries/manifest_poller.rs
```diff
@@ -42,8 +42,12 @@ impl ManifestSource {
                             .collect(),
                     ),
                     Err(e) => {
-                        tracing::error!("error from manifest stream: {}", e);
-                        Event::UpdateManifest(vec![])
+                        tracing::error!(
+                            "transient error fetching persisted query manifest; \
+                             retaining previous catalog: {}",
+                            e
+                        );
```
DaleSeo: This log triggers for every manifest fetch error, and the state machine logs the same event again. This might come across as too nosy, since there's no action for users to take. I suggest we remove this log to avoid having two lines of logs for users.

crates/apollo-mcp-server/src/errors.rs
```diff
@@ -39,6 +39,23 @@ pub enum OperationError {
 
     #[error("Error loading collection: {0}")]
     Collection(CollectionError),
+
+    #[error("Error fetching Uplink persisted-query manifest: {0}")]
```
DaleSeo: The error message mentions "Uplink," but the path also includes local manifest sources. This could mislead operators into thinking, “Is Uplink the problem?” and cause them to looking in the wrong place.

```suggestion
    #[error("Error fetching persisted-query manifest: {0}")]
```

DaleSeo (commented):
Thanks for the fix, @Staticsubh. This looks great! I have a few comments.

crates/apollo-mcp-server/src/server/states.rs
Staticsubh: Good catch — updated the test to first load one operation via `OperationsUpdated`, then fire `ManifestError`, then destructure the resulting `Running` state and assert `operations.len() == 1`. A regression that silently wipes the catalog will now fail this test.

crates/apollo-mcp-server/src/server/states.rs
Staticsubh: Agreed — downgraded to `tracing::warn!` and also dropped "Uplink" from the message text for consistency with the `OperationError` fix. The server stays up and the poll loop self-heals on the next tick, so `warn` is the right level here.

crates/apollo-mcp-registry/src/uplink/persisted_queries/manifest_poller.rs
Staticsubh: Removed — the log was redundant with the one the state machine emits on the keep-last-good path. Single log line per event now.

crates/apollo-mcp-server/src/errors.rs
Staticsubh: Fixed — changed to `"Error fetching persisted-query manifest: {0}"` (no "Uplink") so it reads correctly for both local and Uplink manifest sources. Updated the display assertion in the unit test to match.

DaleSeo (approved):

PR #764, opened 2026-06-12
refactor: tidy auth token validation and scope handling

DaleSeo:
On the validation side, we're moving the `Claims` type and the algorithm mapping out of the long `validate` method. Now, the function reads more clearly as a "try each server, resolve key, decode, check claims" loop, and we can unit-test the claim logic directly. We’re also replacing the `VerificationKey` struct with a simple `(Jwk, String)` tuple returned from `KeyResolver::resolve_key`.

On the scope side, we're pulling the scope-sufficiency check into a new `ScopeMode::is_satisfied_by` method. This way, the scope tests will use the actual logic instead of a copied version. We’re also rendering the `scope_mode` header through `ScopeMode::as_str` instead of going through a JSON round-trip.

### DIFF
--- a/crates/apollo-mcp-server/src/auth/networked_key_resolver.rs
+++ b/crates/apollo-mcp-server/src/auth/networked_key_resolver.rs
@@ -2,12 +2,12 @@ use std::str::FromStr;
 use std::time::Duration;
 
 use jsonwebtoken::jwk::KeyAlgorithm;
-use jwks::Jwks;
+use jwks::{Jwk, Jwks};
 use serde::Deserialize;
 use tracing::{error, info, trace, warn};
 use url::Url;
 
-use super::valid_token::{KeyResolver, VerificationKey};
+use super::valid_token::KeyResolver;
 
 /// [`KeyResolver`] that fetches signing keys from the network via OIDC/OAuth
 /// discovery.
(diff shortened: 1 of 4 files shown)


crates/apollo-mcp-server/src/auth/networked_key_resolver.rs
```diff
@@ -187,17 +187,14 @@ impl KeyResolver for NetworkedKeyResolver<'_> {
     /// `discovery_timeout` on the happy path. The JWKS fetch does not fall
     /// back to alternate discovery URLs on failure; real providers advertise
     /// the same `jwks_uri` from every well-known path.
-    async fn resolve_key(&self, server: &Url, key_id: &str) -> Option<VerificationKey> {
+    async fn resolve_key(&self, server: &Url, key_id: &str) -> Option<(Jwk, String)> {
```
jerelmiller, line 190: Naive question, but what does the tuple approach provide thats superior to the `VerificationKey` struct approach? Is it just easier destructuring?

jerelmiller (approved):

crates/apollo-mcp-server/src/auth/networked_key_resolver.rs
DaleSeo: I went with a tuple because `VerificationKey` felt like an incomplete name for a struct containing both the JWK and the issuer. I couldn't come up with a better name that wasn't overly wordy, so I thought I'd just opt for the tuple instead. :)

PR #765, opened 2026-06-12
refactor: consolidate ScopeMode logic into methods

DaleSeo:
This PR moves the scope-sufficiency `match` out of the `oauth_validate` middleware and into a `ScopeMode::is_satisfied_by` method. This change simplifies the middleware to a single condition. More importantly, it allows the scope-validation tests to call the actual method instead of a hand-copied version that could drift from production. It also updates the `scope_mode` header rendering. Previously, it serialized the enum to JSON just to trim the quotes and included an unreachable `"unknown"` fallback. Now, it uses a straightforward and reliable `ScopeMode::as_str`. A small guard test ensures that `as_str` stays in sync with the serde `snake_case` rename.

### DIFF
--- a/crates/apollo-mcp-server/src/auth.rs
+++ b/crates/apollo-mcp-server/src/auth.rs
@@ -47,6 +47,19 @@ pub enum ScopeMode {
     RequireAny,
 }
 
+impl ScopeMode {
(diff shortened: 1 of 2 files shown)


DaleSeo:
Folded into #764

PR #767, opened 2026-06-22
fix: tolerate transient Uplink manifest errors at startup

DaleSeo:
After #762 the server still treated every persisted-query manifest fetch error as fatal during startup, so a brief Uplink blip at boot would crash the process even though the Uplink poll loop would have recovered on the very next tick. This change classifies manifest errors as transient or permanent in the Uplink stream, where the typed error is still available, so transient errors are logged at warn and left for the poll loop to retry while only non-retryable errors such as an invalid API key remain fatal at startup and surface as an error while running. That brings the manifest path in line with how operation collections already behave.

### DIFF
--- a/crates/apollo-mcp-server/src/server/states.rs
+++ b/crates/apollo-mcp-server/src/server/states.rs
@@ -213,11 +213,7 @@ impl StateMachine {
             },
             ServerEvent::ManifestError(e) => match state {
                 State::Running(running) => {
-                    // Transient uplink errors are filtered out upstream in
-                    // create_uplink_stream, so only non-retryable errors reach here.
-                    // Keep the last known good operations and log at error level,
-                    // matching the CollectionError handling above.
-                    tracing::error!(
+                    tracing::warn!(
                         "Manifest error while running, keeping existing operations: {e}"
                     );
                     running.into()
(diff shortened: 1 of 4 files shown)


DaleSeo:
@claude can you hear me?

DaleSeo:
@claude can you hear me now?

crates/apollo-mcp-server/src/server/states.rs
```diff
@@ -213,9 +213,12 @@ impl StateMachine {
             },
             ServerEvent::ManifestError(e) => match state {
                 State::Running(running) => {
-                    tracing::warn!(
-                        "Transient manifest fetch error while running, \
-                         keeping existing operations: {e}"
+                    // Transient uplink errors are filtered out upstream in
+                    // create_uplink_stream, so only non-retryable errors reach here.
+                    // Keep the last known good operations and log at error level,
+                    // matching the CollectionError handling above.
```
gocamille: I remember reading [in this comment on PR 762](https://github.com/apollographql/apollo-mcp-server/pull/762#discussion_r3420427236) that we wanted to make this a `warn!` rather than an `error!` -- the above comment seems to revert it back to `error!` because `ManifestSource::Uplink` errors are filtered via the `create_uplink_stream`, but it looks like a transient error from `ManifestSource::LocalHotReload` that occurs while the server is healthy (in a `Running` state) don't go through that filter and would fire an error. Would we want to either revert to `warn!` or make them conditional (less desirable)?

DaleSeo: Good catch, @gocamille! Changed it to `warn!`.

gocamille (approved):
:shipit: thank you @DaleSeo !

PR #768, opened 2026-06-22
fix: negotiate MCP protocol version in initialize handler

DaleSeo:
With `enable_output_schema` turned on, the server advertised `ProtocolVersion::default()` in its `initialize` response. For rmcp 1.6, this resolves to the latest version, which is `2025-11-25`. Unlike the stdio transport, the streamable_http transport doesn’t apply the version downgrade, so that value reached clients as is. Any client that didn’t support `2025-11-25` was refused with the message, `Server's protocol version is not supported: 2025-11-25`. This change allows the `initialize` handler to negotiate according to the MCP lifecycle spec. It will echo the client's requested version when we implement it, and otherwise respond with our latest supported version, which is `2025-06-18`. The negotiated version no longer relies on `enable_output_schema`, and the `outputSchema` and `structuredContent` fields are still gated separately by the existing `client_supports_output_schema` check.

The underlying transport gap (streamable_http does not negotiate, unlike stdio) is tracked upstream in modelcontextprotocol/rust-sdk#916, and this handler-side fix stays correct whether or not that is addressed.

# Testing

Start the server with `overrides.enable_output_schema: true` and `transport.type: streamable_http`, then send an initialize request offering a newer version than the server implements:

```sh
   curl -sS -i http://127.0.0.1:8000/mcp \
     -H 'Content-Type: application/json' \
     -H 'Accept: application/json, text/event-stream' \
     -d '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2025-11-25","capabilities":{},"clientInfo":{"name":"test","version":"1.0.0"}}}'
```
Expect HTTP 200 and a result whose `protocolVersion` is `2025-06-18` (not `2025-11-25`), confirming the client is no longer refused. Repeat with `"protocolVersion":"2025-03-26"` and confirm the response echoes `2025-03-26`.

### DIFF
--- a/.changeset/fix_protocol_version_negotiation.md
+++ b/.changeset/fix_protocol_version_negotiation.md
@@ -1,0 +1,9 @@
+---
+default: patch
+---
+
+# Fix protocol version negotiation when output schema is enabled
+
+Enabling `overrides.enable_output_schema` previously forced the MCP `initialize` response to advertise protocol version `2025-11-25`. Over the `streamable_http` transport that version was returned to clients verbatim, so any client that did not support `2025-11-25` was refused with `Server's protocol version is not supported: 2025-11-25`. With the flag disabled the server advertised `2025-03-26` and connected normally.
+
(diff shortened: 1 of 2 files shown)


gocamille (approved):
Verified that this works as expected along with a code review. Thank you for this @DaleSeo !

PR #769, opened 2026-06-23
chore: cover 403 insufficient-scope path and tidy auth module

DaleSeo:
The middleware path that rejects a valid token without the required scope didn't have a test, so we're adding an end-to-end one. It sets up a mock authorization server with OIDC discovery and a symmetric JWKS. Also cleans up a `serialize_with` attribute on `discovery_timeout` that was never used since the type is only deserialized.

### DIFF
--- a/crates/apollo-mcp-server/src/auth.rs
+++ b/crates/apollo-mcp-server/src/auth.rs
@@ -656,6 +656,10 @@ mod tests {
     }
 
     mod oauth_validate {
+        use base64::Engine as _;
+        use base64::engine::general_purpose::URL_SAFE_NO_PAD;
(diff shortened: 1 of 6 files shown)


crates/apollo-mcp-server/src/auth.rs
```diff
@@ -710,6 +719,99 @@ mod tests {
             assert!(www_auth.contains("resource_metadata"));
             assert!(!www_auth.contains("scope="));
         }
+
+        async fn valid_token_with_insufficient_scopes_response() -> (StatusCode, String) {
+            let mut server = mockito::Server::new_async().await;
```
gocamille, line 724: 🎉

gocamille (approved):
Thank you for this @DaleSeo this exactly what it needed 🎉

PR #775, opened 2026-06-26
fix: keep collection polling alive

DaleSeo:
Operation collection polling no longer stops permanently when a post-initial poll receives a GraphOS response error, such as an intermittent permission response or missing GraphQL data. Previously, those response errors were classified as non-transient and caused the poll loop to exit, so later edits to the Studio operation collection were not picked up until the server process restarted.

Apollo MCP Server now keeps serving the last known good operation set and retries on the next poll interval after any poll-loop failure. Initial collection load failures still retain their existing behavior: non-transient errors are reported as terminal startup/configuration failures before the server has a collection to serve.

### DIFF
--- a/crates/apollo-mcp-registry/src/platform_api/operation_collections/collection_poller.rs
+++ b/crates/apollo-mcp-registry/src/platform_api/operation_collections/collection_poller.rs
@@ -271,6 +271,7 @@ async fn write_init_response(
         true
     }
 }
+
 impl CollectionSource {
     pub fn into_stream(self) -> Pin<Box<dyn Stream<Item = CollectionEvent> + Send>> {
         match self {
@@ -303,17 +304,16 @@ impl CollectionSource {
                     OperationCollectionResult::NotFoundError(NotFoundError { message })
                     | OperationCollectionResult::PermissionError(PermissionError { message })
                     | OperationCollectionResult::ValidationError(ValidationError { message }) => {
-                        if let Err(e) = sender
-                            .send(CollectionEvent::CollectionError(CollectionError::Response(
-                                message,
-                            )))
-                            .await
-                        {
+                        let err = CollectionError::Response(message);
+                        tracing::error!(
+                            "Failed to fetch operation collection with permanent error: {err}"
+                        );
+                        if let Err(e) = sender.send(CollectionEvent::CollectionError(err)).await {
                             tracing::debug!(
                                 "failed to send error to collection stream. This is likely to be because the server is shutting down: {e}"
                             );
-                            return;
                         }
+                        return;
                     }
                     OperationCollectionResult::OperationCollection(collection) => {
                         let should_poll = write_init_response(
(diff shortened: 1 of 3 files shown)


crates/apollo-mcp-registry/src/platform_api/operation_collections/collection_poller.rs
```diff
@@ -303,17 +304,16 @@ impl CollectionSource {
...
+                        tracing::error!(
+                            "Failed to fetch operation collection with permanent error: {err}"
+                        );
+                        if let Err(e) = sender.send(CollectionEvent::CollectionError(err)).await {
                             tracing::debug!(
                                 "failed to send error to collection stream. This is likely to be because the server is shutting down: {e}"
                             );
-                            return;
                         }
+                        return;
```
DaleSeo, line 316: This remains fatal because it happens before the initial collection load succeeds. The poller only becomes tolerant after it has a last-known-good operation set to keep serving.

crates/apollo-mcp-registry/src/platform_api/operation_collections/collection_poller.rs
```diff
@@ -366,10 +377,7 @@ impl CollectionSource {
                         tracing::debug!("Operation collection unchanged");
                     }
                     Err(err) => {
-                        if !err.is_transient() {
-                            err.send_to_stream(&sender).await;
-                            break;
-                        }
+                        tracing::warn!("Failed to poll operation collection, will retry: {err}");
```
DaleSeo, line 380: Once the initial load has succeeded, poll errors should not stop the collection stream. We keep the last known good operations and retry on the next interval, regardless of whether the error classifier would call this transient.

This is meant to be a warning instead of an error since the initial load was successful. The server continues to serve the last known good collection and will try again during the next poll interval. So, this is more of an operational issue rather than a complete collection failure.

crates/apollo-mcp-registry/src/platform_api/operation_collections/error.rs
```diff
@@ -22,40 +19,22 @@ pub enum CollectionError {
...
             })
-        ) {
-            tracing::warn!(
-                "Failed to fetch operation collection with transient error, will retry: {self}"
-            );
-            true
-        } else {
-            tracing::error!("Failed to fetch operation collection with permanent error: {self}");
-            false
-        }
```
DaleSeo, line 42: I removed the logging from `is_transient`, so now this method just classifies the error. The same error type is treated differently based on the phase. For example, an initial load error might stop the startup, while poll-loop errors should be logged as retryable operational failures. Keeping the logging at the call site makes that phase-specific behavior clear.

crates/apollo-mcp-registry/src/platform_api/operation_collections/error.rs
```diff
@@ -22,40 +19,22 @@ pub enum CollectionError {
...
-    }
-
-    /// Sends the error to the collection stream.
-    pub(super) async fn send_to_stream(self, sender: &Sender<CollectionEvent>) {
```
DaleSeo, line 46: I removed `send_to_stream` from here because sending an error event is poller behavior, not error-type behavior. Keeping it inline in the poller makes the terminal paths easier to read.

gocamille (approved):
Thank you for doing this @DaleSeo !

PR #777, opened 2026-06-30
feat: add per-issuer JWKS cache to eliminate redundant network calls on warm path

gocamille:
## Summary

Each token validation previously triggered two network calls (discovery + JWKS fetch) regardless of whether we'd seen that issuer before. This PR adds a **per-issuer JWKS cache** to short-circuit those calls on the warm path.

- Adds a `jwks_cache: Arc<RwLock<HashMap<Url, CachedJwks>>>` field to `AuthState`, shared across all middleware invocations via an `Arc` clone
- Threads the cache and a configurable `jwks_cache_ttl` (default 10 min) into `NetworkedKeyResolver`
- `resolve_key` now checks the cache before any network call: on a fresh hit with the requested `kid` present, returns immediately; on any miss (no entry, stale entry, or kid not found) falls through to the existing two-step discovery + JWKS fetch, then populates the cache
- Adds `jwks_cache_ttl: Option<Duration>` to `Config` with `humantime_serde` support so operators can tune it via YAML or env

## What's unchanged

Cold-path behaviour is byte-for-byte identical to before where both network calls still run on any miss, in the same order. 

## New tests
- `warm_hit_returns_without_network` tests for a pre-populated fresh cache entry; asserts zero network calls 
- `cold_miss_populates_cache` tests an empty cache; asserts both discovery and JWKS endpoints are hit exactly once and that the entry is present afterward
- `expired_entry_triggers_refetch` tests for a stale entry (1ms TTL + 10ms sleep); asserts network is hit and cache is overwritten with fresh keys

### DIFF
--- a/.changeset/add_per_issuer_jwks_cache.md
+++ b/.changeset/add_per_issuer_jwks_cache.md
@@ -1,0 +1,9 @@
+---
+default: patch
+---
+
+# Add per-issuer JWKS cache to eliminate redundant network calls
+
+Each token validation previously triggered two network calls (discovery + JWKS fetch) regardless of whether the issuer had been seen before. Apollo MCP Server now caches JWKS responses per issuer and reuses them on the warm path, falling back to a full fetch only when the cache is missing, stale, or does not contain the requested key ID.
+
+The cache TTL defaults to 10 minutes and can be tuned via `jwks_cache_ttl` in YAML config or the corresponding environment variable.
--- a/.changeset/add_per_issuer_jwks_cache.md
+++ b/.changeset/add_per_issuer_jwks_cache.md
@@ -1,9 +1,9 @@
 ---
-default: patch
+default: minor
 ---
 
-# Add per-issuer JWKS cache to eliminate redundant network calls
+# Add per-issuer JWKS cache
 
-Each token validation previously triggered two network calls (discovery + JWKS fetch) regardless of whether the issuer had been seen before. Apollo MCP Server now caches JWKS responses per issuer and reuses them on the warm path, falling back to a full fetch only when the cache is missing, stale, or does not contain the requested key ID.
+Apollo MCP Server now manages JWKS as a cached, per-issuer resource instead of refetching keys for every token validation. Previously, each request triggered two network calls (OIDC discovery and JWKS fetch) regardless of whether the issuer had been seen before. The new cache reuses JWKS responses on the warm path and falls back to a full fetch only when the cache is missing, stale, or does not contain the requested key ID.
 
-The cache TTL defaults to 10 minutes and can be tuned via `jwks_cache_ttl` in YAML config or the corresponding environment variable.
+The cache TTL defaults to 10 minutes and can be tuned via `transport.auth.jwks_cache_ttl` in YAML config or the corresponding environment variable.
--- a/.changeset/add_per_issuer_jwks_cache.md
+++ b/.changeset/add_per_issuer_jwks_cache.md
@@ -1,9 +1,0 @@
----
-default: minor
----
-
-# Add per-issuer JWKS cache
-
-Apollo MCP Server now manages JWKS as a cached, per-issuer resource instead of refetching keys for every token validation. Previously, each request triggered two network calls (OIDC discovery and JWKS fetch) regardless of whether the issuer had been seen before. The new cache reuses JWKS responses on the warm path and falls back to a full fetch only when the cache is missing, stale, or does not contain the requested key ID.
-
-The cache TTL defaults to 10 minutes and can be tuned via `transport.auth.jwks_cache_ttl` in YAML config or the corresponding environment variable.
--- a/.changeset/jwks_cache_singleflight_and_refresh_rate_limit.md
+++ b/.changeset/jwks_cache_singleflight_and_refresh_rate_limit.md
@@ -10,6 +10,6 @@ The cache TTL is 10 minutes. Stale entries continue to serve known key IDs while
 
 When multiple requests arrive simultaneously for an issuer whose JWKS is not yet cached (or whose entry has just expired), Apollo MCP Server now coalesces them into a single outbound discovery + JWKS fetch instead of fanning out one fetch per request. Followers await the leader's result and then read the populated cache; one upstream round-trip serves all concurrent callers.
 
-Apollo MCP Server now allows at most one JWKS refresh per issuer per internal window (default 60 seconds). When a token arrives with a `kid` that is absent from both the fresh and stale cached JWKS and the window is exhausted, the request is rejected with 401 immediately — no outbound HTTP is made. Concurrent misses within an allowed window still coalesce into a single upstream fetch.
+Apollo MCP Server now allows at most one JWKS refresh per issuer per internal window (default 60 seconds). When a token arrives with a `kid` that is absent from both the fresh and stale cached JWKS and the window is exhausted, the request is rejected with 401 immediately — no outbound HTTP request is made. Concurrent misses within an allowed window still coalesce into a single upstream fetch.
 
 This closes the request-amplification vector where an attacker rotating unique bogus `kid`s could drive one upstream discovery + JWKS fetch pair per request.
--- a/.changeset/jwks_cache_singleflight_and_refresh_rate_limit.md
+++ b/.changeset/jwks_cache_singleflight_and_refresh_rate_limit.md
@@ -1,0 +1,15 @@
+---
+default: minor
+---
+
+# Per-issuer JWKS cache with singleflight fetches and refresh rate limiting
+
+Apollo MCP Server now manages JWKS as a cached, per-issuer resource instead of refetching keys for every token validation. Previously, each request triggered two network calls (OIDC discovery and JWKS fetch) regardless of whether the issuer had been seen before. The new cache reuses JWKS responses on the warm path and triggers a refresh only when the entry is missing, past its TTL, or does not contain the requested key ID.
+
+The cache TTL is 10 minutes. Stale entries continue to serve known key IDs while a refresh is in progress or when a refresh fails — keys are only evicted when a successful refresh no longer includes them.
+
+When multiple requests arrive simultaneously for an issuer whose JWKS is not yet cached (or whose entry has just expired), Apollo MCP Server now coalesces them into a single outbound discovery + JWKS fetch instead of fanning out one fetch per request. Followers await the leader's result and then read the populated cache; one upstream round-trip serves all concurrent callers.
+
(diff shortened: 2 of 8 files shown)


crates/apollo-mcp-server/src/auth/networked_key_resolver.rs
```diff
@@ -1,5 +1,7 @@
+use std::collections::HashMap;
 use std::str::FromStr;
-use std::time::Duration;
+use std::sync::{Arc, RwLock};
```
gocamille: I noticed the use of the `tokio` `RwLock` as convention: here all network I/O happens between dropping the read lock and acquiring the write lock. No lock is ever held across an `await` point. So here `std::sync::RwLock` is the right tool 👍

crates/apollo-mcp-server/src/auth/networked_key_resolver.rs
```diff
@@ -182,8 +205,40 @@ impl KeyResolver for NetworkedKeyResolver<'_> {
...
+        {
+            let mut cache = self.jwks_cache.write().unwrap_or_else(|e| e.into_inner());
+            if let Some(entry) = cache.get(server)
+                && entry.is_fresh(self.ttl)
+                && let Some(jwk) = entry.keys.keys.get(key_id)
+            {
+                return Some((jwk.clone(), entry.issuer.clone()));
+            }
+            cache.insert(
+                server.clone(),
```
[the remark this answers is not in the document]
gocamille: Good catch, Claude!

.changeset/add_per_issuer_jwks_cache.md
```diff
@@ -0,0 +1,9 @@
+---
+default: patch
```
DaleSeo: Shouldn't we consider this a new feature?

crates/apollo-mcp-server/src/auth.rs
```diff
@@ -229,6 +229,13 @@ pub struct Config {
     #[serde(default, deserialize_with = "deserialize_header_map")]
     #[schemars(with = "HashMap<String, String>")]
     pub discovery_headers: HeaderMap,
+
+    /// How long a cached JWKS entry stays fresh before a request forces
+    /// a re-fetch. Stale entries behave identically to a cache miss.
+    /// Defaults to 10 minutes.
+    #[serde(deserialize_with = "humantime_serde::deserialize", default)]
+    #[schemars(with = "Option<String>")]
+    pub jwks_cache_ttl: Option<Duration>,
```
DaleSeo: Can we document this new config option?

.changeset/add_per_issuer_jwks_cache.md
gocamille: I agree @DaleSeo as this introduces tablestakes JWKS key set lifecycle management, which is pretty foundational machinery for treating JWKS as a managed resource rather than a raw HTTP fetch as we were doing before. I will update this 🙏

crates/apollo-mcp-server/src/auth.rs
gocamille: Yes, the docs appear in this commit 3164dfd332e81ee37b4663900aa37a12f5643e61

crates/apollo-mcp-server/src/auth/networked_key_resolver.rs
```diff
@@ -182,13 +218,37 @@ impl KeyResolver for NetworkedKeyResolver<'_> {
...
+        // Return immediately if the key is cached and fresh.
+        if let Ok(cache) = self.jwks_cache.read()
+            && let Some(entry) = cache.get(server)
+            && entry.is_fresh(self.ttl)
+            && let Some(result) = entry.lookup(key_id, server)
+        {
+            return Some(result);
+        }
+
         let metadata = discover_metadata(self.client, server, self.discovery_timeout).await?;
```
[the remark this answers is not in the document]
gocamille: @claude This will be coming in a followup that is already anticipated 👍

DaleSeo (approved):
Thanks!

.changeset/jwks_cache_singleflight_and_refresh_rate_limit.md
```diff
@@ -0,0 +1,15 @@
+---
+default: minor
+---
+
+# Per-issuer JWKS cache with singleflight fetches and refresh rate limiting
+
+Apollo MCP Server now manages JWKS as a cached, per-issuer resource instead of refetching keys for every token validation. Previously, each request triggered two network calls (OIDC discovery and JWKS fetch) regardless of whether the issuer had been seen before. The new cache reuses JWKS responses on the warm path and triggers a refresh only when the entry is missing, past its TTL, or does not contain the requested key ID.
```
pragl:
```suggestion
Apollo MCP Server now manages JSON Web Key Set (JWKS) as a cached, per-issuer resource instead of refetching keys for every token validation. Previously, each request triggered two network calls—OpenID Connect (OIDC) discovery and JWKS fetch—regardless of whether the issuer had been seen before. The new cache reuses JWKS responses on the warm path and triggers a refresh only when the entry is missing, past its TTL, or does not contain the requested key ID.
```

.changeset/jwks_cache_singleflight_and_refresh_rate_limit.md
```diff
@@ -0,0 +1,15 @@
...
+
+The cache TTL is 10 minutes. Stale entries continue to serve known key IDs while a refresh is in progress or when a refresh fails — keys are only evicted when a successful refresh no longer includes them.
```
pragl:
```suggestion
The cache TTL is 10 minutes. Stale entries continue to serve known key IDs while a refresh is in progress or when a refresh fails—keys are only evicted when a successful refresh no longer includes them.
```

.changeset/jwks_cache_singleflight_and_refresh_rate_limit.md
```diff
@@ -0,0 +1,15 @@
...
+
+When multiple requests arrive simultaneously for an issuer whose JWKS is not yet cached (or whose entry has just expired), Apollo MCP Server now coalesces them into a single outbound discovery + JWKS fetch instead of fanning out one fetch per request. Followers await the leader's result and then read the populated cache; one upstream round-trip serves all concurrent callers.
+
+Apollo MCP Server now allows at most one JWKS refresh per issuer per internal window (default 60 seconds). When a token arrives with a `kid` that is absent from both the fresh and stale cached JWKS and the window is exhausted, the request is rejected with 401 immediately — no outbound HTTP is made. Concurrent misses within an allowed window still coalesce into a single upstream fetch.
```
pragl:
```suggestion
Apollo MCP Server now allows at most one JWKS refresh per issuer per internal window (default 60 seconds). When a token arrives with a `kid` that is absent from both the fresh and stale cached JWKS and the window is exhausted, the request is rejected with 401 immediately — no outbound HTTP request is made. Concurrent misses within an allowed window still coalesce into a single upstream fetch.
```

docs/source/auth.mdx
```diff
@@ -224,6 +224,23 @@ transport:
 - Increase the timeout if your OAuth server is on a slow network or responds slowly.
 - Decrease the timeout in high-performance environments where fast failure is preferred.
 
+### JWKS cache
+
+Token validation requires Apollo MCP Server to fetch each authorization server's JSON Web Key Set (JWKS) — preceded by an OIDC/OAuth discovery call to locate the `jwks_uri`. Without caching, every authorized request pays the cost of both round trips.
```
pragl:
```suggestion
Token validation requires Apollo MCP Server to fetch each authorization server's JSON Web Key Set (JWKS)—preceded by an OIDC/OAuth discovery call to locate the `jwks_uri`. Without caching, every authorized request pays the cost of both round trips.
```

docs/source/auth.mdx
```diff
@@ -224,6 +224,23 @@ transport:
...
+
+Apollo MCP Server caches the JWKS per issuer in memory and reuses it on the warm path. Cached entries stay fresh for 10 minutes by default before the next request triggers a re-fetch.
```
pragl:
```suggestion
Apollo MCP Server caches the JWKS per issuer in memory and reuses it on the warm path. Cached entries stay fresh for 10 minutes before the next request triggers a re-fetch.
```

docs/source/auth.mdx
```diff
@@ -224,6 +224,23 @@ transport:
...
+
+**Behavior:**
+
+- On a fresh hit where the requested `kid` is present, the cached JWKS is used and no network calls are made.
+- When there is no cached entry, the cached entry is stale, or the requested `kid` is not present in the cached entry, the next request may trigger a discovery + JWKS fetch and repopulate the cache — subject to the per-issuer refresh rate limit below. Newly rotated keys are picked up the next time a refresh fires for that issuer.
```
pragl:
```suggestion
- When there is no cached entry, when the cached entry is stale, or when the requested `kid` is not present in the cached entry, the next request might trigger a discovery plus JWKS fetch and repopulate the cache, subject to a per-issuer refresh rate limit. Newly rotated keys are picked up the next time a refresh fires for that issuer.
```

docs/source/auth.mdx
```diff
@@ -224,6 +224,23 @@ transport:
...
+- Stale entries are still served for known `kid`s. If a refresh is rate-limited or fails — for example, during an identity provider outage — a `kid` present in the cached JWKS is still accepted. Only requests presenting a `kid` that is absent from the cache are rejected. Stale entries persist until the next successful refresh, so there is no upper bound on how long a known `kid` remains accepted while refreshes keep failing.
```
pragl:
```suggestion
- Stale entries are still served for known `kid`s. If a refresh is rate-limited or fails—for example, during an identity-provider outage—a `kid` present in the cached JWKS is still accepted. Only requests presenting a `kid` that is absent from the cache are rejected. Stale entries persist until the next successful refresh, so there is no upper bound on how long a known `kid` remains accepted while refreshes keep failing.
```

docs/source/auth.mdx
```diff
@@ -224,6 +224,23 @@ transport:
...
+- The cache is process-local. Each replica warms its own cache independently after startup.
+
+**Considerations:**
+
+- When a request presents a `kid` that is not in the cached JWKS, Apollo MCP Server triggers a re-fetch — but to prevent fabricated `kid` values from driving one upstream round-trip per request, at most one re-fetch per issuer is allowed within a short internal window. Concurrent requests during an active re-fetch share that single upstream call. If the window is exhausted, the request is rejected with `401 Unauthorized` and the client can retry.
```
pragl:
```suggestion
- When a request presents a `kid` that is not in the cached JWKS, Apollo MCP Server triggers a re-fetch, but to prevent fabricated `kid` values from driving one upstream round-trip per request, up to one re-fetch per issuer is allowed within a short internal window. Concurrent requests during an active re-fetch share that single upstream call. If the window is exhausted, the request is rejected with `401 Unauthorized`, and the client can retry.
```

pragl (approved):
Largely minor style guide recommendations. None blocking.

I liked Claude's recommendation about dropping the "by default" language in the bit about cached entries staying fresh for 10 minutes, but please review to make sure the logic is sound.

PR #778, opened 2026-06-30
feat: configure Rhai scripts directory

DaleSeo:
This lets deployments configure where Apollo MCP Server loads Rhai scripts from instead of always looking for `rhai/` under the process working directory. The default remains `rhai`, so existing setups keep working, while containerized deployments can mount scripts somewhere like `/config/rhai` and point the server there with `rhai.scripts`. The configured scripts directory is used both for startup loading and for hot reload watching.

### DIFF
--- a/crates/apollo-mcp-server/src/runtime/rhai.rs
+++ b/crates/apollo-mcp-server/src/runtime/rhai.rs
@@ -1,0 +1,42 @@
+use std::path::PathBuf;
+
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Clone, Deserialize, Serialize, JsonSchema)]
+#[serde(deny_unknown_fields)]
+#[serde(default)]
+pub struct RhaiConfig {
+    /// Directory containing Rhai scripts. The server loads `main.rhai` from this directory.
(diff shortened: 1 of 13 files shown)


crates/apollo-mcp-server/src/runtime/rhai.rs
```diff
@@ -0,0 +1,41 @@
...
+
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Clone, Deserialize, Serialize, JsonSchema)]
+#[serde(deny_unknown_fields)]
+#[serde(default)]
+pub struct RhaiConfig {
+    /// Directory containing Rhai scripts. The server loads `main.rhai` from this directory.
+    pub scripts: PathBuf,
```
gocamille: Would the name `scripts_dir` or `directory` be clearer here?

gocamille (approved):

crates/apollo-mcp-server/src/runtime/rhai.rs
DaleSeo: Makes sense! Reamed it to `scripts_dir`.

Issue #779, opened 2026-07-01
Support OR-of-AND per-operation required scopes


Staticsubh:
## Problem

Apollo Router supports Boolean scope requirements through `@requiresScopes`, where the inner array is AND and outer arrays are OR:

```graphql
@requiresScopes(scopes: [["scope1", "scope2"], ["admin"]])
```

That authorizes either:

- `scope1` AND `scope2`
- OR `admin`

Apollo MCP Server currently supports global `scope_mode: require_all|require_any`, but per-operation `overrides.required_scopes` is modeled as an operation-to-flat-scope-list map. That makes operation-specific alternatives difficult to express without weakening the global MCP boundary policy.

This matters for deployments where the same MCP tool should be callable by alternative valid authority paths, for example user-token scopes OR system-token scopes, admin override scopes OR domain-specific scopes, or Router-aligned scope policy migration into MCP tools.

## Desired config syntax

Support Router-aligned nested scope requirements for `overrides.required_scopes`:

```yaml
overrides:
  required_scopes:
    GetPatient:
      - ["user/Patient.rs"]
      - ["system/Patient.rs"]
    UpdatePatient:
      - ["user/Patient.write", "tenant:clinician"]
      - ["system/Patient.write"]
```

Semantics:

- Inner list: all scopes are required.
- Outer list: any inner list may satisfy the operation.

Current flat syntax should continue to work:

```yaml
overrides:
  required_scopes:
    DeleteUser:
      - user:write
      - admin
```

Backward-compatible interpretation:

```yaml
DeleteUser:
  - user:write
  - admin
```

means:

```yaml
DeleteUser:
  - ["user:write", "admin"]
```

## Backward compatibility

This should be additive:

- Existing `Map<String, List<String>>` configs remain valid.
- Existing behavior remains AND across all listed scopes.
- New nested configs enable OR-of-AND for only the operations that opt in.
- Existing global `scope_mode` behavior remains unchanged and applies only to `transport.auth.scopes`.

## Migration risk

Risk is low if deserialization accepts both legacy flat lists and new nested lists.

Main risks to handle explicitly:

- YAML ambiguity between a flat list and nested list.
- Avoid silently treating malformed nested values as permissive.
- Preserve current behavior for deployments using `required_scopes` as a strict AND gate.
- Ensure `scope_mode: require_any` remains global-only and is not confused with per-operation OR alternatives.
- Ensure docs distinguish execution authorization from `tools/list` discovery visibility.
[...]

PR #780, opened 2026-07-01
feat(auth): support per-operation scope alternatives

Staticsubh:
Refs #779.

## Summary

Adds support for per-operation scope alternatives in `overrides.required_scopes`.

- Preserves the existing flat list syntax as "all listed scopes are required".
- Adds nested list syntax for alternatives, matching Apollo Router-style OR-of-AND semantics.
- Introduces `OperationRequiredScopes` for parsing, evaluating, and rendering per-operation scope requirements.
- Rejects empty nested scope alternatives, at both the parser and the generated JSON schema, so `- []` cannot accidentally make a protected operation public.
- Updates auth/config/tool-definition docs with the new syntax and challenge-header behavior.

## Behavior

Existing configs continue to work unchanged:

```yaml
overrides:
  required_scopes:
    DeleteUser:
      - user:write
      - admin
```

The flat form still requires every listed scope.

New nested configs allow alternatives:

```yaml
overrides:
  required_scopes:
    UpdateUser:
      - [user:write, tenant:admin]
      - [admin]
```

This requires either `user:write` plus `tenant:admin`, or `admin`.

This change only affects per-operation `tools/call` authorization. It does not change global auth scope behavior, stdio behavior, unauthenticated/no-auth behavior, or `tools/list` visibility. If auth is not configured, `required_scopes` remains ignored as before.

For insufficient scopes, the server still returns `403` with `WWW-Authenticate: Bearer error="insufficient_scope"`. Because the OAuth `scope` auth-param is space-delimited and cannot represent grouped OR conditions, and per [RFC 6750 Section 3](https://datatracker.ietf.org/doc/html/rfc6750#section-3) `scope` describes what the resource requires rather than what the caller's token is missing, nested alternatives always report the complete first-listed alternative — this doesn't vary by presented token, and gives operators control over which scope set gets advertised. Since every scope in the advertised group is required, the per-operation challenge also explicitly reports `scope_mode="require_all"`, regardless of the globally configured `scope_mode`.

## Maintainer Review Note

The internal/public Rust type for per-operation required scopes changes from `HashMap<String, Vec<String>>` to `HashMap<String, OperationRequiredScopes>`. YAML compatibility is preserved for existing configs, but embedded Rust callers that construct the server directly may need to update their call sites.

## Local Testing

Tested locally against the pinned `1.92.0` toolchain (`AWS_LC_SYS_CMAKE_BUILDER=1` for `aws-lc-sys`):

```sh
cargo fmt --check
cargo test -p apollo-mcp-server --lib
cargo test -p apollo-mcp-server --bin apollo-mcp-server
cargo clippy -p apollo-mcp-server --all-targets -- -D warnings
```

Results:

- `cargo fmt --check` passed.
- `cargo test -p apollo-mcp-server --lib` passed: 631 passed, 0 failed related to this change (6 pre-existing, unrelated failures — one network test and five `insta` snapshot-drift tests in `introspection`/`schema_walker` — reproduce identically on a clean checkout of `main`).
- `cargo test -p apollo-mcp-server --bin apollo-mcp-server` passed: 86 passed, 0 failed.
- `cargo clippy -p apollo-mcp-server --all-targets -- -D warnings` passed clean.


Staticsubh:
Addressed the current review feedback in commit `9df6bd0`:

- tightened docs wording for Apollo style review: Apollo MCP Server branding, active voice, code-formatted `403 Forbidden`, and descriptive link text;
- added a changeset for the scope-alternatives feature;
- preserved flat-map builder compatibility via `OperationScopeRequirements`, so existing `HashMap<String, Vec<String>>` builder inputs still map to flat AND semantics;
- updated generated config-schema wording in `Overrides`;
- fixed the integration-test builder call that CI flagged by typing the empty legacy flat-scope map.

Local validation passed in Docker (`rust:1.92.0`):

```sh
cargo fmt --check
cargo test -p apollo-mcp-server --lib
cargo test -p apollo-mcp-server --bin apollo-mcp-server
cargo clippy -p apollo-mcp-server --lib --bin apollo-mcp-server -- -D warnings
CARGO_BUILD_JOBS=1 cargo test -p apollo-mcp-server --test collection_sync --no-run
cargo clippy -p apollo-mcp-server --all-targets -- -D warnings
```

gocamille:
Hi @Staticsubh, thank you for the draft, and sorry for the delay getting back to you. We've been focused on a broader tool update design for the server, and I wanted to check how this intersects with that before commenting!

This looks like it cleanly resolves #779. I also noticed you cross-referenced #781 (scope-first tool authorization mappings) from this PR and just want to confirm: is that meant to be a separate, related ask, or did you intend #780 to also cover that shape of the problem? We're scoping #781 too, and there's some conceptual overlap between "which scopes gate an operation" and "which toolset a caller sees," so I want to make sure I'm reviewing this against the right scope.

That overlap may or may not change the shape of what lands — I don't want to commit to a merge path until I've looked at both side by side. Will have concrete feedback for you shortly!

Staticsubh:
Hi @gocamille,

Good question, and that boundary is intentional. #780 is scoped strictly to #779 — the boolean OR-of-AND logic for how many scopes gate a single operation. #781 is a separate, later concern about authoring ergonomics (writing `scope -> [tools]` once instead of repeating the same scope list per operation) — it's about reducing config repetition, not about boolean satisfaction logic.

#780 doesn't touch or presuppose anything about #781's `scope_tool_mappings` shape. If/when #781 lands, it would most likely expand into `overrides.required_scopes` at config-load time (as noted in #781's own "Desired semantics" section), so it should compose cleanly with the `OperationRequiredScopes` enum introduced here rather than compete with it. Happy to confirm that compatibility explicitly once you've scoped #781 — but there's no need to block #780 on that design; they solve different problems.

crates/apollo-mcp-server/src/scope_requirements.rs
```diff
@@ -0,0 +1,186 @@
...
+///
+/// A nested list mirrors Apollo Router's `@requiresScopes` semantics: each
+/// inner list is an AND group, and the outer list is OR.
+///
+/// ```yaml
+/// GetUser:
+///   - [user:read]
+///   - [admin]
+/// ```
+#[derive(Clone, Debug, JsonSchema, PartialEq, Eq)]
```
gocamille: The derived `JsonSchema` here is externally tagged (`{"All": [...]}` / `{"AnyOf": [[...]]}`), but the manual `Deserialize` below is untagged. The `schemars` derive reads serde attributes on this type (there are none) but can’t see into the manual implementation. So the released `config.schema.json` (generated by the `config-schema` binary in `release-bins.yml`) will now mark every valid config invalid and vice versa.

An update here could be to derive `JsonSchema` on the untagged helper enum and forward this type’s implementation to it so the schema and parser can’t drift:

```
#[derive(Deserialize, JsonSchema)]
#[serde(untagged)]
enum OperationRequiredScopesDefinition { ... }

impl JsonSchema for OperationRequiredScopes {
    fn schema_name() -> Cow<'static, str> { "OperationRequiredScopes".into() }
    fn json_schema(gen: &mut SchemaGenerator) -> Schema {
        OperationRequiredScopesDefinition::json_schema(gen)
    }
}
```

We could also optionally add `#[schemars(length(min = 1))]` on the inner groups to mirror the empty-group validation. It would also be great here to add a test validating the documented YAML examples against the generated schema so this can’t regress.

crates/apollo-mcp-server/src/scope_requirements.rs
```diff
@@ -0,0 +1,186 @@
...
+    AnyOf(Vec<Vec<String>>),
+}
+
+impl<'de> Deserialize<'de> for OperationRequiredScopes {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        match OperationRequiredScopesDefinition::deserialize(deserializer)? {
+            OperationRequiredScopesDefinition::All(required) => {
```
gocamille: This rejects an empty inner group, but an empty value still slips through:
```
overrides:
  required_scopes:
    DeleteUser: []
```
`Serde` tries the `All` variant first, and an empty YAML list matches it. So this parses as `All(vec![])`. An empty requirement list is satisfied by every token. An empty value should be a config error, not quiet full access (this is the same reasoning that led to the explicit `allow_any_audience` flag instead of treating an empty audience list as “skip validation”). The single-type redesign suggested on the enum definition handles this in one place. Minimally, we could reject empty top-level values here too.

crates/apollo-mcp-server/src/scope_requirements.rs
```diff
@@ -0,0 +1,186 @@
...
+/// A nested list mirrors Apollo Router's `@requiresScopes` semantics: each
+/// inner list is an AND group, and the outer list is OR.
+///
+/// ```yaml
+/// GetUser:
+///   - [user:read]
+///   - [admin]
+/// ```
+#[derive(Clone, Debug, JsonSchema, PartialEq, Eq)]
+pub enum OperationRequiredScopes {
```
gocamille: The “no empty groups” rule only exists in the deserializer. Both variants are `pub`, so code that builds this value directly skips the check. That allows two values that would be incorrect: `AnyOf(vec![vec![]])` contains an empty group, every token satisfies it, and the operation is left unprotected. `AnyOf(vec![])` is the opposite: no token satisfies it, and the `403` header comes out as `scope=""` (see my comment on `challenge_scopes`).

It seems as well that a flat list means the same thing as a nested list with one group: `["a","b"]` is equivalent to `[["a","b"]]`. So consider storing everything in that one form, and checking the rules when the value is built:
```
/// Always at least one group; every group has at least one scope.
pub struct OperationRequiredScopes(Vec<Vec<String>>);  // field private

impl OperationRequiredScopes {
    pub fn new(groups: Vec<Vec<String>>) -> Result<Self, InvalidScopeRequirement> { ... }
}
```
With that shape it’s impossible to build a value that skips the checks. The `DeleteUser: []` case from my other comment on the `Deserialize` implementation gets caught in the same place. The existing `From<HashMap<String, Vec<String>>>` for builder callers still works.

crates/apollo-mcp-server/src/scope_requirements.rs
```diff
@@ -0,0 +1,186 @@
...
+            }
+        }
+    }
+}
+
+impl OperationRequiredScopes {
+    /// Returns true when the present token scopes satisfy this requirement.
+    pub fn is_satisfied_by(&self, present: &[String]) -> bool {
+        match self {
+            OperationRequiredScopes::All(required) => {
```
gocamille: Small observation that supports the single-type suggestion on the enum definition: this `All` arm is the one-group case of the `AnyOf` arm below. The same is true in `challenge_scopes`. If the type stores one normalized form, both match statements collapse to a single branch.

crates/apollo-mcp-server/src/scope_requirements.rs
```diff
@@ -0,0 +1,186 @@
...
+    ///
+    /// The OAuth bearer `scope` auth-param is a space-delimited list and cannot
+    /// represent grouped OR conditions. For alternatives, return the complete
+    /// group that requires the fewest additional scopes for this token.
+    pub fn challenge_scopes(&self, present: &[String]) -> Vec<String> {
+        match self {
+            OperationRequiredScopes::All(required) => required.clone(),
+            OperationRequiredScopes::AnyOf(alternatives) => alternatives
+                .iter()
+                .min_by_key(|required| missing_scope_count(required, present))
```
gocamille: I noticed two things here. First: when two alternatives are tied for fewest missing scopes, `min_by_key` picks the first one listed. That’s stable behaviour operators could rely on and worth one sentence in `auth.mdx`: “fewest missing scopes; ties go to the first listed alternative.”

Second: the `.unwrap_or_default()` below only fires if the outer list is empty, and in that case the `403` header comes out as `scope=""`. YAML can’t produce that value, but the public Rust API can (see my comment on the enum definition). Under the single-type suggestion this branch becomes unreachable and can be dropped.

crates/apollo-mcp-server/src/scope_requirements.rs
```diff
@@ -0,0 +1,186 @@
...
+}
+
+impl From<HashMap<String, OperationRequiredScopes>> for OperationScopeRequirements {
+    fn from(required_scopes: HashMap<String, OperationRequiredScopes>) -> Self {
+        Self(required_scopes)
+    }
+}
+
+#[derive(Deserialize)]
+#[serde(untagged)]
```
gocamille, line 126: A mixed entry (`admin` alongside `[a, b]`) or a non-string value is rejected at startup, which is good. But the error the user sees is `data did not match any variant of untagged enum OperationRequiredScopesDefinition`. There’s already a custom error for empty groups. Worth adding one on the helper call in the Deserialize implementation, such as `required_scopes entries must be a list of scopes, or a list of scope lists`

.changeset/support_operation_scope_alternatives.md
```diff
@@ -0,0 +1,9 @@
+---
+default: patch
```
gocamille: This adds new user-facing config syntax and changes the public builder signature (`required_scopes` now takes `impl Into<OperationScopeRequirements>`). That should be minor. The builder change is also worth a sentence here for anyone embedding the server as a library.

crates/apollo-mcp-server/src/auth.rs
```diff
@@ -1650,6 +1649,43 @@ emyPxgcYxn/eR44/KJ4EBs+lVDR3veyJm+kXQ99b21/+jh5Xos1AnX5iItreGCc=
...
+                "tenant:admin".to_string(),
+                "other:scope".to_string(),
+            ];
+            let required = alternative_required();
+            let result = missing_scopes_for_operation(&peek, &required, &scopes);
+            assert!(result.is_none());
+        }
+
+        #[test]
+        fn returns_required_scopes_when_token_satisfies_no_complete_alternative() {
```
gocamille, line 1861: Two cases from #779’s acceptance list are missing here. First: a token whose scopes are spread across several groups without completing any of them, which must fail. This test’s token only touches one group. Second: a token that satisfies more than one group at once, which must pass.

crates/apollo-mcp-server/src/auth.rs
```diff
@@ -600,14 +584,15 @@ async fn oauth_validate(
...
+        let challenge_scopes = required.challenge_scopes(&valid_token.scopes);
         tracing::warn!(
             required = ?required,
             present = ?valid_token.scopes,
             "Token has insufficient scopes for operation"
         );
         tracing::Span::current().record("reason", "insufficient_scope");
         tracing::Span::current().record("status_code", StatusCode::FORBIDDEN.as_u16());
-        return Err(forbidden_error(required));
+        return Err(forbidden_error(&challenge_scopes));
```
gocamille: `challenge_scopes` has unit tests, but no test checks the `WWW-Authenticate` header a client actually receives when a nested requirement fails. The assertion would be that `scope="..."` contains exactly the chosen alternative. Since this PR changes what goes in the header, a test on the full `403` response would close the gap for both.

The infrastructure for this already exists in this file: `valid_token_with_insufficient_scopes_response()` (in the oauth_validate test module) mocks discovery and JWKS, mints a valid JWT with chosen scopes, runs the middleware with oneshot, and returns the status and raw header.

gocamille:
Thank you again @Staticsubh for this contribution! Appreciate your note on the scoping, and we won’t block #780 on #781. One thing to settle before anyone builds on the composability point: you note #781 would compose with the `OperationRequiredScopes` enum introduced here, but I’ve suggested a single-type redesign (see comment on the enum definition) that would change that shape. If we take it, that’s the type #781’s config-load expansion should target. 

Also: since what goes in the `403` `WWW-Authenticate` header when alternatives are configured is a behaviour clients will build against, we're validating the design decision internally and will follow up in this PR shortly!

crates/apollo-mcp-server/src/scope_requirements.rs
Staticsubh: Fixed — `JsonSchema` is now implemented by hand for `OperationRequiredScopes`, forwarding to the untagged `OperationRequiredScopesDefinition` helper (which now derives `JsonSchema` directly, since it's the type that actually carries `#[serde(untagged)]`), so schema and parser can't drift apart again. Also added `#[schemars(length(min = 1))]` on both variants' fields, and a test validating the documented flat/nested YAML examples against the generated schema (`generated_schema_accepts_documented_yaml_examples`).

One caveat I left as a comment on that test: the `length` constraint only reaches the outer list, not each inner group, so `[[]]` still passes schema validation even though the parser correctly rejects it. Closing that fully would need a dedicated newtype for a scope group — felt like more surface than this warranted given `OperationRequiredScopes::new` is the real enforcement either way, but happy to add it if you'd rather have it. See 6a753ff.

crates/apollo-mcp-server/src/scope_requirements.rs
Staticsubh: Fixed as part of the single-type redesign below — `DeleteUser: []` now goes through the same `OperationRequiredScopes::new` validation as everything else and is rejected (`flat_empty_list_is_rejected_not_treated_as_no_requirement`). See 6a753ff.

crates/apollo-mcp-server/src/scope_requirements.rs
Staticsubh: Agreed, and done — collapsed to `OperationRequiredScopes(Vec<Vec<String>>)` with a private field and a validating `new()` that rejects both zero groups and any empty group, so neither of the two bad states you called out (`AnyOf(vec![vec![]])`, `AnyOf(vec![])`) can be constructed anymore, from YAML or from Rust directly.

Kept `From<HashMap<String, Vec<String>>>` for builder callers as you suggested — it now filters out empty scope lists rather than storing a vacuous requirement, which felt more correct than panicking for what's effectively "no override" input. See 6a753ff.

crates/apollo-mcp-server/src/scope_requirements.rs
Staticsubh: Yep — both collapsed to a single branch each under the redesign. See 6a753ff.

crates/apollo-mcp-server/src/scope_requirements.rs
Staticsubh: Documented the tie-breaking rule in auth.mdx ("ties go to the first listed alternative"), and the `unwrap_or_default()` fallback is gone — `challenge_scopes` now does a plain `.expect(...)` justified by `new()` guaranteeing at least one group, with a `#[allow(clippy::expect_used)]` and a comment explaining why. See 6a753ff.

crates/apollo-mcp-server/src/scope_requirements.rs
Staticsubh: Added a custom message for this case too: "required_scopes entries must be a list of scopes, or a list of scope lists", tested in `malformed_entry_reports_a_clear_error`. See 6a753ff.

.changeset/support_operation_scope_alternatives.md
Staticsubh: Bumped to `minor` and added a sentence calling out the builder signature change for embedders explicitly. See 6a753ff.

crates/apollo-mcp-server/src/auth.rs
Staticsubh: Added both: `returns_required_scopes_when_scopes_are_scattered_across_groups` (touches two groups, completes neither) and `returns_none_when_token_satisfies_more_than_one_group` (completes both).

Used a new fixture (`overlapping_multi_scope_alternatives`) with two 2-scope groups rather than reusing `alternative_required`, since its single-scope `admin` group can't represent "touches a group without completing it" — any presence of `admin` alone already completes it. See 6a753ff.

crates/apollo-mcp-server/src/auth.rs
Staticsubh: Added `insufficient_nested_scope_response()` and two tests against it, using exactly the pattern from `valid_token_with_insufficient_scopes_response()` — mocks discovery + JWKS, mints a valid HS512 token, runs it through the full middleware via `oneshot`, and asserts both the `403` status and the literal `scope="sensitive:read tenant:admin"` value in the response header for a token that's closer to one alternative than the other. See 6a753ff.

Staticsubh:
Thanks for the thorough review — addressed all nine comments in `6a753ff`:

- Collapsed `OperationRequiredScopes` from the `All`/`AnyOf` enum to a single `OperationRequiredScopes(Vec<Vec<String>>)` with a private field and a validating `new()`, per your suggestion. This closes the empty-flat-list gap (`DeleteUser: []`) and the two public-API-only states (`AnyOf` with an empty group, `AnyOf` with zero groups) at the same point, for both YAML config and direct Rust construction. `is_satisfied_by`/`challenge_scopes` are now single-branch. `From<HashMap<String, Vec<String>>>` still works for existing builder callers, now filtering out empty scope lists instead of storing a vacuous requirement.
- Fixed the `JsonSchema`/parser drift: `JsonSchema` is now implemented by hand, forwarding to the untagged helper enum used for deserialization, plus a test validating the documented YAML examples against the generated schema.
- Added a clear custom error for entries that match neither the flat nor nested shape.
- Added the two acceptance-list cases from #779 that weren't covered (scattered-across-groups-completes-none, satisfies-more-than-one-group).
- Added an integration test asserting the actual `WWW-Authenticate` `scope=` value on a real `403` response for a nested-requirement failure.
- Documented the `challenge_scopes` tie-breaking rule in `auth.mdx`.
- Bumped the changeset to `minor` and called out the builder signature change for embedders.

Local validation (pinned `1.92.0` toolchain, `AWS_LC_SYS_CMAKE_BUILDER=1`):

```sh
cargo fmt --check
cargo test -p apollo-mcp-server --lib
cargo test -p apollo-mcp-server --bin apollo-mcp-server
cargo clippy -p apollo-mcp-server --lib --bin apollo-mcp-server -- -D warnings
```

- `cargo fmt --check` passed.
- `cargo test -p apollo-mcp-server --lib` passed: 591 passed, 0 failed related to this change (6 pre-existing, unrelated failures — one network test and five `insta` snapshot-drift tests in `introspection`/`schema_walker` — reproduce identically on a clean checkout).
- `cargo test -p apollo-mcp-server --bin apollo-mcp-server` passed: 80 passed, 0 failed.
- `cargo clippy -p apollo-mcp-server --lib --bin apollo-mcp-server -- -D warnings` passed.

Replied inline on each of the nine comments with specifics. One open question I left on the schema-drift thread: the `length(min = 1)` schemars attribute only reaches the outer list, not each inner group, so `[[]]` still validates against the schema even though the parser rejects it — real enforcement is in `OperationRequiredScopes::new` either way, but let me know if you'd rather close that gap too with a dedicated newtype.

Staticsubh:
Thanks @gocamille — following up on both points from this comment:

On the type #781 should target: we took the single-type redesign you suggested (pushed in 6a753ff, and replied inline on that thread). `OperationRequiredScopes` is now `Vec<Vec<String>>` behind a private field with a validating `new()`, not the old `All`/`AnyOf` enum. So if/when #781's config-load expansion lands, that's the shape it should build on.

On the `WWW-Authenticate` challenge-header behavior: understood, and makes sense given it's a contract clients will build against. I won't preempt that with a design change — happy to wait for your team's read on it and adjust once you have it.

Also fixed a clippy failure from CI in 14da029 (my local `cargo clippy --lib --bin` missed a `#[cfg(test)]`-only lint that `--all-targets` catches) — unrelated to the design discussion, just cleaning up the red check.

Staticsubh:
@gocamille Can you please provide an update here?

gocamille:
Hi @Staticsubh, apologies for the silence here, this fell between some competing priorities on our side and that's on me. The implementation itself is in good shape; the remaining open item is the one I flagged earlier: the `WWW-Authenticate` challenge behaviour for alternatives is a contract clients will build against and I'm getting the team's read on whether we want the "fewest additional scopes" heuristic or a more operator-controlled rule (i.e. always revealing the first-listed group). I've raised it internally and will come back with a concrete answer rather than another "shortly" :sweat_smile:

In the meantime (and only if you have the bandwidth): the branch has picked up conflicts with `main` and I believe a rebase would put this in a good spot to move quickly once we settle the header question. Thank you for your patience and for the continued rigour on this one!

gocamille:
Thanks for your patience here @Staticsubh, and sorry we kept you waiting. We've made the remaining design call:

- We're happy with the nested `required_scopes` syntax and its Router-compatible OR-of-AND semantics and that's the contract we'll commit to.
- For a failed requirement with alternatives, `WWW-Authenticate` should advertise the complete first-listed alternative rather than the alternative requiring the fewest additional scopes for the presented token.

Per https://datatracker.ietf.org/doc/html/rfc6750#section-3, the `scope` attribute describes what the resource requires, not what the caller's token is missing, so the challenge shouldn't vary by token. First-listed also gives operators control over which scope set gets advertised.

One related fix while you're in there: the per-operation `403` currently inherits the global `scope_mode` hint. Since every scope in the advertised group is required, the per-operation challenge should omit `scope_mode` or report `require_all`.

Could you update `challenge_scopes` and its tests/docs accordingly? After that we can give it a final pass. Thank you again for working through all the earlier feedback, this has been a great contribution to review!

Staticsubh:
Thanks @gocamille — rebased onto `main` and addressed both points in `9762a71`:

- `challenge_scopes` now always returns the complete first-listed alternative, regardless of the presented token. Dropped the `present` parameter entirely since the challenge no longer depends on it — per RFC 6750 §3, `scope` describes what the resource requires, not what the caller is missing.
- The per-operation `403` no longer inherits the global `scope_mode` hint. Since every scope in the advertised group is required, it now always reports `scope_mode="require_all"`.

Updated `auth.mdx` and the changeset to describe both behaviors, and updated the nested-scope integration test so it actually distinguishes the two behaviors: the token is now closer to satisfying the *second* group and the global `scope_mode` is `require_any`, so the assertions (`scope="sensitive:read tenant:admin"`, `scope_mode="require_all"`) only pass under the new logic, not the old "fewest missing" heuristic.

Local validation (pinned `1.92.0` toolchain, `AWS_LC_SYS_CMAKE_BUILDER=1`):

```sh
cargo fmt --check
cargo test -p apollo-mcp-server --lib
cargo test -p apollo-mcp-server --bin apollo-mcp-server
cargo clippy -p apollo-mcp-server --all-targets -- -D warnings
```

- `cargo fmt --check` passed.
- `cargo test -p apollo-mcp-server --lib` passed: 631 passed, 0 failed related to this change (6 pre-existing, unrelated failures — the same network test and five `insta` snapshot-drift tests noted earlier in this thread — reproduce identically on a clean `main` checkout).
- `cargo test -p apollo-mcp-server --bin apollo-mcp-server` passed: 86 passed, 0 failed.
- `cargo clippy -p apollo-mcp-server --all-targets -- -D warnings` passed clean.

Let me know if you'd like the `require_all` reported some other way (e.g. omitted rather than explicit) — happy to switch, just picked the more explicit of your two suggested options.

docs/source/auth.mdx
```diff
@@ -173,15 +173,32 @@ overrides:
...
 
-`required_scopes` adds to the global requirement. It does not replace it. A token must satisfy the global `scopes` and `scope_mode` requirement. For a listed operation, the token must also carry every scope listed for that operation. This is true no matter what `scope_mode` is set to: per-operation validation always requires all listed scopes.
+`required_scopes` adds to the global requirement. It does not replace it. A token must satisfy the global `scopes` and `scope_mode` requirement. For a listed operation, the token must also carry every scope listed for that operation. This is true no matter what `scope_mode` is set to: per-operation validation always requires all listed scopes. Because of this, the `WWW-Authenticate` challenge for a per-operation `403` reports `scope_mode="require_all"` regardless of the globally configured `scope_mode`.
```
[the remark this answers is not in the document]
Staticsubh: Fixed in 856a197: updated the example `WWW-Authenticate` header to advertise the nested `UpdateUser` example's first-listed alternative (`user:write tenant:admin`) plus `scope_mode="require_all"`, matching current code behavior. Also corrected the requirement text — flat lists require every listed scope, nested alternatives require every scope in one selected inner group, not every scope across all groups.

gocamille:
Thanks @Staticsubh! The first-listed behaviour and explicit `scope_mode="require_all"` look right to me.

One more thing: it looks like the generated schema still allows `[[]]` even though runtime rejects it. Could we reject empty inner groups there too?`inner(length(min = 1))` should do it without a newtype.

Could you also clean up the avoidable clones/`expect`s and update the PR description to reflect the new behaviour? after that I think this is ready 👍

Staticsubh:
Thanks @gocamille! Addressed both in `273ec3a`:

- Added `inner(length(min = 1))` on the `AnyOf` variant's `Vec<Vec<String>>` field — the generated schema now rejects `[[]]` the same way `OperationRequiredScopes::new` already did, no newtype needed. Flipped the test that used to document this as a known gap to assert rejection instead.
- `challenge_scopes` now returns `&[String]` instead of cloning the first group and unwrapping it with an `expect()`. Falls back to an empty slice via `map_or` rather than `expect` — no clone, no panic path, not even an unreachable one.

Also updated the PR description to describe the finalized challenge behavior (first-listed alternative + explicit `scope_mode="require_all"`) instead of the earlier "fewest additional scopes" draft language, and refreshed the local test counts.

Local validation (pinned `1.92.0` toolchain, `AWS_LC_SYS_CMAKE_BUILDER=1`):

```sh
cargo fmt --check
cargo test -p apollo-mcp-server --lib
cargo test -p apollo-mcp-server --bin apollo-mcp-server
cargo clippy -p apollo-mcp-server --all-targets -- -D warnings
```

- `cargo fmt --check` passed.
- `cargo test -p apollo-mcp-server --lib` passed: 631 passed, 0 failed related to this change (same 6 pre-existing, unrelated failures noted earlier in this thread).
- `cargo test -p apollo-mcp-server --bin apollo-mcp-server` passed: 86 passed, 0 failed.
- `cargo clippy -p apollo-mcp-server --all-targets -- -D warnings` passed clean.

Let me know if there's anything else you'd like adjusted before the final pass!

gocamille (approved):
Thank you @Staticsubh, this is great! The design and implementation look good to me. You’ll just need to rebase and resolve the merge conflicts, then we can merge this 👍

Staticsubh:
Thanks again @gocamille! Rebased onto the latest `main` (resolved conflicts with #811, which touched the same doc/comment sections) as `02b3605`. Local `cargo fmt --check`, `cargo test --lib`/`--bin`, and `cargo clippy --all-targets -- -D warnings` all pass clean.

CI is green on everything that runs on this fork: Build, Run Tests, Clippy (Linux + Windows), Cargo Format, Cargo Doc, Test Coverage, codecov, CodeRabbit, and the AI Style Review. The two red checks (`Verify`, `End-to-End Tests`) are unrelated to this change — both fail on fork-PR permission restrictions (a bot comment step gets `403 Resource not accessible by integration`, and `google-github-actions/auth` doesn't get OIDC token injection on fork-triggered workflows), not on anything in the diff.

I don't have merge rights on this repo, so this is ready whenever you or another maintainer can merge it. Thanks for all the review rounds on this one!

Staticsubh:
@gocamille , Can you please merge this PR. It can again run into merge conflict if kept open for long time.

PR #784, opened 2026-07-02
Cache JWKS per issuer and rate-limit refreshes

gocamille:
## Summary

Before this change, the server fetched the issuer's JWKS from the network on every token validation. That adds latency to every authenticated request and puts unnecessary load on the identity provider. This PR fixes that, in three layers:

- **Per-issuer JWKS cache.** Keys are cached with a TTL (`jwks_cache_ttl`, default 10m), so validating a token with a known `kid` costs zero network calls while the cache is fresh.
- **Cold-miss coalescing.** If many requests hit an empty cache at once, they share a single in-flight fetch instead of each starting their own.
- **Refresh rate limit.** A token with an unknown `kid` can still trigger a refresh (that's how key rotation gets picked up), but at most once per issuer per `jwks_min_refresh_interval` (default 60s). Within the window, unknown kids just fail auth with no network call.

### Tradeoff worth knowing

`jwks_min_refresh_interval` bounds how quickly a newly rotated signing key is picked up. The docs cover how to tune it if an IdP rotates aggressively.

## Testing

All assertions are hard counts against a mock server (e.g. 1000 unique bogus kids → exactly one upstream fetch; warm path --> zero requests). No sleeps — expiry tests back-date timestamps.

## Note: Stacked on #783

This base branch is `gocamille/add-jwks-cache`, not `main`. When #783 merges, this will be retargeted to `main`.

### DIFF
--- a/crates/apollo-mcp-server/src/auth.rs
+++ b/crates/apollo-mcp-server/src/auth.rs
@@ -236,6 +236,14 @@ pub struct Config {
     #[serde(deserialize_with = "humantime_serde::deserialize", default)]
     #[schemars(with = "Option<String>")]
     pub jwks_cache_ttl: Option<Duration>,
+
+    /// Minimum time between JWKS refreshes for a single issuer. A token whose
+    /// `kid` is not in the cached JWKS triggers at most one refresh per issuer
+    /// per this interval; further misses inside the window are rejected with
+    /// 401 without any outbound request. Defaults to 60 seconds.
+    #[serde(deserialize_with = "humantime_serde::deserialize", default)]
+    #[schemars(with = "Option<String>")]
+    pub jwks_min_refresh_interval: Option<Duration>,
 }
 
 /// TLS configuration for OAuth server connections
@@ -403,6 +411,11 @@ const DEFAULT_DISCOVERY_TIMEOUT: Duration = Duration::from_secs(5);
 /// Default TTL for cached JWKS entries when `transport.auth.jwks_cache_ttl`
 /// is not configured.
 const DEFAULT_JWKS_CACHE_TTL: Duration = Duration::from_secs(600); // 10 min
+
+/// Default minimum interval between JWKS refreshes per issuer when
+/// `transport.auth.jwks_min_refresh_interval` is not configured.
+#[allow(dead_code)] // consumed in Step 3
+const DEFAULT_JWKS_MIN_REFRESH_INTERVAL: Duration = Duration::from_secs(60);
 
 /// MCP discovery methods that are allowed without authentication when
 /// `allow_anonymous_mcp_discovery` is enabled.
@@ -653,6 +666,7 @@ mod tests {
             tls: TlsConfig::default(),
             discovery_timeout: None,
             jwks_cache_ttl: None,
+            jwks_min_refresh_interval: None,
             discovery_headers: HeaderMap::new(),
         }
     }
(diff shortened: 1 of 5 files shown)


crates/apollo-mcp-server/src/auth.rs
```diff
@@ -546,6 +558,10 @@ async fn oauth_validate(
 
     let jwks_cache_ttl = auth_config.jwks_cache_ttl.unwrap_or(DEFAULT_JWKS_CACHE_TTL);
 
+    let jwks_min_refresh_interval = auth_config
+        .jwks_min_refresh_interval
+        .unwrap_or(DEFAULT_JWKS_MIN_REFRESH_INTERVAL);
```
DaleSeo, line 563: What happens if `jwks_cache_ttl` is set lower than `jwks_min_refresh_interval`? Once the entry goes stale, would a valid token get rejected until the refresh window reopens? Worth guarding or documenting?

> - Decrease the TTL if your identity provider rotates signing keys frequently and you cannot tolerate keys that were rotated mid-TTL still being accepted until the next miss. Note that revocation of a still-cached `kid` is not picked up until the entry expires.

I'm flagging this because the docs kind of suggest lowering `jwks_cache_ttl` for frequently-rotating IdPs.

docs/source/config-file.mdx
```diff
@@ -318,6 +318,7 @@ These fields are under the top-level `transport` key, nested under the `auth` ke
 | `discovery_timeout`               | `Duration`            | `5s`          | Timeout for authorization server metadata discovery requests. Supports human-readable durations (e.g., "5s", "10s", "30s").                                                                                      |
 | `discovery_headers`               | `Map<string, string>` | `{}`          | Custom headers to include in OIDC discovery and JWKS requests. Useful when upstream OAuth servers or WAFs require headers like `User-Agent`. See [discovery headers](/apollo-mcp-server/auth#discovery-headers). |
 | `jwks_cache_ttl`                  | `Duration`            | `10m`         | How long a cached JWKS entry stays fresh before the next request forces a re-fetch. See [JWKS cache](/apollo-mcp-server/auth#jwks-cache).                                                                        |
+| `jwks_min_refresh_interval`       | `Duration`            | `60s`         | Minimum time between JWKS refreshes per issuer. Cache misses inside the window are rejected with 401 without an upstream request. See [JWKS refresh rate limit](/apollo-mcp-server/auth#jwks-refresh-rate-limit). |
```
DaleSeo, line 321: Should this be public config this release, or stay internal?

Config keys are public API. Once this ships, we'll either keep it or formally deprecate it. However, adding a config key later will still be backward compatible. My concern is that If we ever change the implementation, this knob may not map cleanly onto the new behavior, yet we cannot drop it without a breaking change.

How about we expose this only if real demand shows up? The security benefit of this work comes from the behavior, not from the interval being user-tunable.

crates/apollo-mcp-server/src/auth.rs
DaleSeo: Oh, Claude already pointed this out too: https://github.com/apollographql/apollo-mcp-server/pull/784#discussion_r3516254726

DaleSeo:
@gocamille I guess https://github.com/apollographql/apollo-mcp-server/pull/783#issuecomment-4878597792 caused a conflict on this PR? idk lol

docs/source/config-file.mdx
gocamille: I was thinking of this as well @DaleSeo, this is certainly a one-way door and introduces more of the OAuth/JWKS mechanics that in fact may not be necessary for someone to configure when setting up OAuth 👍 I will change this to keep it internal.

PR #785, opened 2026-07-06
fix: migrate TheSpaceDevs example to new router

DaleSeo:
The TheSpaceDevs example in our repo still uses the router, which is under Watson's personal account. It looks like he recently shut it down, so the example doesn't work anymore. This PR updates the example to point to the-space-devs-router.up.railway.app, which is hosted in our company's Railway account. It has a slightly different schema, so I modifed the operations to match.


gocamille (approved):
Thank you for the fix, @DaleSeo !

Issue #787, opened 2026-07-09
Support for legacy/internal IdPs that don't publish discovery metadata


Staticsubh:
## Problem

Apollo MCP Server's `transport.auth.servers` requires every configured authorization server to expose an RFC 8414 (`/.well-known/oauth-authorization-server`) or OIDC Discovery (`/.well-known/openid-configuration`) metadata endpoint. Apollo MCP Server tries both, in a fixed priority order, and uses whichever responds first — but there's no way to configure a server that exposes neither.

Some authorization servers genuinely don't expose either endpoint — for example, internal or legacy identity providers that issue OAuth-shaped JWTs without publishing standard discovery metadata. Operators in that position currently have no supported path: they either can't use `transport.auth` for that server at all, or have to stand up and maintain a separate shim service that fakes a discovery document just so Apollo MCP Server will talk to the real IdP.

By contrast, Apollo Router's JWT authentication plugin (`authentication.router.jwt`) never performs discovery at all — every JWKS source in `jwks` is configured with an explicit `url`, and `issuers`/`audiences` are optional post-verification allowlists, not a lookup mechanism. Apollo MCP Server has no equivalent for operators whose authorization server matches Router's model rather than the discovery-only model.

## Desired config syntax

Add a `static_servers` list alongside the existing `servers` list, for authorization servers configured with pre-resolved discovery metadata:

```yaml
transport:
  auth:
    servers:
      - https://idp-with-discovery.example.com
    static_servers:
      - issuer: https://idp-without-discovery.example.com
        jwks_uri: https://idp-without-discovery.example.com/oauth2/v1/keys
```

Semantics:

- `issuer` and `jwks_uri` are exactly the two fields Apollo MCP Server already extracts from a discovery response, supplied directly instead of fetched.
- Apollo MCP Server fetches signing keys from `jwks_uri` directly and skips the discovery request entirely for that server.
- `issuer` plays the same role a discovery response's `issuer` field plays today: when `issuers` is configured, a token's `iss` claim must match it, bound to the server whose JWKS verified the signature (the existing cross-server binding from #757).
- Caching, singleflight, and the per-issuer refresh rate limit all apply to `static_servers` the same way they apply to `servers` — only the discovery HTTP round-trip is skipped, not the JWKS fetch or its surrounding protections.
- `servers` and `static_servers` can be used together; a token is checked against every configured server (of either kind), same as today.
[...]

PR #788, opened 2026-07-09
feat(auth): support authorization servers without a discovery endpoint

Staticsubh:
Refs #787.

Draft PR for maintainer review of the config shape.

## Summary

Adds `transport.auth.static_servers`, for authorization servers that don't expose an RFC 8414 or OIDC discovery endpoint.

- `servers` is unchanged — same type, same discovery behavior, same tests.
- `static_servers` is new: each entry sets `issuer` and `jwks_uri` directly, the same two fields a discovery response otherwise provides.
- Apollo MCP Server fetches signing keys from `jwks_uri` directly for these entries, skipping the discovery round trip. Caching, singleflight, rate-limiting, and issuer binding are all unmodified, shared code paths.
- `servers` and `static_servers` can be used together; both contribute to the `authorization_servers` field in this server's RFC 9728 protected-resource metadata.

## Behavior

```yaml
transport:
  auth:
    servers:
      - https://idp-with-discovery.example.com
    static_servers:
      - issuer: https://idp-without-discovery.example.com
        jwks_uri: https://idp-without-discovery.example.com/oauth2/v1/keys
```

## Local Testing

Tested locally via the pinned `1.92.0` toolchain (`AWS_LC_SYS_CMAKE_BUILDER=1` for `aws-lc-sys`):

```sh
cargo fmt --check
cargo test -p apollo-mcp-server --lib
cargo test -p apollo-mcp-server --bin apollo-mcp-server
cargo clippy -p apollo-mcp-server --lib --bin apollo-mcp-server -- -D warnings
```

Results:

- `cargo fmt --check` passed.
- `cargo test -p apollo-mcp-server --lib` passed: 602 passed, 0 failed related to this change (5 pre-existing, unrelated `insta` snapshot failures in `introspection`/`schema_walker` reproduce identically on a clean checkout of `main`).
- `cargo test -p apollo-mcp-server --bin apollo-mcp-server` passed: 81 passed, 0 failed.
- `cargo clippy -p apollo-mcp-server --lib --bin apollo-mcp-server -- -D warnings` passed.


Staticsubh:
@DaleSeo Can you please review this PR

gocamille:
Hi @Staticsubh, thank you for following up, and sorry for the silence here! I’ve posted my current read on the design in #787.

The short version is that after digging further into the MCP discovery requirements `static_servers` isn’t something we can merge as proposed: advertising an auth server with no discovery endpoint leaves conformant MCP clients with no path to authorize.

There may still be an underlying token-validation use case worth supporting but I want to discuss that with the team before suggesting a different direction or asking you to rework anything. One of the maintainers I need in that conversation is away this week, so please hold off on rebasing or making further changes for now. I’ll come back with a concrete answer once we’ve had that discussion. Thank you for your patience!

Staticsubh:
Hi @gocamille, thanks for digging into this — that's a real gap in the original proposal, not a nitpick. If a `static_servers` entry lands in `authorization_servers` in the protected-resource metadata, a conformant client will try RFC 8414/OIDC discovery against it per the MCP spec and have nowhere to go. That defeats the point of the field for exactly the case it was meant to serve, so I agree this needs to be rethought at the protocol boundary rather than patched within the current shape.

Understood on holding off — I won't rebase or push further changes on this branch until you're back with a direction.

On the "validating already-issued tokens" thread you mentioned: if that ends up being the shape, my read is it'd need to stay entirely out of `authorization_servers`/PRM — i.e., a way to accept tokens from an issuer whose JWKS we know directly, without ever telling a client "go authorize here." Happy to help think through that (or whatever else the team lands on) once you've had the discussion — no rush, and thanks for taking this as seriously as you did.

Staticsubh:
@gocamille Any update on this PR?

gocamille:
Hi @Staticsubh, thanks for your patience, and sorry this took a while. We've landed and it's a no on `static_servers` for the reason you named yourself: that anything we list in `authorization_servers` is a promise that a client can discover an authorization endpoint there, and an entry without discovery metadata breaks that. No config shape fixes it.

The validation-only shape you sketched afterwards is a real design and I think you framed it correctly. It's not something we're taking on right now because supporting tokens obtained outside the MCP authorization flow is a product decision on our side, not one a PR can answer. I'm leaving #787 open so the need stays on the record, and come back to it there if that changes 👍 

I'm going to close this PR so it isn't sitting in limbo and so you're not carrying a branch while this is unsettled. The discovery shim is still the pattern for now, I know that's the answer you already had, and I'm sorry I don't have a better one today 🙏 

Thank you for the #787 write-up. It was thorough enough that it made the problem with the approach easy to see, which is the best kind of proposal to get!

PR #791, opened 2026-07-17
feat: validate authorization server metadata issuer

DaleSeo:
As a follow up to PR #757, this PR adds the [RFC 8414 section 3.3](https://datatracker.ietf.org/doc/html/rfc8414#section-3.3) check that was missing: the advertised `issuer` must match the server identifier the discovery URL was built from. On a mismatch the metadata is now rejected before any JWKS fetch or cache write.

### DIFF
--- a/.changeset/validate_authorization_server_metadata_issuer.md
+++ b/.changeset/validate_authorization_server_metadata_issuer.md
@@ -4,4 +4,4 @@ default: patch
 
 # Validate authorization server metadata issuer
 
-Apollo MCP Server now checks that the `issuer` in an authorization server's discovery metadata matches the server it was fetched from, as required by RFC 8414. Metadata that advertises a mismatched issuer is rejected before its keys are trusted, so a token cannot be bound to an issuer identity the signing server did not actually claim. Servers whose discovery document returns a matching issuer (the normal case) are unaffected.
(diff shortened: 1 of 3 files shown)


.changeset/validate_authorization_server_metadata_issuer.md
```diff
@@ -0,0 +1,7 @@
+---
+default: patch
+---
+
+# Validate authorization server metadata issuer
+
+Apollo MCP Server now checks that the `issuer` in an authorization server's discovery metadata matches the server it was fetched from, as required by RFC 8414. Metadata that advertises a mismatched issuer is rejected before its keys are trusted, so a token cannot be bound to an issuer identity the signing server did not actually claim. Servers whose discovery document returns a matching issuer (the normal case) are unaffected.
```
tninesling: Is it worth linking the RFC here (like you did in the doc comment) so it appears in the changelog?

tninesling (approved):
Looks good, just one small question about docs

.changeset/validate_authorization_server_metadata_issuer.md
DaleSeo: Good call, @tninesling! Added.

PR #795, opened 2026-07-22
Adding hybrid search support

alocay:
## Summary

Adds **hybrid schema search** to the MCP `search` tool: operation-anchored lexical BM25 fused with dense-vector semantic search via Reciprocal Rank Fusion (RRF), plus a shared Postgres embedding cache so restarts (and other replicas) skip re-embedding. Everything degrades gracefully to lexical-only if the embedder or cache is unavailable.

Stacked on `tninesling/field-search`.

## What changed

**`apollo-schema-index` — operation-anchored lexical search**
- New `OperationRef` + `SchemaSearch` trait. The unit of retrieval is now an *invocable operation* (a root Query/Mutation field), not an arbitrary schema field. `OperationRef` is the shared, hashable hit identity both backends return, and the fusion/dedupe key.
- Indexing rewritten to one enriched document per operation (name, args, return type, description, and a bounded return-type flatten). The old per-field up-walk / `PathNode` machinery is deleted.
- Prefix-derived service `scope` (`derive_scope`, e.g. `slack_userByEmail` → `slack`), used for filtering and stripped from the scored text.
- Search index now rebuilds on schema reload (fixes the stale-index-on-reload bug).

**`apollo-schema-search` — new crate: semantic half + fusion**
- `Embedder` trait — `FastembedEmbedder` (ONNX `bge-small-en-v1.5`, 384-dim) for prod, `FakeEmbedder` for offline tests. Native ONNX is confined to this crate.
- `VectorStore` / `InMemoryVectorStore` — brute-force cosine, L2-normalized, scope pre-filter.
- `VectorSearch` — semantic `SchemaSearch` backend over the **same** operation corpus BM25 indexes (`enumerate_operation_documents`), so both rank identical content.
- `HybridSearch` + `rrf_fuse` — fuses N `SchemaSearch` backends via RRF (`score = Σ 1/(k + rank + 1)`, default `k = 60`) with a deterministic tiebreaker.

**Embedding cache (Postgres, shared)**
- `EmbeddingStore` trait with a `PostgresCache` backend. Content-addressed by SHA-256 of each operation's document text, so only changed operations re-embed — incremental by construction.
- Rows are **generation-keyed** by the exact `(model_id, dim, dtype, doc_builder_ver, op_key)` tuple, so a model / dimensionality / format change starts a new generation without deleting or colliding with existing rows. Concurrent writers are idempotent (`ON CONFLICT DO NOTHING`), so multiple replicas can populate one shared cache safely — the first pays the cold embed, the rest reuse.
- Fail-open: any connection or I/O error falls back to embedding from scratch.
- Enabled via `introspection.search.semantic.cache_url` (a Postgres connection string); unset = embed on every start.

**MCP tool + config**
- `search.rs` composes `HybridSearch[lexical, semantic]` with two layers of graceful degradation: a backend that errors mid-search is skipped, and a failed semantic build drops the whole tool to lexical-only. An unknown `scope` falls back to a global search.
- The index build (model load, embedding, and the synchronous Postgres client) runs off the async runtime via `spawn_blocking`, in both initial build and schema-reload rebuild — so it can't stall the executor or panic the sync PG client on a runtime thread.
- Config under `introspection.search`: `enabled`, `default_limit` / `max_limit`, `flatten_depth`, `semantic.{enabled, model, inference_threads, cache_url}`, `hybrid.rrf_k`.

**Image packaging (`Dockerfile`)**
- Bumped both build and runtime stages to Debian 13 (trixie) so the ONNX Runtime prebuilt links (bookworm/glibc 2.36 can't).
- Bakes the `bge-small-en-v1.5` model into the image (only the files fastembed reads, ~128 MB) with `HF_HUB_OFFLINE=1`, so a locked-down pod never needs HuggingFace egress; sets `HOME=/tmp` for the non-root user's HF lock file.
- Verified: image builds, runs semantic search fully offline (`--network=none`, embeds from the baked model, no lexical-only fallback), and deployment-style against Postgres — cold boot embeds + writes, warm restart logs `Loaded all embeddings from cache`, `/mcp` serves.

## Abstraction goals
- **`OperationRef` + `SchemaSearch`** is the seam that lets the lexical and semantic backends be fused and deduped uniformly (and lets more backends be added later) — one typed hit, one search contract.
- **`Embedder` / `VectorStore` / `EmbeddingStore`** isolate ONNX inference, vector storage, and cache persistence behind traits, so tests run offline with doubles and the native/DB dependencies stay contained.

## Not included / follow-ups
- **Helm (in `constellation-runtime`, separate repo):** the `embedding-db` Postgres `StatefulSet` + Service + Secret, raising the mcp memory limit, a `startupProbe` for the ~140 s cold embed, wiring `EMBEDDING_CACHE_DATABASE_URL` + `cache_url`, and `maxSurge: 1` for cold-start serialization.
- **CI (optional):** an offline smoke-test job (`docker run --network=none`, assert semantic initialized, not lexical-only). Multi-arch is already handled by the existing per-arch workflow.
- **`load-dynamic` (hardening, optional):** switch `ort` off `download-binaries` so the ONNX Runtime is `dlopen`'d at runtime — decouples the build from the base glibc and makes a missing/incompatible runtime degrade gracefully instead of failing the build. Contained to this repo's Cargo + Dockerfile; not required to ship.
- **Cross-replica cold-start coordination:** a session-scoped Postgres advisory lock so exactly one replica embeds in every scenario (not just rollouts). Concurrent cold embeds are correct today, just redundant.


adamd-apollo:
@claude perform an adversarial code review

adamd-apollo:
https://github.com/mdg-private/constellation-runtime/commit/74c694fb49aefd595e0daeaa2e073378444faeea has the tool descriptions in constellation-runtime helm chart.  I think that overrides any tool description changes that are made in the crates in this code review.  Perhaps @claude can verify how this works.

crates/apollo-mcp-server/src/introspection/tools/introspect.rs
```diff
@@ -125,7 +125,7 @@ fn tool_description(
         "Get GraphQL type information - T=type,I=input,E=enum,U=union,F=interface;s=String,i=Int,f=Float,b=Boolean,d=ID;@D=deprecated;!=required,[]=list,<>=implements;".to_string()
     } else {
         format!(
-            "Get information about a given GraphQL type defined in the schema. Instructions: Use this tool to explore the schema by providing specific type names. Start with the root query ({}) or mutation ({}) types to discover available fields. If the search tool is also available, use this tool first to get the fields, then use the search tool with relevant field return types and argument input types (ignore default GraphQL scalars) as search terms.",
+            "Get information about a specific GraphQL type by name — its exact fields and argument shapes. Use search to discover operations and types; use this tool to confirm an operation you've already found before executing it. Not for browsing the schema from the root {}/{} types.",
```
adamd-apollo: I understand that you're making these changes here because it's on a branch and not the public mcp server.  We're going to want to resolve discrepancies between this and the hints done in https://github.com/mdg-private/constellation-runtime/commit/74c694fb49aefd595e0daeaa2e073378444faeea, which were curated with several rounds of benchmarking.

adamd-apollo: and we're going to want to converge on the right place to put these kinds of instructions instead of having them live in two places.  In the end, it probably ends up in the Constellation Gateway component we're intending to build.

crates/apollo-mcp-server/src/runtime/introspection.rs
```diff
@@ -70,10 +80,59 @@ impl Default for SearchConfig {
             leaf_depth: 1,
             minify: false,
             hint: None,
+            default_limit: 10,
```
adamd-apollo, line 83: this was previously 5, no?  did you have any metrics around whether this increase helps in results?

crates/apollo-mcp-server/src/server/states/running.rs
```diff
@@ -131,6 +131,15 @@ impl Running {
         // Update the schema itself
         *self.schema.write().await = schema;
 
+        // Rebuild the search index so it reflects the new schema instead of
+        // serving stale operation hits from the previous schema.
+        if let Some(search) = &self.search_tool {
+            let schema_guard = self.schema.read().await;
+            if let Err(error) = search.rebuild(&schema_guard).await {
+                error!("Failed to rebuild search index on schema update: {error}");
+            }
```
adamd-apollo: would this potentially block for ~140s on a complete rebuild?  would there potentially be things in-flight that this would be blocking?

crates/apollo-mcp-server/src/server/states/running.rs
```diff
@@ -131,6 +131,15 @@ impl Running {
         // Update the schema itself
         *self.schema.write().await = schema;
 
+        // Rebuild the search index so it reflects the new schema instead of
+        // serving stale operation hits from the previous schema.
+        if let Some(search) = &self.search_tool {
+            let schema_guard = self.schema.read().await;
+            if let Err(error) = search.rebuild(&schema_guard).await {
+                error!("Failed to rebuild search index on schema update: {error}");
```
adamd-apollo: a failed rebuild logs and continues.  if it fails, wouldn't the search index still be the stale one?  is that acceptable?

crates/apollo-mcp-server/src/introspection/tools/search.rs
```diff
@@ -52,25 +81,151 @@ pub enum IndexingError {
...
+            None
+        }
+    }
+}
+
+/// Build the search backend from a schema. With an embedder, composes
+/// `HybridSearch[lexical, semantic]`; if the semantic index fails to build,
+/// logs a warning and degrades to lexical-only. Without an embedder, returns
+/// the bare lexical index. Both branches yield a `SchemaSearch`.
+#[allow(clippy::too_many_arguments)]
```
adamd-apollo, line 107: why is this here? seems like we ought to be able to fix this in code rather than allowing

crates/apollo-mcp-server/src/introspection/tools/search.rs
```diff
@@ -52,25 +81,151 @@ pub enum IndexingError {
...
+            "Search the federated GraphQL schema for operations and types relevant to a query. Returns complete type definitions, including all related types needed to construct GraphQL operations. Hybrid retriever: a lexical (keyword/BM25) arm and a semantic (dense-embedding) arm combined with Reciprocal Rank Fusion — provide a short natural-language query that expresses one intent and contains the domain nouns you expect in the target operation or type name; because either arm can surface a result, exact-match identifiers still rank even when phrasing is imperfect. Prefer one query per domain. Instructions: This is the tool for finding operations and types — use it, not introspect, for discovery. Reach for introspect only to confirm the exact fields and argument shapes of an operation you've already found here. Avoid repeating queries you've already run, for more efficient exploration.{}",
```
adamd-apollo: same as other comment about potentially conflicting with my previous PR

crates/apollo-mcp-server/src/introspection/tools/search.rs
```diff
@@ -79,69 +234,134 @@ impl Search {
...
+            embedder,
+            rrf_k,
+            cache_url,
+            model_id: model_id.to_string(),
             tool: Tool::new(SEARCH_TOOL_NAME, description, schema_from_type!(Input)),
         })
     }
 
+    /// Rebuild the search index from an updated schema, replacing the current index.
+    pub async fn rebuild(&self, schema: &Valid<Schema>) -> Result<(), IndexingError> {
```
adamd-apollo, line 262: is there a test for this?

crates/apollo-mcp-server/src/introspection/tools/search.rs
```diff
@@ -52,25 +81,151 @@ pub enum IndexingError {
...
+    ) -> Result<Self, IndexingError> {
+        // Model load, embedding, and Postgres I/O are all blocking — and the sync
+        // Postgres client cannot even be created on a runtime thread — so run the
+        // whole construction off the async runtime.
+        let description_hint = description_hint.map(str::to_string);
+        let semantic_model = semantic_model.to_string();
+        tokio::task::spawn_blocking(move || {
+            let embedder: Option<Arc<dyn Embedder>> = if semantic_enabled {
+                match FastembedEmbedder::new(&semantic_model, semantic_inference_threads) {
+                    Ok(e) => Some(Arc::new(e)),
```
adamd-apollo, line 175: should we have some log line when Ok so we can track when it *does* succeed?  and maybe how often it happens?

crates/apollo-schema-index/src/backend.rs
```diff
@@ -0,0 +1,84 @@
+//! The search backend contract shared by lexical and (Phase 2) semantic search.
+
+use crate::OperationType;
+use crate::error::SearchError;
+use crate::path::Scored;
+use std::collections::BTreeSet;
+
+/// A retrievable operation: a root Query/Mutation field the agent can invoke.
+/// Identity for fusion/dedupe is `(operation_type, field_name)`.
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
```
adamd-apollo, line 10: doc seems to diverge from actual

alocay:
> [mdg-private/constellation-runtime@74c694f](https://github.com/mdg-private/constellation-runtime/commit/74c694fb49aefd595e0daeaa2e073378444faeea) has the tool descriptions in constellation-runtime helm chart. I think that overrides any tool description changes that are made in the crates in this code review. Perhaps @claude can verify how this works.

I believe the `hint` field only appends to the existing description and doesn't act like an override.

crates/apollo-mcp-server/src/introspection/tools/search.rs
alocay: Figured for some short-term code it wouldn't be a big deal. The alternative is just creating a struct, storing all the config values in there, and then passing that around. Didn't seem worth it for code that will eventually be moved.

crates/apollo-mcp-server/src/server/states/running.rs
alocay: I would argue yes. It would use a stale search which is the current behavior of the mcp server search (today it doesn't try to rebuild on an updated schema notification I believe). For the short term Ithink this is fine, we can decide how to address this properly when we move to a service.

crates/apollo-mcp-server/src/server/states/running.rs
alocay: Good catch, this is likely a problem although 140s would be the worst case scenario. Let me look into it.

alocay: It seems for this we don't want the `operations_lock` to remain held on to so we don't block any other incoming list tools, tool calls, or initialize request. I'll add a clone and move the drop to occur above the rebuild so it frees things up instead of blocking.

crates/apollo-mcp-server/src/introspection/tools/search.rs
alocay: No there isn't a dedicated test for this. There is one for `build_backend()` which is what's mainly called here.

crates/apollo-mcp-server/src/introspection/tools/search.rs
alocay: I've reverted this for now. We can visit these changes later.

calvincestari:
I can't find anything that needs to be addressed beyond the comments already raised. The change looks good to me, it's understandable.

crates/apollo-mcp-server/src/introspection/tools/introspect.rs
alocay: I reverted these tool descriptions for now, we can come back to them later.

PR #796, opened 2026-07-22
fix: reject auth under stdio transport

DaleSeo:
An `auth` block nested under a `stdio` transport was parsed successfully and then silently dropped, so an operator could believe authentication was configured while the server ran with none of it applied. The root cause is that `Transport` is an internally tagged enum, and serde only enforces `deny_unknown_fields` on struct variants, not on the `Stdio` unit variant, so extra keys under `stdio` were ignored instead of rejected. This changes `Stdio` into an empty struct variant so misplaced keys now fail to parse, matching how top-level `auth` and the `streamable_http` transport already reject unknown fields.

### DIFF
--- a/.changeset/reject_auth_under_stdio_transport.md
+++ b/.changeset/reject_auth_under_stdio_transport.md
@@ -1,0 +1,7 @@
+---
+default: patch
+---
+
+# Reject auth configuration under the stdio transport
+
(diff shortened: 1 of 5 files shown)


gocamille (approved):
Thank you @DaleSeo !

PR #798, opened 2026-07-23
feat: add graphos schema source for non-federated graphs

DaleSeo:
Non-federated, or monograph, graphs do not have a live schema path today. Uplink only distributes composed supergraphs, and monographs never compose, so the only option has been a manually maintained local file.

This PR adds a `schema` configuration with `source` set to `graphos`. It fetches the latest publication for the configured variant from the Platform API using the existing `graphos` credentials. That gives registered monographs live schema delivery and tool generation.

## Testing

Manually testing a canary build with GraphOS MCP Server


gocamille (approved):
Thank you for this @DaleSeo ! The logic is sound and I've tested the happy/sad path of this flow.

PR #800, opened 2026-07-28
chore: upgrade rmcp to 3

### LINKED ISSUE #803: Stateful transports can advertise a protocol version newer than the server implements
## Description

Follow-up from #802, which capped protocol version negotiation at `MAX_SUPPORTED_PROTOCOL_VERSION` (currently `2025-11-25`).

The cap holds on stateless streamable HTTP and stdio, where our `initialize` handler's negotiation result is what the client sees. On **stateful** streamable HTTP sessions, rmcp's `serve_server` handshake re-negotiates after our handler runs and echoes any version in its `KNOWN_VERSIONS` — including `2026-07-28`, which this server doesn't implement (SEP-2243 standard headers, discovery requests). A stateful client told `2026-07-28` may send follow-up requests we can't handle.

This divergence is currently pinned by the `stateful_echoes_newer_known_version_due_to_rmcp_override` test in `crates/apollo-mcp-server/src/server/states/running.rs`, and the `rmcp_known_versions_audit` canary forces a review when rmcp's version list changes — but neither closes the gap.

## Possible directions

- Enforce the supported-version cap in the shared transport layer so stateful handshakes can't advertise versions above `MAX_SUPPORTED_PROTOCOL_VERSION`
- Upstream: ask rmcp for a way to configure the maximum negotiated version rather than always using `KNOWN_VERSIONS`
- If the server adopts the `2026-07-28` revision before either of the above, bumping `MAX_SUPPORTED_PROTOCOL_VERSION` makes the gap moot

## References

- #794 (original bug)
- #802 (stateless fix + pinning tests)

DaleSeo:
Upgrades rmcp from 2.1 to 3.2.0, the latest release.

This is a major version jump, so it touches several APIs: `Meta` is renamed to `MetaObject`, `Tool` drops its `execution` field, `ListToolsResult`/`ListResourcesResult` are now built with `with_all_items` instead of constructed by hand, the corresponding `ServerHandler` methods return `CallToolResponse`/`ReadResourceResponse`/`GetPromptResponse` instead of their `*Result` types (mapped in via `Into::into`), and `StreamableHttpServerConfig::with_stateful_mode` is renamed to `with_legacy_session_mode`.

It also adds a `ServerHandler::supported_protocol_versions` hook that rmcp consults whenever it re-negotiates the protocol version after `initialize`, on every transport. This server overrides it to cap at the newest protocol version it implements (2025-11-25), closing a gap where stateful sessions and stdio could echo back a version from rmcp's `KNOWN_VERSIONS` that this server doesn't actually support, such as `2026-07-28`.

Closes #803

### DIFF
--- a/crates/apollo-mcp-server/src/server/states/running.rs
+++ b/crates/apollo-mcp-server/src/server/states/running.rs
@@ -1,5 +1,6 @@
+use std::borrow::Cow;
 use std::collections::HashMap;
-use std::sync::Arc;
+use std::sync::{Arc, LazyLock};
 
 use apollo_compiler::{Schema, validation::Valid};
 use opentelemetry::KeyValue;
@@ -7,9 +8,10 @@ use parking_lot::Mutex;
 use reqwest::header::HeaderMap;
 use rmcp::ErrorData;
 use rmcp::model::{
-    ClientCapabilities, Extensions, GetPromptRequestParams, GetPromptResult, Implementation,
-    ListPromptsResult, ListResourcesResult, PromptMessage, PromptsCapability, ReadResourceResult,
-    ResourcesCapability, Role, ToolsCapability,
+    CallToolResponse, ClientCapabilities, Extensions, GetPromptRequestParams, GetPromptResponse,
+    GetPromptResult, Implementation, ListPromptsResult, ListResourcesResult, PromptMessage,
+    PromptsCapability, ReadResourceResponse, ReadResourceResult, ResourcesCapability, Role,
+    ToolsCapability,
 };
 use rmcp::{
     Peer, RoleServer, ServerHandler, ServiceError,
@@ -289,10 +291,8 @@ impl Running {
             let app = self.apps.iter().find(|app| app.name == app_name);
 
             match app {
-                Some(app) => ListToolsResult {
-                    next_cursor: None,
-                    tools: self
-                        .operations
+                Some(app) => ListToolsResult::with_all_items(
+                    self.operations
                         .read()
                         .await
                         .iter()
@@ -311,8 +311,7 @@ impl Running {
                                 .collect::<Vec<_>>(),
                         )
                         .collect(),
-                    meta: None,
-                },
+                ),
                 None => {
                     return Err(McpError::new(
                         ErrorCode::INVALID_REQUEST,
@@ -322,10 +321,8 @@ impl Running {
                 }
             }
         } else {
-            ListToolsResult {
-                next_cursor: None,
-                tools: self
-                    .operations
+            ListToolsResult::with_all_items(
+                self.operations
                     .read()
                     .await
                     .iter()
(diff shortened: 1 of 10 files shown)


crates/apollo-mcp-server/src/server/states/running.rs
```diff
@@ -669,20 +660,36 @@ impl ServerHandler for Running {
...
+    /// version from rmcp's `KNOWN_VERSIONS` (e.g. `2026-07-28`'s SEP-2243
+    /// headers and discovery, which this server doesn't yet handle).
+    fn supported_protocol_versions(&self) -> Cow<'static, [ProtocolVersion]> {
+        Cow::Owned(
+            ProtocolVersion::KNOWN_VERSIONS
+                .iter()
+                .filter(|version| **version <= MAX_SUPPORTED_PROTOCOL_VERSION)
+                .cloned()
+                .collect(),
+        )
```
inanna-apollo: this is totally just code golf, but this could be a clone-once-per-program since it's pulling from a constant, eg something like:

```
fn supported_protocol_versions(&self) -> Cow<'static, [ProtocolVersion]> {
    static SUPPORTED: LazyLock<Vec<ProtocolVersion>> = LazyLock::new(|| {
        ProtocolVersion::KNOWN_VERSIONS
            .iter()
            .filter(|v| **v <= MAX_SUPPORTED_PROTOCOL_VERSION)
            .cloned()
            .collect()
    });
    Cow::Borrowed(&SUPPORTED)
}
```

inanna-apollo (changes requested):
looks really good, one (code golf tier) note around reducing allocations (also the automated review bot seems to have found a stale reference in AGENTS.md, specifically a mention of rmcp 2.1 in https://github.com/apollographql/apollo-mcp-server/blob/main/AGENTS.md#key-dependencies)

crates/apollo-mcp-server/src/server/states/running.rs
DaleSeo: @inanna-apollo Good catch, thanks!

crates/apollo-mcp-server/src/server/states/running.rs
```diff
@@ -2973,14 +2981,11 @@ mod integration_tests {
...
+        async fn stateful_caps_at_max_supported_when_client_requests_newer_known_version() {
             // rmcp's handshake re-negotiates after our `initialize` handler
-            // and echoes any version in its KNOWN_VERSIONS, overriding our
-            // cap at MAX_SUPPORTED_PROTOCOL_VERSION. This test pins that
-            // behavior so an rmcp upgrade that changes it is caught; if it
-            // starts failing, re-check whether the cap now holds on all
-            // transports and close #803 accordingly.
+            // on stateful paths too, but it's now bounded by
+            // `supported_protocol_versions`, so it can't echo back a version
+            // this server doesn't implement (closes #803).
```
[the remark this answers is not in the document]
DaleSeo: The suggested fix isn't quite constructible. Every legacy version in `KNOWN_VERSIONS` is already at or below `MAX_SUPPORTED_PROTOCOL_VERSION`, so there's no scenario where a legacy version needs to be capped. SEP-2567 always makes 2026-07-28 stateless, so it can never reach legacy-session negotiation in the first place. I'll rename this test to make that clear and add a new test for the code path that wasn't actually covered.

inanna-apollo (approved):
looks great, thanks!

PR #802, opened 2026-07-29
fix: negotiate protocol version in stateless streamable HTTP mode

### LINKED ISSUE #794: 1.16 does not negotiate protocol version down to 2025-06-18
Attempting to integrate with AWS AgentCore.

AgentCore sends 2025-06-18 as protocol version. Per release notes in 1.16 Apollo MCP should negotiate this protocol version, but instead it sends back 2025-11-25, which AgentCore does not understand.

See log: [apollomcp_log.txt](https://github.com/user-attachments/files/30259326/apollomcp_log.txt)
Config: [apollomcp_config.yaml](https://github.com/user-attachments/files/30259371/apollomcp_config.yaml)

gocamille:
## Summary

Fixes #794

In stateless streamable HTTP mode the server answered `initialize` with protocol version `2025-11-25` regardless of the version the client requested, breaking clients that pin an older version (i.e. AWS AgentCore requesting `2025-06-18`).

rmcp negotiates the protocol version in two places: its default `ServerHandler::initialize` and the `serve_server` handshake in its service layer. On the stateless path neither negotiation layer ran and the hardcoded `LATEST` version went out unconditionally.

 ## Changes

- Negotiate the protocol version inside our `initialize` handler: echo the client's requested version when the server supports it, otherwise fall back to the newest supported version.
- Add `MAX_SUPPORTED_PROTOCOL_VERSION`, a server-owned cap on the advertised protocol version. rmcp's `KNOWN_VERSIONS` includes 2026-07-28, which this server doesn't implement (SEP-2243 headers, discovery) so negotiation caps at what the server actually supports rather than what the SDK has constants for.
- Update comments in `initialize` and `get_info` that claimed rmcp handles negotiation on all paths.

## Tests

- `stateless_echoes_supported_version_when_client_requests_older_version`: regression test for the reported behaviour
- `stateless_negotiates_down_when_client_requests_unknown_version`: an unrecognized version falls back to the max supported version
- `stateless_caps_at_max_supported_when_client_requests_newer_known_version`: `2026-07-28` is in rmcp's `KNOWN_VERSIONS` but not implemented by this server; the stateless path caps it
- `stateful_echoes_newer_known_version_due_to_rmcp_override`: pins a known upstream divergence--on stateful paths rmcp re-negotiates after our handler and echoes `2026-07-28`; breaks if an rmcp upgrade changes this
- `rmcp_known_versions_audit`: canary that fails when an rmcp upgrade changes `KNOWN_VERSIONS`, forcing a deliberate decision on whether `MAX_SUPPORTED_PROTOCOL_VERSION` moves


crates/apollo-mcp-server/src/server/states/running.rs
```diff
@@ -613,6 +613,27 @@ impl Running {
...
+/// Echoes the client-requested protocol version when rmcp supports it,
+/// otherwise falls back to `server_fallback`.
+///
+/// Mirrors rmcp's `negotiate_protocol_version`, which is `pub(crate)` and so
+/// can't be called directly.
+fn negotiate_protocol_version(
+    client_requested: &ProtocolVersion,
+    server_fallback: ProtocolVersion,
+) -> ProtocolVersion {
+    if ProtocolVersion::KNOWN_VERSIONS.contains(client_requested) {
```
DaleSeo: `KNOWN_VERSIONS` means versions that rmcp has a constant for, not necessarily versions that Apollo MCP Server implements. In rmcp 2.1.0, `LATEST` is `2025-11-25`, while `KNOWN_VERSIONS` includes `2026-07-28`. Then 3.0 adds `STANDARD_HEADERS = V_2026_07_28` and `DiscoverRequest`, neither of which exists in 2.1.0. So if we tell a client `2026-07-28`, it may follow up with SEP-2243 headers or a discover request that we cannot handle yet.

crates/apollo-mcp-server/src/server/states/running.rs
```diff
@@ -2823,6 +2847,74 @@ mod integration_tests {
...
+            let service = create_stateless_service(running, Arc::clone(&session_manager));
+
+            let response = service
+                .oneshot(build_initialize_request("2999-01-01"))
+                .await
+                .unwrap();
+
+            assert_eq!(response.status(), StatusCode::OK);
+            let body = extract_json_body(response).await;
+            assert_eq!(body["result"]["protocolVersion"], "2025-11-25");
```
DaleSeo: Would you rather keep the literal so any `LATEST` change forces a deliberate review, or assert against `LATEST` so the test focuses only on the fallback behavior?

crates/apollo-mcp-server/src/server/states/running.rs
```diff
@@ -2823,6 +2847,74 @@ mod integration_tests {
...
+            )
+        }
+
+        #[tokio::test]
+        async fn stateless_echoes_supported_version_when_client_requests_older_version() {
+            // Regression for #794: a stateless client requesting a supported
+            // older version must get that version echoed back, not the latest.
+            let running = create_running_with_output_schema();
+            let session_manager: Arc<LocalSessionManager> = LocalSessionManager::default().into();
+            let service = create_stateless_service(running, Arc::clone(&session_manager));
```
DaleSeo: nit

```suggestion
            let service = create_stateless_service(running, LocalSessionManager::default().into());
```

crates/apollo-mcp-server/src/server/states/running.rs
gocamille: @DaleSeo I was intuitively concerned about this with the release of `2026-07-28` so thank you for identifying the recent `STANDARD_HEADERS` and `DiscoverRequest` inclusions. We should change the current negotiation approach to default to a `MAX_SUPPORTED_PROTOCOL_VERSION` constant with negotiation capping there instead of trusting `KNOWN_VERSIONS`. 

I'm also adding a test pinning that upstream divergence, plus a canary test on `KNOWN_VERSIONS` so an rmcp upgrade forces a deliberate decision about bumping the cap 👍

crates/apollo-mcp-server/src/server/states/running.rs
gocamille: @DaleSeo great question -- the rework above makes this moot in its original form since the fallback proposed is now `MAX_SUPPORTED_PROTOCOL_VERSION` rather than `LATEST`, and the deliberate-review pressure moved into the `rmcp_known_versions_audit` canary. I switched the remaining literal assertions to the proposed constant which lets these tests focus purely on fallback behaviour (see my last commit d6c4855). 

This design is a change from how we handle version negotiation though: what are your thoughts on it?

crates/apollo-mcp-server/src/server/states/running.rs
```diff
@@ -638,10 +668,13 @@ impl ServerHandler for Running {
         // TODO: how to remove these?
         let mut peers = self.peers.write().await;
         peers.push(context.peer);
-        // rmcp negotiates the protocol version during `initialize`: it echoes the
-        // client's requested version when supported and otherwise falls back to
-        // the version advertised by `get_info`.
-        Ok(self.get_info())
+        // Echo the client's requested protocol version when supported,
+        // falling back to our max supported version otherwise (#794). This
+        // result stands on stateless HTTP and stdio; on stateful sessions
```
DaleSeo:
```suggestion
        // result stands only on stateless HTTP; on stateful sessions
```

crates/apollo-mcp-server/src/server/states/running.rs
```diff
@@ -2823,6 +2900,124 @@ mod integration_tests {
...
+            assert_eq!(body["result"]["protocolVersion"], "2025-06-18");
+        }
+
+        #[tokio::test]
+        async fn stateless_negotiates_down_when_client_requests_unknown_version() {
+            // A client offering a version rmcp doesn't know must fall back to
+            // our max supported version rather than having it echoed back.
+            let running = create_running_with_output_schema();
+            let session_manager: Arc<LocalSessionManager> = LocalSessionManager::default().into();
+            let service = create_stateless_service(running, Arc::clone(&session_manager));
```
DaleSeo: Same nit here as https://github.com/apollographql/apollo-mcp-server/pull/802#discussion_r3676079120 I also see the same pattern in the next test.

.changeset/negotiate_protocol_version_in_stateless_streamable_http_mode.md
```diff
@@ -0,0 +1,7 @@
+---
+default: patch
+---
+
+# Negotiate protocol version in stateless streamable HTTP mode
+
+When running streamable HTTP in stateless mode, the server answered `initialize` with its latest supported MCP protocol version regardless of the version the client requested, breaking clients that require an older version (such as AWS AgentCore requesting `2025-06-18`). The server now echoes the client's requested protocol version when it is supported and falls back to the latest supported version otherwise, matching the negotiation behavior of the stdio and stateful HTTP transports.
```
DaleSeo: The changeset predates the cap?

DaleSeo (approved):
Thanks for addressing my feedback. Left minor comments.

crates/apollo-mcp-server/src/server/states/running.rs
gocamille: Good catch — inlined in all three, including the stateful one. 75c4340

.changeset/negotiate_protocol_version_in_stateless_streamable_http_mode.md
gocamille: Yep — it was written before the cap went in. Rewritten in 80706a5 to describe the cap, and I dropped the "matches stdio and stateful" line since the stateful path still echoes 2026-07-28 (#803).

PR #806, opened 2026-08-03
feat: Advertise a server icon in initialize

### LINKED ISSUE #805: feat: Expose server_info.icons so operators can advertise a server icon
## Motivation

The MCP protocol grew an `icons` field on the `Implementation` object in
revision `2025-11-25` so hosts (Claude Desktop, Cursor, Zed, …) can show a
per-server icon next to the server's name. `apollo-mcp-server` renders every
other `Implementation` field on `initialize` — `name`, `version`, `title`,
`description`, `website_url` — but `icons` is currently unreachable from
config, so any downstream project that wraps this binary shows up in the
host UI with the generic default icon.

Concretely we wrap `apollo-mcp-server` as the runtime for `shyftplan-mcp`
and we'd like to advertise our own brand icon in the hosts that support it.
Without a config surface we would have to fork just to call
`Implementation::with_icons(...)`, which we'd rather not do.

## Proposal

Add an optional `icons` field to the existing `server_info` config
section:

```yaml
server_info:
  name: "shyftplan MCP"
  icons:
    - src: "https://cdn.shyftplan.com/mcp-icon.svg"
      mime_type: "image/svg+xml"
      sizes: ["any"]
```

Each entry is a 1:1 mapping of the [MCP `Icon` object][mcp-icon]: `src`
(required), plus optional `mime_type`, `sizes`, and `theme` (`light` /
`dark`). The list is forwarded verbatim on the `Implementation` returned by
`get_info`, using `rmcp`'s existing `Implementation::with_icons` +
`Icon::new` builders — no new SDK surface required.

Defaults are unchanged: no icon is advertised unless `server_info.icons`
is set. Clients that don't understand the field ignore it, per the MCP
spec.

I already have this working locally against `rmcp = 2.1`, gated by the
existing `server_info` config path, with unit tests covering:

1. `icons` omitted from config → `Implementation.icons == None`.
2. `icons` set with a mix of fully-populated and minimal entries → each
   field is forwarded correctly.
3. Deserialization round-trip from YAML with only `src` set.

If the approach looks reasonable to the maintainers, I'll open a PR
against `main`. Happy to adjust the config shape (e.g. rename fields to
camelCase to match the outgoing JSON, put the field somewhere other than
`server_info`) based on your feedback before I do.

[mcp-icon]: https://modelcontextprotocol.io/specification/2025-11-25/schema#icon

alexhof-shyftplan:
Fixes #805.

> [!NOTE]
> Draft — opened alongside the linked enhancement issue so maintainers can
> see the shape of the proposal before signing off on the approach.
> CONTRIBUTING asks for agreement on features before code lands, so please
> read the issue first and I'll adjust or hold off based on your feedback.

## What

Adds an optional `server_info.icons` config option. Each entry maps to the
MCP [`Icon` object][mcp-icon] (`src` required, plus optional `mime_type`,
`sizes`, `theme`) and is forwarded verbatim on the `Implementation` returned
from `initialize`, via `rmcp`'s existing `Implementation::with_icons`
builder. No icon is advertised by default.

```yaml
server_info:
  name: "Acme MCP"
  icons:
    - src: "https://acme.com/mcp-icon.svg"
      mime_type: "image/svg+xml"
      sizes: ["any"]
```

## Why

MCP revision `2025-11-25` grew an `icons` field on `Implementation` so
hosts (Claude Desktop, Cursor, Zed, …) can render a per-server icon next
to the server's name. `apollo-mcp-server` already forwards every other
`Implementation` field (`name`, `version`, `title`, `description`,
`website_url`) but leaves `icons` unreachable, so downstream projects that
wrap this binary show up with the default host icon.

## How

- `ServerInfoConfig` grows an `icons: Vec<IconConfig>` field. `IconConfig`
  and `IconThemeConfig` mirror the SDK types 1:1, live in
  `crates/apollo-mcp-server/src/server_info.rs`, and use `#[serde(default,
  deny_unknown_fields)]` to match the surrounding config style.
- `From<&IconConfig> for rmcp::model::Icon` walks the optional fields via
  the SDK's builder methods.
- `Running::get_info` calls `impl_.with_icons(...)` when
  `ServerInfoConfig::icons()` returns `Some`.
- Docs updated in `docs/source/config-file.mdx` with the field table, YAML
  example, and per-field description.
- Changeset added at `.changeset/server_info_icons.md` (`minor`).

## Testing

New unit tests:

- `server_info::tests::icons_is_none_when_no_entries_configured`
- `server_info::tests::icons_forwards_optional_fields`
- `server_info::tests::icons_can_be_deserialized_from_yaml_with_minimal_fields`
- `server::states::running::tests::get_info::get_info_omits_icons_by_default`
- `server::states::running::tests::get_info::get_info_forwards_configured_icons_with_optional_fields`

Existing runtime snapshot (`it_merges_env_and_file_with_uplink_endpoints`)
updated for the new field. Full crate suite (`cargo test -p
apollo-mcp-server`) plus `cargo clippy --all-targets -- -D warnings` and
`cargo fmt --check` pass locally on the pinned `1.92.0` toolchain.

## Compatibility

Purely additive:

- Field is optional and defaults to an empty vec, so existing configs
  continue to serialize/deserialize unchanged.
- No new SDK surface — the `rmcp = 2.1` pin already exposes `Icon`,
  `IconTheme`, and `Implementation::with_icons`.
- Older clients (`2025-06-18` and earlier) ignore the field on the wire per
  the MCP spec, so operators can turn it on without waiting for host
  upgrades.

## Checklist

- [x] Tests added
- [x] Docs updated
- [x] Changeset added
- [ ] Approach agreed with maintainers (opened issue #805, waiting for
      sign-off before promoting from draft)

[mcp-icon]: https://modelcontextprotocol.io/specification/2025-11-25/schema#icon

### DIFF
--- a/crates/apollo-mcp-server/src/server_info.rs
+++ b/crates/apollo-mcp-server/src/server_info.rs
@@ -1,5 +1,6 @@
+use rmcp::model::{Icon, IconTheme};
 use schemars::JsonSchema;
-use serde::Deserialize;
+use serde::{Deserialize, Deserializer};
 
 /// Server metadata configuration returned in the MCP initialize response.
 /// All fields are optional and fall back to defaults if not provided.
@@ -20,6 +21,147 @@ pub struct ServerInfoConfig {
 
     /// A brief description of the server
     pub description: Option<String>,
+
+    /// Icons representing the server, for clients that display one
+    pub icons: Vec<IconConfig>,
+}
+
+/// A single icon representing the server. Clients pick whichever entry best
+/// fits the surface they are rendering, so several may be supplied.
+#[derive(Debug, Clone, Deserialize, JsonSchema)]
+#[serde(deny_unknown_fields)]
+pub struct IconConfig {
+    /// URI of the icon: an `https://` URL or a base64-encoded `data:` URI
+    #[serde(deserialize_with = "deserialize_icon_src")]
+    pub src: String,
+
+    /// MIME type of the icon, such as `image/png`. Set this when the source
+    /// serves no MIME type of its own or serves a generic one
+    #[serde(default)]
+    pub mime_type: Option<String>,
+
+    /// Sizes the icon is available in, each `WxH` (e.g. `48x48`) or `any` for
+    /// scalable formats
+    #[serde(default, deserialize_with = "deserialize_icon_sizes")]
+    pub sizes: Option<Vec<String>>,
+
+    /// Background theme this icon is designed for. Omit to let the client use
+    /// it with any theme
+    #[serde(default)]
+    pub theme: Option<IconThemeConfig>,
+}
+
+/// Reject `src` values whose scheme the MCP spec instructs clients to refuse
+/// (`javascript:`, `file:`, `ftp:`, `ws:`, and anything else that isn't
+/// `https://` or `data:`), and reject well-schemed but empty inputs like a
+/// bare `https://` or a `data:` URI without an `image/*` payload.
+fn deserialize_icon_src<'de, D>(deserializer: D) -> Result<String, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let raw = String::deserialize(deserializer)?;
+    let ok = if let Some(rest) = raw.strip_prefix("https://") {
+        has_host(rest)
+    } else if let Some(rest) = raw.strip_prefix("data:") {
+        is_image_data_uri(rest)
+    } else {
+        false
+    };
+    if ok {
+        Ok(raw)
+    } else {
+        Err(serde::de::Error::custom(format!(
+            "server_info.icons[].src must be an `https://` URL with a host or an \
+             `image/*` `data:` URI (got {raw:?})"
+        )))
+    }
+}
+
+fn has_host(after_scheme: &str) -> bool {
+    let host_end = after_scheme
+        .find(['/', '?', '#'])
+        .unwrap_or(after_scheme.len());
+    !after_scheme[..host_end].is_empty()
+}
+
+/// A `data:` URI is `mediatype[;base64],data`. Accept it only when the media
+/// type is `image/<subtype>` and the payload is non-empty; the RFC allows
+/// omitting either but neither shape makes sense for an icon source.
+fn is_image_data_uri(after_scheme: &str) -> bool {
+    let (metadata, data) = match after_scheme.split_once(',') {
+        Some(pair) => pair,
+        None => return false,
+    };
+    if data.is_empty() {
+        return false;
+    }
+    let mime = metadata.split(';').next().unwrap_or("");
+    match mime.strip_prefix("image/") {
+        Some(subtype) => !subtype.is_empty(),
+        None => false,
+    }
+}
+
+fn deserialize_icon_sizes<'de, D>(deserializer: D) -> Result<Option<Vec<String>>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let raw = Option::<Vec<String>>::deserialize(deserializer)?;
+    if let Some(entries) = &raw {
+        for entry in entries {
+            if entry != "any" && !is_wxh_size(entry) {
+                return Err(serde::de::Error::custom(format!(
+                    "server_info.icons[].sizes entries must be `WxH` (e.g. `48x48`) or `any` (got {entry:?})"
+                )));
+            }
+        }
+    }
+    Ok(raw)
+}
+
+fn is_wxh_size(entry: &str) -> bool {
+    match entry.split_once('x') {
+        Some((width, height)) => {
+            !width.is_empty()
+                && !height.is_empty()
+                && width.bytes().all(|b| b.is_ascii_digit())
+                && height.bytes().all(|b| b.is_ascii_digit())
+        }
+        None => false,
+    }
+}
+
+/// Background theme an icon is designed for.
+#[derive(Debug, Clone, Copy, Deserialize, JsonSchema)]
+#[serde(rename_all = "lowercase")]
+pub enum IconThemeConfig {
+    Light,
+    Dark,
+}
+
+impl From<IconThemeConfig> for IconTheme {
+    fn from(theme: IconThemeConfig) -> Self {
+        match theme {
+            IconThemeConfig::Light => IconTheme::Light,
+            IconThemeConfig::Dark => IconTheme::Dark,
+        }
+    }
+}
+
+impl From<&IconConfig> for Icon {
+    fn from(config: &IconConfig) -> Self {
+        let mut icon = Icon::new(config.src.clone());
+        if let Some(mime_type) = config.mime_type.as_deref() {
+            icon = icon.with_mime_type(mime_type);
+        }
+        if let Some(sizes) = config.sizes.clone() {
+            icon = icon.with_sizes(sizes);
+        }
+        if let Some(theme) = config.theme {
+            icon = icon.with_theme(theme.into());
+        }
+        icon
+    }
 }
 
 impl ServerInfoConfig {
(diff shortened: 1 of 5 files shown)


gocamille:
Thank you for this @alexhof-shyftplan and thank you for your patience as we got to #805.

Answering the question you asked first: yes, adding `icon` is something we would want here. An operator wrapping our binary should be able to advertise their own icon, and `server_info` is the right home for it. 

Your no-default call is right, and for a better reason than the one in your comment. `serverInfo` describes the running instance, which may be yours; `server.json` describes the package, which is ours. Those two diverging is correct, not an inconsistency. Keep `icons()` returning `None` when unset.

The one thing to change is the validation layer: #805 proposed a plain pass-through, but I saw that the scheme allowlist and the `WxH` check traced back to review comments against it from our own bot, twice, before the squash. You already pushed back on one of its suggestions (`mime_type`) which was the right call. A four-item copy of the spec's unsafe-scheme list on our side is a partial boundary that looks like protection and isn't.

What is needed though is the smaller version of what you're already doing. The rule should be: validate exactly what the `Icon` schema constrains, nothing more. `src` is `@format uri`, so it has to parse -- for that, use `url::Url::parse`, already a dependency and already how we validate URL config in `auth.rs`. `theme` is a closed `light`/`dark` enum, so `IconThemeConfig` stays as-is. `sizes` is `string[]` with no pattern, so `serde`'s `array` check is the whole job and the custom deserializer goes. `mimeType` is a plain string, so nothing is needed there. Every check that survives is one the schema asks for; the allowlist and the `WxH` rule are ones we invented.

crates/apollo-mcp-server/src/server_info.rs
```diff
@@ -20,6 +21,147 @@ pub struct ServerInfoConfig {
...
+/// (`javascript:`, `file:`, `ftp:`, `ws:`, and anything else that isn't
+/// `https://` or `data:`), and reject well-schemed but empty inputs like a
+/// bare `https://` or a `data:` URI without an `image/*` payload.
+fn deserialize_icon_src<'de, D>(deserializer: D) -> Result<String, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let raw = String::deserialize(deserializer)?;
+    let ok = if let Some(rest) = raw.strip_prefix("https://") {
+        has_host(rest)
```
gocamille: `has_host` is a substring scan so this admits `https://%%%` and `https://` and rejects `HTTPS://example.com/icon.png` and `data:IMAGE/PNG;base64,...`, both of which are valid. For a field marked `@format uri` that's incorrect in both directions. `url::Url::parse` would cover all for correctly, and it's already a direct dependency that we use in `auth.rs` for exactly this scenario.

crates/apollo-mcp-server/src/server_info.rs
```diff
@@ -20,6 +21,147 @@ pub struct ServerInfoConfig {
...
+/// Reject `src` values whose scheme the MCP spec instructs clients to refuse
+/// (`javascript:`, `file:`, `ftp:`, `ws:`, and anything else that isn't
+/// `https://` or `data:`), and reject well-schemed but empty inputs like a
+/// bare `https://` or a `data:` URI without an `image/*` payload.
+fn deserialize_icon_src<'de, D>(deserializer: D) -> Result<String, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let raw = String::deserialize(deserializer)?;
+    let ok = if let Some(rest) = raw.strip_prefix("https://") {
```
gocamille: Drop the scheme allowlist: `http://localhost:3000/icon.png` should work; our own default endpoint is `http://localhost:4000` and every local-dev doc example is plain http. Nothing in the spec forbids `http:` for icons; the MUST that names schemes is scoped to the unsafe ones and doesn't list it.

crates/apollo-mcp-server/src/server_info.rs
```diff
@@ -20,6 +21,147 @@ pub struct ServerInfoConfig {
...
+    /// Background theme this icon is designed for. Omit to let the client use
+    /// it with any theme
+    #[serde(default)]
+    pub theme: Option<IconThemeConfig>,
+}
+
+/// Reject `src` values whose scheme the MCP spec instructs clients to refuse
+/// (`javascript:`, `file:`, `ftp:`, `ws:`, and anything else that isn't
+/// `https://` or `data:`), and reject well-schemed but empty inputs like a
+/// bare `https://` or a `data:` URI without an `image/*` payload.
```
gocamille: The `javascript:`/`file:`/`ftp:`/`ws:` list is quoted accurately. "And anything else that isn't `https://` or `data:`" is ours, not the spec's -- worth removing that half.

crates/apollo-mcp-server/src/server_info.rs
```diff
@@ -20,6 +21,147 @@ pub struct ServerInfoConfig {
...
+        return false;
+    }
+    let mime = metadata.split(';').next().unwrap_or("");
+    match mime.strip_prefix("image/") {
+        Some(subtype) => !subtype.is_empty(),
+        None => false,
+    }
+}
+
+fn deserialize_icon_sizes<'de, D>(deserializer: D) -> Result<Option<Vec<String>>, D::Error>
```
gocamille: `WxH` is a formatting example in the schema's prose, not a constraint -- the field is `string[]` with no pattern. `48X48` would fail here on the uppercase `X`. Drop the deserializer and let `sizes` be `Option<Vec<String>>`.

docs/source/config-file.mdx
```diff
@@ -559,8 +560,23 @@ server_info:
...
 ```
 
+Each icon entry supports the fields defined by the [MCP `Icon` object](https://modelcontextprotocol.io/specification/2025-11-25/schema#icon):
+
+| Option      | Type       | Description                                                                                   |
+| :---------- | :--------- | :-------------------------------------------------------------------------------------------- |
+| `src`       | `string`   | URI of the icon: an `https://` URL or a base64-encoded `data:` URI. Required.                 |
+| `mime_type` | `string`   | Explicit MIME type such as `image/png` or `image/svg+xml`. Optional.                          |
+| `sizes`     | `string[]` | Available sizes, each `WxH` (e.g. `48x48`) or `any` for scalable formats. Optional.           |
+| `theme`     | `string`   | Background theme the icon is designed for: `light` or `dark`. Omit this field to render with either theme. Optional. |
```
gocamille, line 576: Three things here once the scheme rule goes: 
- The `src` row says `https://` directly under the icon spec page that says HTTP is allowed, so it just wants "a parseable URI" here. 
- The `sizes` row reads as enforced and won't be -- it should describe `WxH`/`any` as what clients expect, not what we check.
- Good to add a sentence here that the key is `mime_type`, not `mimeType`; with `deny_unknown_fields`, pasting from the linked spec will error out.

Also, nested config types in this docfile get their own heading, so this will need `#### Icon` as a header.

gocamille (changes requested):
Thanks again for putting this together @alexhof-shyftplan. The overall direction is right: custom icons belong under `server_info` and leaving them unset by default is the correct behaviour.

The remaining change is to simplify the validation so that it follows the constraints expressed by the MCP `Icon` schema without adding stricter scheme or size rules. I’ve left inline notes with the specific adjustments.

I’ve left inline notes with the specific adjustments. If anything is unclear, please let me know. Once those changes are in place, please re-request review. Thanks again for your contribution!

alexhof-shyftplan:
Thanks @gocamille — the `serverInfo` (running instance) vs `server.json` (package) framing is a much better articulation than the one in my comment, and I've taken it. `icons()` still returns `None` when nothing is configured.

You're right about the validation layer, and the provenance you traced is accurate: the scheme allowlist and the `WxH` rule both came out of bot review rounds, not out of #805. A four-entry copy of the spec's unsafe-scheme list is exactly the "partial boundary that looks like protection" you describe. It would have blocked `javascript:` while waving through anything else a client is also supposed to refuse.

`server_info.rs` now validates exactly what the `Icon` schema constrains, and nothing else:

- **`src`** — `url::Url::parse`, matching `deserialize_auth_servers` in `auth.rs`. The scheme allowlist and the `data:` media-type inspection are gone. It stays a `String` rather than becoming a `Url` for the same reason `servers` does: the value is advertised to clients verbatim, and `Url::parse` normalizes its input. `#[schemars(with = "Url")]` keeps `format: uri` in the generated config schema.
- **`sizes`** — custom deserializer removed; now a plain `Option<Vec<String>>`. Serde's array-of-strings check is the whole job, since the schema attaches no pattern. `WxH` / `any` is documented, not enforced.
- **`mimeType`** — untouched, plain optional string.
- **`theme`** — `IconThemeConfig` unchanged.

One judgement call I'd like your read on. The docs table described `src` as "URI of the icon: an `https://` URL or a base64-encoded `data:` URI", which read as a constraint the server enforced. I changed it to "such as" and added a short Note that the spec makes the client responsible for refusing unsafe sources and sanitizing before rendering. Happy to drop the Note if you'd rather the docs not editorialize about client behavior.

DaleSeo (approved):
Thanks, @alexhof-shyftplan!

gocamille (approved):
thank you @alexhof-shyftplan !

PR #808, opened 2026-08-05
docs: clarify scope enforcement in the auth guide

DaleSeo:
This PR polishes our auth docs and clarifies how scope enforcement works. `scopes` is not just a validation rule. It also tells clients what the server supports through the `401` challenge and Protected Resource Metadata. The server checks the standard `scope` claim first, then falls back to `scp` for identity providers like Okta. Per-operation `required_scopes` is added to the global `scopes` and `scope_mode` checks. It does not replace them, and it only applies to `tools/call`. The `scope` header in the `403` response lists all the scopes required for the operation, not just the ones missing from the token. The previous version of the docs said otherwise.

### DIFF
--- a/docs/source/auth.mdx
+++ b/docs/source/auth.mdx
@@ -83,6 +83,15 @@ When `issuers` is empty (the default), the server skips issuer validation, so yo
 
 Use the `scope_mode` option to control how the MCP Server validates scopes from OAuth tokens.
 
+The `scopes` list has two functions. First, the mode below controls how the server validates it against incoming tokens. Second, the server also sends this same list to clients, in two places:
+
+- the `scope` parameter of `401` [`WWW-Authenticate` challenges](#www-authenticate-header)
+- the `scopes_supported` field of its [Protected Resource Metadata](https://datatracker.ietf.org/doc/html/rfc9728) document
+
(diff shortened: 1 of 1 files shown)


docs/source/auth.mdx
```diff
@@ -141,6 +150,10 @@ Use `scope_mode: disabled` only when downstream services, such as subgraphs, han
 
 </Caution>
 
+### Scope claim: `scope` vs `scp`
+
+Apollo MCP Server reads the standard `scope` claim first. This claim is a space-separated string, defined in [RFC 6749](https://datatracker.ietf.org/doc/html/rfc6749) and [RFC 9068](https://datatracker.ietf.org/doc/html/rfc9068). If a token has no `scope` claim, the server reads `scp` instead. This is a non-standard claim that some IdPs use, for example Okta. IdPs send it as an array, or as a space-separated string. If a token has both claims, `scope` wins.
```
gocamille: Nit: we mention Okta, but for our enterprise partners, would we also want to mention Microsoft Entra ID (formerly known as Azure AD)?

gocamille (approved):
Thank you @DaleSeo ! Really glad to see these clarifications on OAuth go out!

docs/source/auth.mdx
DaleSeo: Thanks @gocamille! Updated. I didn't now they are our enterprise partners. Good to know haha

PR #810, opened 2026-08-15
fix: propagate W3C baggage through MCP telemetry

### LINKED ISSUE #807: Telemetry: W3C baggage is not propagated (only TraceContextPropagator is registered)
### Summary

Apollo MCP Server registers only `TraceContextPropagator`, so W3C Baggage is neither
extracted from incoming requests nor injected into the outgoing request to the router.
Any context an upstream service places in baggage is silently dropped at the MCP hop.

`baggage` does not appear anywhere in the Rust sources as of v1.17.0:

```
$ git clone --depth 1 https://github.com/apollographql/apollo-mcp-server.git
$ grep -rni baggage --include='*.rs' . | wc -l
0
```

The propagator is registered here:

- `crates/apollo-mcp-server/src/runtime/telemetry.rs:390` — `global::set_text_map_propagator(TraceContextPropagator::new());`
- `crates/apollo-mcp-server/src/server/states/telemetry.rs:105` and `:144` — same, for the server states

### Why this matters

Baggage is the standard OpenTelemetry mechanism for carrying request-scoped
attributes (tenant/account ID, request classification, experiment assignment)
across service boundaries so each service can attach them to its own spans.

In a chain like `client → MCP server → Apollo Router → subgraphs`, the MCP server
is the only hop that drops baggage. That breaks it for **everything downstream**,
not just for the MCP server: the router (which does support
`telemetry.exporters.tracing.propagation.baggage`) and all subgraphs receive no
baggage, so an attribute set once at the edge cannot be used to segment or
attribute telemetry anywhere past the MCP server.

Trace context itself is unaffected — spans still join into one trace correctly.
It is specifically the baggage-carried attributes that are lost.

### Reproduction

1. Configure the MCP server with `endpoint` pointing at an Apollo Router and OTLP tracing enabled.
2. Send an MCP `tools/call` request including a W3C baggage header, e.g.
   `baggage: my.tenant.id=acme`, alongside a valid `traceparent`.
3. Inspect the request the MCP server makes to the router, and the exported spans.

Observed: the outgoing router request carries no `baggage` header, and no MCP
server span carries any baggage-derived attribute.

Expected: the incoming baggage is extracted into the active context, injected into
the outgoing router request, and available for span attributes.

### Suggested fix

Register a composite propagator so baggage travels with trace context:

```rust
use opentelemetry::propagation::TextMapCompositePropagator;
use opentelemetry_sdk::propagation::{BaggagePropagator, TraceContextPropagator};

global::set_text_map_propagator(TextMapCompositePropagator::new(vec![
    Box::new(TraceContextPropagator::new()),
    Box::new(BaggagePropagator::new()),
]));
```

This matches what the Apollo Router already does when baggage propagation is
enabled, and what the OpenTelemetry SDK defaults to in most other languages
(`tracecontext,baggage`).

If propagation should stay opt-in, a `telemetry.propagation` config block
(mirroring the router's) would work too.

### Workaround

Adding `baggage` to `forward_headers` restores the header for everything
downstream of the MCP server:

```yaml
forward_headers:
  - baggage
```

That is enough to unblock the router and subgraphs, but it does not let the MCP
server's own spans carry baggage-derived attributes, since the value never enters
its OpenTelemetry context.

### Version

apollo-mcp-server v1.17.0 (also present in v1.16.0)

trippyogi:
## Summary
Fixes #807.

Apollo MCP Server currently registers only the W3C Trace Context propagator. Incoming baggage is therefore not extracted into the OpenTelemetry context and cannot be propagated through the server's tracing spans or injected into downstream GraphQL requests.

This registers W3C Baggage alongside Trace Context and updates the default CORS configuration so browser clients can send the baggage header.

## Changes
- Register a composite OpenTelemetry propagator containing TraceContextPropagator and BaggagePropagator.
- Allow `baggage` in the default CORS request headers.
- Add regression coverage for incoming baggage extraction.
- Add an end-to-end regression covering incoming HTTP baggage → middleware extraction → tracing/OpenTelemetry parent context → outgoing GraphQL injection through reqwest-tracing.
- Update CORS and telemetry documentation.
- Add the corresponding changeset.

Baggage remains request-scoped telemetry context. This change does not treat baggage as trusted identity and does not automatically copy baggage entries onto exported span attributes.

## Test plan
- [x] Focused telemetry / GraphQL / CORS tests pass
- [x] `cargo clippy -- -D warnings` passes for the affected workspace
- [x] `cargo fmt` clean


docs/source/cors.mdx
```diff
@@ -115,15 +115,16 @@ cors:
...
+  # Enable these request headers for cross-origin requests:
+  # Defaults required by MCP, W3C Trace Context, and W3C Baggage
   allow_headers:
     - accept
     - content-type
     - mcp-protocol-version
     - mcp-session-id
     - traceparent # W3C Trace Context
     - tracestate # W3C Trace Context
+    - baggage # W3C Baggage
```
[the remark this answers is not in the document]
trippyogi: Removed `accept` from the example. It now matches `CorsConfig::default()` (no accept).

crates/apollo-mcp-server/src/graphql.rs
```diff
@@ -294,6 +294,96 @@ mod test {
...
+                        }
+                    }
+                }),
+            )
+            .layer(axum::middleware::from_fn(otel_context_middleware));
+
+        let request = HttpRequest::builder()
+            .method("POST")
+            .uri("/mcp")
+            .header("baggage", "userId=alice")
```
gocamille, line 378: An incoming `baggage: userId=alice;property=one%0Atwo` currently panics when the context is injected into the downstream GraphQL request. OTel decodes the newline, then `reqwest-tracing` hits `HeaderValue::from_str(...).expect(...)` with `InvalidHeaderValue`. Since this is caller-controlled could we drop baggage that can’t be safely re-injected and add this case to the end-to-end test?

crates/apollo-mcp-server/src/runtime/telemetry.rs
```diff
@@ -361,6 +362,14 @@ fn init_tracer_provider(telemetry: &Telemetry) -> Result<SdkTracerProvider, anyh
     Ok(tracer_provider)
 }
 
+/// Composite propagator for W3C Trace Context and W3C Baggage.
+fn w3c_text_map_propagator() -> TextMapCompositePropagator {
+    TextMapCompositePropagator::new(vec![
+        Box::new(TraceContextPropagator::new()),
+        Box::new(BaggagePropagator::new()),
```
gocamille: I noticed another W3C parsing case here where `baggage: userId=ali=ce` currently extracts as `userId=ali`. The spec explicitly allows any number of `=` characters in a value. This looks like it comes from the SDK parser but we should preserve the full value and add a regression before describing this as W3C Baggage propagation.

docs/source/telemetry.mdx
```diff
@@ -164,7 +164,7 @@ Spans are generated for the following actions:
 - **Incoming HTTP Requests**: A root span is created for every HTTP request to the MCP server.
 - **MCP Handler Methods**: Nested spans are created for each of the main MCP protocol methods (`initialize`, `call_tool`, `list_tools`).
 - **Tool Execution**: `call_tool` spans contain nested spans for the specific tool being executed (e.g., `introspect`, `search`, or a custom GraphQL operation).
-- **Downstream GraphQL Calls**: The `execute` tool and custom operation tools create child spans for their outgoing `reqwest` HTTP calls, capturing the duration of the downstream request. The `traceparent` and `tracestate` headers are propagated automatically, enabling distributed traces.
+- **Downstream GraphQL Calls**: The `execute` tool in Apollo MCP Server and custom operation tools create child spans for their outgoing `reqwest` HTTP calls, capturing the duration of the downstream request. The `traceparent`, `tracestate`, and `baggage` headers propagate automatically. Baggage is request-scoped context, not identity. Apollo MCP Server doesn't automatically copy baggage onto exported span attributes.
```
gocamille: “Not identity” is good. Could we also say baggage is caller-controlled, shouldn’t contain secrets or PII or be used for authorization, and should be filtered when the GraphQL endpoint crosses a trust boundary?

crates/apollo-mcp-server/src/server/states/telemetry.rs
```diff
@@ -23,6 +27,15 @@ impl<'a> Extractor for HeaderExtractor<'a> {
     }
 }
 
+/// Composite propagator for W3C Trace Context and W3C Baggage.
```
gocamille, line 121: This test helper duplicates the production propagator constructor in `runtime/telemetry.rs` so the end-to-end test could drift from what the server actually registers. Could we put the constructor in one shared module and use it from both?

crates/apollo-mcp-server/src/server/states/telemetry.rs
```diff
@@ -156,6 +166,68 @@ mod tests {
...
+
+        let app = Router::new()
+            .route("/test", get(test_handler))
+            .layer(axum::middleware::from_fn(otel_context_middleware));
+
+        let traceparent = "00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01";
+        let request = Request::builder()
+            .uri("/test")
+            .header("traceparent", traceparent)
+            .header("baggage", "userId=alice,serverNode=DF28")
```
gocamille, line 290: This covers the comma-joined form but W3C also allows the same context to be split across multiple `baggage` headers. With `HeaderExtractor` currently using `HeaderMap::get` the second header is dropped. Could we add that case here and combine `get_all("baggage")` before extraction?

gocamille (changes requested):
Thanks for this @trippyogi! The overall shape looks right. I found two things I think need fixing before merge, which are that percent-encoded baggage metadata can currently `panic` the downstream GraphQL request and baggage split across multiple header fields is partially dropped. There’s also a valid `=` value case that gets truncated by the current SDK parser.

Could we cover those three inputs, update to `opentelemetry_sdk` 0.32.1 and tighten the trust-boundary note in the docs? I ran the full server tests, clippy and formatting; everything else passed cleanly.

crates/apollo-mcp-server/src/graphql.rs
trippyogi: Fixed. Added a separate end-to-end regression for the %0A case that verifies the unsafe member is dropped, the safe member still propagates downstream, and the GraphQL request completes without panicking.

crates/apollo-mcp-server/src/runtime/telemetry.rs
trippyogi: Fixed: extra raw = characters in baggage values are preserved through extraction/re-injection, with a regression for userId=ali=ce.

docs/source/telemetry.mdx
trippyogi: Updated: the docs now call out that baggage is caller-controlled, shouldn't contain secrets/PII or be used for authorization, and should be filtered across trust boundaries.

crates/apollo-mcp-server/src/server/states/telemetry.rs
trippyogi: Fixed: production and tests now use the same shared w3c_text_map_propagator() constructor.

crates/apollo-mcp-server/src/server/states/telemetry.rs
trippyogi: Fixed: HeaderExtractor now combines get_all("baggage") before extraction, with regression coverage using multiple baggage header fields.

trippyogi:
@gocamille Thanks for the review! I pushed a follow-up covering all five threads plus the opentelemetry_sdk 0.32.1 bump. The three baggage edge cases now have regressions as well. Ready for another pass when you have a chance.

gocamille (approved):
Thank you @trippyogi ! I am going to reach out to my docs team for them to give a quick eye on the doc updates, but the implementation looks sound.

docs/source/telemetry.mdx
```diff
@@ -164,7 +164,7 @@ Spans are generated for the following actions:
...
 - **Tool Execution**: `call_tool` spans contain nested spans for the specific tool being executed (e.g., `introspect`, `search`, or a custom GraphQL operation).
-- **Downstream GraphQL Calls**: The `execute` tool and custom operation tools create child spans for their outgoing `reqwest` HTTP calls, capturing the duration of the downstream request. The `traceparent` and `tracestate` headers are propagated automatically, enabling distributed traces.
+- **Downstream GraphQL Calls**: The `execute` tool in Apollo MCP Server and custom operation tools create child spans for their outgoing `reqwest` HTTP calls, capturing the duration of the downstream request. The `traceparent`, `tracestate`, and `baggage` headers propagate automatically. Baggage is caller-controlled, request-scoped context, not identity. Do not put secrets or personally identifiable information in baggage, and never use it for authorization. If your GraphQL endpoint crosses a trust boundary, filter baggage before forwarding it. Apollo MCP Server doesn't automatically copy baggage onto exported span attributes.
```
pragl:
```suggestion
- **Downstream GraphQL Calls**: The `execute` tool in Apollo MCP Server and custom operation tools create child spans for their outgoing `reqwest` HTTP calls, capturing the duration of the downstream request. The `traceparent`, `tracestate`, and `baggage` headers propagate automatically. Baggage is caller-controlled, request-scoped context, not identity. Don't put secrets or personally identifiable information in baggage, and never use it for authorization. If your GraphQL endpoint crosses a trust boundary, filter baggage before forwarding it. Apollo MCP Server doesn't automatically copy baggage onto exported span attributes.
```

pragl (approved):
The tiniest of copyedits. Docs LGTM!

gocamille:
Thank you again @trippyogi! Merging now—I’m also looking into a fast-follow to add configurable baggage controls for allowing all keys, allowlisting specific keys, or disabling propagation 👍

PR #814, opened 2026-08-26
fix: disable ANSI styling for non-terminal logs

### LINKED ISSUE #813: Disable ANSI styling in non-terminal log output and honor NO_COLOR
## Summary

Apollo MCP Server emits ANSI SGR escape sequences in stdout logs even when
stdout is not connected to a terminal, such as when the server runs in a
container whose output is collected by a logging pipeline.

There is currently no configuration option to request plain-text stdout logs.

## Version tested

I reproduced this from current `main` at commit
`5b5a079cabc266ca9be9818c94ad09cb45742061`, which reports Apollo MCP Server
version `1.17.0`.

## Reproduction

I built the repository's Dockerfile:

```console
docker build -t apollo-mcp-server:ansi-repro .
```

I then started the example server without allocating a TTY and redirected
stdout to a binary file:

```console
docker run --rm --name apollo-mcp-ansi-repro \
  -v "$PWD/examples:/data/examples:ro" \
  apollo-mcp-server:ansi-repro \
  examples/weather/config.yaml > out.bin
```

After startup, I stopped the container from another terminal and counted the
escape-sequence introducers:

```console
docker stop apollo-mcp-ansi-repro
python3 -c "print(open('out.bin','rb').read().count(b'\x1b['))"
```

Observed result:

```text
48
```

The beginning of the captured output, represented as bytes, was:

```text
b'\x1b[2m2026-08-26T13:57:53.519718Z\x1b[0m \x1b[32m INFO\x1b[0m Apollo MCP Server v1.17.0 ...'
```

The timestamp and severity token are therefore wrapped in ANSI escape
sequences even though stdout is a pipe rather than a terminal.

I repeated the same non-TTY run with conventional color controls. None changed
the result:

| Environment | Escape-sequence count |
|---|---:|
| Baseline | 48 |
| `NO_COLOR=1` | 48 |
| `RUST_LOG_STYLE=never` | 48 |
| `CLICOLOR=0` | 48 |
| `TERM=dumb` | 48 |

The current `logging` configuration supports `level`, `path`, and `rotation`,
but does not expose a plain-text or structured/JSON output setting.

## Production impact

Container log collection pipelines commonly parse the severity token to
classify records for retention and sampling. Parsers expecting a plain
`INFO`, `WARN`, or `ERROR` token may fail to classify a token surrounded by
ANSI bytes. Such records can fall back to low-priority handling and be sampled
even when they contain startup or error information that should be retained.

## Expected behavior

By default, ANSI styling should only be emitted when the output stream is
connected to a terminal. Setting [`NO_COLOR`](https://no-color.org/) should
also disable styling.

## Proposed approach

I propose a small, focused change to:

1. Use `std::io::IsTerminal` when configuring stdout logging, enabling ANSI
   only when stdout is a terminal.
2. Disable ANSI when `NO_COLOR` is set.
3. Preserve the existing plain output for configured log files.
4. Add regression tests covering:
   - ANSI enabled for terminal output.
   - ANSI disabled for non-terminal output.
   - ANSI disabled when `NO_COLOR` is set.

An explicit configuration override could allow operators to force styling
either on or off, but I would keep that out of the initial bug fix unless the
maintainers prefer it in scope.

Structured/JSON logging is a related gap, but it is a larger feature and I
would not bundle it into this fix without agreement.

Does this scope and implementation approach look acceptable? Would you prefer
an explicit logging override in the same change, or should that and structured
logging be tracked separately?

Staticsubh:
## Summary

- enable ANSI log styling only when stdout or stderr is attached to a terminal
- honor a non-empty `NO_COLOR` environment variable
- document the behavior and include the required patch changeset

Fixes #813

## Byte-level verification

Tested against current `main` at `5b5a079cabc266ca9be9818c94ad09cb45742061` (package version `1.17.0`).

Before this change:

```text
non-TTY stdout:                48 occurrences of b'\x1b['
non-TTY stdout + NO_COLOR=1:   48 occurrences of b'\x1b['
```

With this change:

```text
non-TTY stdout:                 0 occurrences of b'\x1b['
TTY stdout:                    48 occurrences of b'\x1b['
TTY stdout + NO_COLOR=1:        0 occurrences of b'\x1b['
```

The counts were obtained from captured output using:

```shell
python3 -c "print(open('out.bin','rb').read().count(b'\x1b['))"
```

This preserves styled output for interactive terminals while making redirected output suitable for container log collection pipelines.

## Testing

- `cargo test`
- `cargo clippy --all-targets -- --deny warnings`
- `cargo fmt`
- four focused ANSI decision tests
- runtime byte capture with non-TTY, TTY, and TTY plus `NO_COLOR=1`

### DIFF
--- a/crates/apollo-mcp-server/src/runtime/logging.rs
+++ b/crates/apollo-mcp-server/src/runtime/logging.rs
@@ -11,6 +11,8 @@ mod trace_id_format;
 use log_rotation_kind::LogRotationKind;
 use schemars::JsonSchema;
 use serde::Deserialize;
+use std::ffi::OsStr;
+use std::io::IsTerminal;
 use std::path::PathBuf;
 use tracing::Level;
 use tracing_appender::rolling::RollingFileAppender;
@@ -73,6 +75,7 @@ impl Logging {
     }
 
     pub fn logging_layer(logging: &Logging) -> Result<LoggingLayerResult, anyhow::Error> {
+        let no_color = std::env::var_os("NO_COLOR");
         let (writer, guard, with_ansi) = match logging.path.clone() {
             Some(path) => std::fs::create_dir_all(&path)
                 .map(|_| path)
@@ -97,9 +100,17 @@ impl Logging {
                 })
                 .unwrap_or_else(|| {
                     eprintln!("Log file setup failed - falling back to stderr");
-                    (BoxMakeWriter::new(std::io::stderr), None, true)
+                    (
+                        BoxMakeWriter::new(std::io::stderr),
+                        None,
+                        should_use_ansi(std::io::stderr().is_terminal(), no_color.as_deref()),
+                    )
                 }),
-            None => (BoxMakeWriter::new(std::io::stdout), None, true),
+            None => (
+                BoxMakeWriter::new(std::io::stdout),
+                None,
+                should_use_ansi(std::io::stdout().is_terminal(), no_color.as_deref()),
+            ),
         };
 
         let inner_format = tracing_subscriber::fmt::format::Format::default()
@@ -114,6 +125,10 @@ impl Logging {
             guard,
         ))
     }
+}
+
+fn should_use_ansi(is_terminal: bool, no_color: Option<&OsStr>) -> bool {
+    is_terminal && no_color.is_none_or(OsStr::is_empty)
 }
 
 fn level(generator: &mut schemars::SchemaGenerator) -> schemars::Schema {
@@ -132,4 +147,30 @@ fn level(generator: &mut schemars::SchemaGenerator) -> schemars::Schema {
     }
 
     Level::json_schema(generator)
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::ffi::OsStr;
+
+    #[test]
+    fn ansi_is_enabled_for_terminal_output() {
+        assert!(should_use_ansi(true, None));
+    }
+
+    #[test]
+    fn ansi_is_disabled_for_non_terminal_output() {
+        assert!(!should_use_ansi(false, None));
+    }
+
+    #[test]
+    fn ansi_is_disabled_when_no_color_is_set() {
+        assert!(!should_use_ansi(true, Some(OsStr::new("1"))));
+    }
+
+    #[test]
+    fn ansi_is_enabled_when_no_color_is_empty() {
+        assert!(should_use_ansi(true, Some(OsStr::new(""))));
+    }
 }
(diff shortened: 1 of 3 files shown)


Staticsubh:
Thanks @coderabbitai for the review! Both findings were valid and are fixed in `86ad681`:

- The changeset claimed stderr fallback output always uses plain text, but `logging.rs` (lines ~101-107) enables ANSI for it too when stderr is a terminal and `NO_COLOR` is unset/empty, same as stdout. Dropped the incorrect qualifier.
- `config-file.mdx` linked directly from the `NO_COLOR` variable name. Kept `NO_COLOR` in code font and moved the link to a descriptive "NO_COLOR website" phrase at the end of the sentence, matching the project's link style.

Local validation (pinned `1.92.0` toolchain, `AWS_LC_SYS_CMAKE_BUILDER=1`):

- `cargo fmt --check` passed.
- `cargo test -p apollo-mcp-server --bin apollo-mcp-server` passed: 88 passed, 0 failed (includes the 4 ANSI decision tests).
- `cargo test -p apollo-mcp-server --lib` passed: 613 passed, 0 failed related to this change (same pre-existing, unrelated snapshot/network failures noted in other PRs in this repo).
- `cargo clippy -p apollo-mcp-server --all-targets -- -D warnings` passed clean.

CI is green on everything that runs on a fork PR (Build, Tests, Clippy Linux/Windows, Cargo Format/Doc, Test Coverage, codecov, CodeRabbit, AI Style Review, the Claude review workflow). `Verify` and `End-to-End Tests` fail on the same fork-PR permission restrictions covered in earlier PRs (a bot-comment step without write access, and no OIDC token injection for `google-github-actions/auth` on fork-triggered workflows) — unrelated to this diff.

crates/apollo-mcp-server/src/runtime/logging.rs
```diff
@@ -116,6 +127,10 @@ impl Logging {
     }
 }
 
+fn should_use_ansi(is_terminal: bool, no_color: Option<&OsStr>) -> bool {
+    is_terminal && no_color.is_none_or(OsStr::is_empty)
+}
+
```
DaleSeo, line 133: GitHub Actions capture through a pipe yet display color. Was a three-way `auto | always | never` option considered, either as a `logging.ansi` config key or by honoring `CLICOLOR_FORCE` and `FORCE_COLOR` alongside `NO_COLOR`?

crates/apollo-mcp-server/src/runtime/logging.rs
```diff
@@ -133,3 +148,29 @@ fn level(generator: &mut schemars::SchemaGenerator) -> schemars::Schema {
 
     Level::json_schema(generator)
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::ffi::OsStr;
```
DaleSeo, line 155: This is already pulled in from the parent module.

```suggestion
```

PR #816, opened 2026-08-29
feat: configure which requests skip OAuth token validation

DaleSeo:
`allow_anonymous_mcp_discovery` only supported a fixed set of three discovery methods. Deployments need more flexibility, including public tools that can be reached without a token and a second credential, such as an API key validated by an upstream API, that can coexist with OAuth.

This PR replaces `allow_anonymous_mcp_discovery` with `skip_token_validation`. It is configurable through three lists in a new `SkipTokenValidation` config struct. Each list matches a different part of the request.

- `methods` matches the JSON-RPC method name. `tools/call` is rejected here during deserialization because every tool call uses the same method name. Individual tools should be listed under `tools` instead.
- `tools` matches the tool name in a `tools/call` request.
- `headers` matches HTTP header names. This list does not need the request body, so it is checked before the body is buffered. It also covers non-POST requests, such as the GET stream.

Each list only applies when the request has no `Authorization` header. Requests with a token are always validated, so an expired token is still rejected even if the request uses a listed method, tool, or header.

`allow_anonymous_mcp_discovery` is deprecated but still supported. At startup, its values are added to `skip_token_validation.methods` as `initialize`, `tools/list`, and `resources/list`, and a deprecation warning is logged. Setting both `allow_anonymous_mcp_discovery` and `skip_token_validation.methods` causes a startup error because they would define the same list twice.

### DIFF
--- a/crates/apollo-mcp-server/src/auth.rs
+++ b/crates/apollo-mcp-server/src/auth.rs
@@ -24,6 +24,7 @@ use tower_http::cors::{Any, CorsLayer};
 use tracing::warn;
 use url::Url;
 
+use crate::apps::app_param_from_query;
 use crate::scope_requirements::OperationRequiredScopes;
 
 mod networked_key_resolver;
@@ -94,6 +95,12 @@ pub enum TlsConfigError {
     ServerUrlMissingHost { index: usize, url: String },
     #[error("Resource URL has non-HTTP(S) scheme '{scheme}': {url}")]
     ResourceUrlInvalidScheme { url: String, scheme: String },
+    #[error(
+        "transport.auth sets both `allow_anonymous_mcp_discovery` and \
+         `skip_token_validation.methods`. `allow_anonymous_mcp_discovery` is deprecated: remove it \
+         and list the methods you want in `skip_token_validation.methods`."
+    )]
+    AnonymousDiscoveryConflict,
 }
 
 impl TlsConfig {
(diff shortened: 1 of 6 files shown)


crates/apollo-mcp-server/src/auth.rs
```diff
@@ -522,11 +656,23 @@ async fn oauth_validate(
         )
     };
 
+    // Every skip list applies only to a request that presents no token. A caller
+    // that sends one always gets it validated, so an expired token is rejected
+    // here instead of passing as an anonymous request.
+    let skip = &auth_state.skip_token_validation;
+
+    // The header list needs no body, so it answers before the body is buffered.
+    if token.is_none() && skip.matches_headers(request.headers()) {
```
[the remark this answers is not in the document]
DaleSeo: `is_missing` is only true when the `Authorization` header is completely absent. If the header is present but uses the wrong scheme, like `Authorization: Basic ...`, it fails to decode as `Authorization<Bearer>`. In that case, `is_missing` is `false`, so this returns `Err(...)` rather than `Ok(None)`.

Because `token` is extracted as a handler parameter in `oauth_validate` through `from_fn_with_state`, that `Err` short-circuits the request and becomes the extractor's rejection response. The function body never runs.

crates/apollo-mcp-server/src/auth.rs
```diff
@@ -313,9 +419,35 @@ struct AuthState {
...
+    /// Skip lists resolved at startup, with the deprecated
+    /// `allow_anonymous_mcp_discovery` flag already folded in.
+    skip_token_validation: Arc<SkipTokenValidation>,
 }
 
 impl Config {
+    /// Folds the deprecated `allow_anonymous_mcp_discovery` flag into
+    /// `skip_token_validation`, so the request path reads a single list.
+    #[allow(deprecated)]
+    fn resolve_skip_token_validation(&self) -> Result<SkipTokenValidation, TlsConfigError> {
```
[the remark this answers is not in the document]
DaleSeo: The runtime/config.rs file validates the document structure, such as rejecting auth under a stdio transport, but it does not resolve the semantics inside a module's own config struct.

crates/apollo-mcp-server/src/auth.rs
```diff
@@ -94,6 +94,12 @@ pub enum TlsConfigError {
     ServerUrlMissingHost { index: usize, url: String },
     #[error("Resource URL has non-HTTP(S) scheme '{scheme}': {url}")]
     ResourceUrlInvalidScheme { url: String, scheme: String },
+    #[error(
+        "transport.auth sets both `allow_anonymous_mcp_discovery` and \
+         `skip_token_validation.methods`. `allow_anonymous_mcp_discovery` is deprecated: remove it \
+         and list the methods you want in `skip_token_validation.methods`."
+    )]
+    AnonymousDiscoveryConflict,
```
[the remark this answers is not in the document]
DaleSeo: This variant follows the existing naming convention.

crates/apollo-mcp-server/src/auth.rs
```diff
@@ -1780,6 +1932,371 @@ emyPxgcYxn/eR44/KJ4EBs+lVDR3veyJm+kXQ99b21/+jh5Xos1AnX5iItreGCc=
...
+                    all_lists().skip_token_validation.headers,
+                    vec![HeaderName::from_static("x-api-key")]
+                );
+            }
+
+            #[test]
+            fn the_lists_default_to_empty() {
+                let lists = parse("").unwrap().skip_token_validation;
+                assert!(
+                    !lists.needs_body() && lists.headers.is_empty(),
```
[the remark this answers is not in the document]
DaleSeo: If `needs_body` ever gets a condition, the request-path test is the one that would catch that behavioral change. So I'm keeping the parse-level assertion as is instead of splitting it into three field-level tests.

crates/apollo-mcp-server/src/auth.rs
```diff
@@ -162,6 +168,131 @@ where
...
+        if self.methods.iter().any(|method| method == &peek.method) {
+            return true;
+        }
+        if peek.method != TOOL_CALL_METHOD {
+            return false;
+        }
+        peek.params
+            .as_ref()
+            .and_then(|params| params.name.as_deref())
+            .is_some_and(|tool| self.tools.iter().any(|allowed| allowed == tool))
```
gocamille, line 303: I traced a `tools/call` with `?app=MyApp` through the auth check and it looks like this check reads only `params.name`, so `tools: [Hello]` allows the request. Later, `Running::call_tool_impl` reads `?app=` and dispatches `MyApp` plus `Hello`.

The existing `calls_app_tool_instead_of_operation_when_app_param_present` test demo's that difference. The app tool can execute `AppHello`, not the ordinary `Hello` operation. So auth identifies the target as `Hello`, while dispatch identifies it as `MyApp + Hello`. Should we key this policy by app plus tool name? Another safe option may be to reject app-qualified calls from this list.

crates/apollo-mcp-server/src/auth.rs
```diff
@@ -522,14 +721,43 @@ async fn oauth_validate(
...
     };
 
+    // Every skip list applies only to a request that presents no token. A caller
+    // that sends one always gets it validated, so an expired token is rejected
+    // here instead of passing as an anonymous request.
+    let skip = &auth_state.skip_token_validation;
+
+    // The header list needs no body, so it answers before the body is buffered.
+    if token.is_none() && skip.matches_headers(request.headers()) {
+        let response = next.run(request).await;
```
gocamille, line 742: While following this early return into the next layer (`listed header > next.run(request) > StreamableHttpService > rmcp body.collect().await`) this branch returns before `extract_body` -- if I'm reading this correctly, then a tokenless caller can supply any value and send a very large body. Should we apply a streaming request-body limit before this branch?

docs/source/auth.mdx
```diff
@@ -333,29 +333,115 @@ These headers are sent with every request the server makes to discover authoriza
...
+# Without it, the header is dropped and nothing downstream can validate it.
+forward_headers:
+  - x-api-key
 transport:
   type: streamable_http
   auth:
     servers:
       - https://auth.example.com
-    allow_anonymous_mcp_discovery: true
+    resource: https://mcp.example.com
```
gocamille: I compared this example with the config reference and the Auth0 guide and both identify the resource as `https://mcp.example.com/mcp`. The MCP auth spec also seems to recommend using the most specific available URI. Since this server is mounted at `/mcp`, can we use `https://mcp.example.com/mcp` here for consistency?

crates/apollo-mcp-server/src/auth.rs
```diff
@@ -162,6 +168,131 @@ where
...
+#[derive(Debug, Clone, Default, Deserialize, JsonSchema)]
+#[serde(deny_unknown_fields)]
+pub struct SkipTokenValidation {
+    /// JSON-RPC method names, such as `tools/list`.
+    ///
+    /// `tools/call` is rejected here, because every tool call uses that one
+    /// method name. Name the tools in `tools` instead. Other methods that name
+    /// one of many items through a request parameter, such as `resources/read`
+    /// and `prompts/get`, are rejected for the same reason.
+    #[serde(default, deserialize_with = "deserialize_skip_methods")]
```
gocamille, line 250: I compared the custom serde validation with the generated config schema. It looks like serde rejects `tools/call`, the item-selecting methods, invalid header names, and auth. The generated schema sees these fields as `Vec<String>` so a schema validator can accept a config that startup rejects.

Is that mismatch intentional? If not, should we express these exclusions through a custom schema?

docs/source/auth.mdx
```diff
@@ -333,29 +333,115 @@ These headers are sent with every request the server makes to discover authoriza
...
+
+### The header list moves authentication, it does not remove it
+
+A request that carries a listed header skips token validation, and a later layer authenticates it. Apollo MCP Server cannot judge a credential it does not understand, so it only reads whether the header is present.
+
+A listed header is checked before the JSON-RPC body, so a match skips validation for every method and tool, including ones absent from `methods` and `tools`. `headers` is not scoped down by the other two lists.
+
+Make sure something downstream validates that header. If nothing does, the header is an open door: any caller can send it with any value.
+
+That validator only gates requests that reach your API. Requests the MCP server answers itself, such as the `initialize` handshake, `notifications/initialized`, `tools/list`, `resources/read`, and the `GET` stream, never reach your API, so nothing validates the header value on them even with a correctly configured downstream validator. A caller that sends the header with any value gets a working session, the full tool list, and your schema. Downstream validation gates tool execution, not discovery.
```
gocamille, line 396: This line is helpful! I followed what happens after the header matches and want to confirm the intended security model, to make sure I understand.

The OAuth middleware checks only that the header exists. It does not validate the header value.
For a GraphQL-backed tool, the server forwards the header. The GraphQL API can then validate its value.
Some requests never reach the GraphQL API. The MCP server handles them directly. These include initialization, discovery, resources, streams, session deletion, and local tools.

For those requests, nothing downstream validates the header. A caller can send any value and still skip OAuth.

Am I reading that correctly? If so, what is intended to protect those requests?
- Does a trusted gateway validate the header before the request reaches this server?
- Does this server need to validate the alternate credential?
- Does header-based skipping need to apply only to GraphQL-backed calls?
- 
If these local actions are intentionally public, should we state that directly in the documentation?

crates/apollo-mcp-server/src/auth.rs
DaleSeo: Good catch, @gocamille. I made an `?app=` qualified `tools/call` never match the tools list, so those calls always require a token.

docs/source/auth.mdx
DaleSeo: Switched to `https://mcp.example.com/mcp`. Also updated the resource_metadata example.

crates/apollo-mcp-server/src/auth.rs
DaleSeo: This is resolved now. I rebased onto main to pick up the rmcp 3.2 upgrade. It reads the body frame by frame against a cap, with` max_request_body_bytes` defaulting to 4 MiB.

crates/apollo-mcp-server/src/auth.rs
DaleSeo: `schemars::schema_for!(Config)` only appears in a test, and we do not publish a schema artifact. So the loop currently does not disagree with startup. I would rather not hand-write custom schemas for these types, since the rejected value set and header-name syntax are awkward to express. We would also need to keep them manually synced with the serde implementations.

docs/source/auth.mdx
DaleSeo: Your reading is correct, and you are right that I left DELETE off the list. Those local actions are intentionally public. There is no gateway in front of them, the upstream API handles validation, and the discovery and docs tools are supposed to work without an account. I will make that clear in the docs instead of leaving it implied. Restricting header skipping to GraphQL-backed calls would be tighter, but it would break the handshake and GET stream for the exact callers this list is meant to support.

gocamille: This is really good context, thank you @DaleSeo !

gocamille (approved):

PR #817, opened 2026-09-02
fix: annotate built-in introspection tools

### LINKED ISSUE #793: Built-in introspection tools (execute/introspect/search/validate) ship without tool annotations, so clients can't auto-allow them
### Problem

Operation tools get `ToolAnnotations` — `readOnlyHint` since #103, `destructiveHint` since #510, per-operation `overrides.annotations` since #726 (v1.13.0) — but the four built-in introspection tools (`execute`, `introspect`, `search`, `validate`) are constructed with a bare `Tool::new(...)` and carry **no annotations at all** (checked on `main` / v1.16.0, e.g. `crates/apollo-mcp-server/src/introspection/tools/execute.rs`).

`overrides.annotations` can't reach them either: the override map is applied only in `operations/operation.rs`, keyed by operation name, so a config entry like `execute:` never matches the built-in tools.

### Impact

MCP clients that gate auto-approval on `readOnlyHint: true` (Claude Desktop's tool-permission grouping, VS Code Copilot's confirmation dialogs) prompt the user on **every** call to the built-in tools — even when the upstream API is query-only and `overrides.mutation_mode: none` makes them read-only in practice. The curated operation tools on the same server are auto-allowed, so the two tool layers behave inconsistently.

### Request

Annotate the built-in tools to match the operation defaults:

- `introspect` / `search` / `validate`: always `read_only_hint: true`, `destructive_hint: false`, `idempotent_hint: true`, `open_world_hint: true` — they only read the local SDL / validate against it.
- `execute`: `read_only_hint` / `destructive_hint` should follow `mutation_mode` (read-only when `none`), analogous to how operations derive them from the operation type.

Alternatively (or additionally): allow `overrides.annotations` entries to target the built-in tool names.

### Workaround we use today

A small stdio shim wrapping the binary that rewrites `tools/list` responses to add the annotations to the four built-in tools — it works, but it's exactly the kind of glue this config should make unnecessary.

trippyogi:
Fixes #793.

The built-in execute, introspect, search, and validate tools currently omit MCP tool annotations, so clients that use readOnlyHint for approval decisions cannot distinguish the read-only tools.

This adds annotations to the built-ins:

- introspect, search, and validate are read-only, non-destructive, idempotent, and closed-world.
- execute is read-only unless mutation_mode: all allows ad hoc mutations, and remains open-world because it calls the configured GraphQL endpoint.

Also adds coverage at the tools/list boundary.

### Testing

- cargo test -p apollo-mcp-server
- cargo clippy --all-targets -- --deny warnings
- cargo fmt --check

### DIFF
--- a/.changeset/annotate_builtin_introspection_tools.md
+++ b/.changeset/annotate_builtin_introspection_tools.md
@@ -1,0 +1,7 @@
+---
+default: patch
+---
+
+# Annotate built-in introspection tools
+
+The built-in `introspect`, `search`, `validate`, and `execute` tools now expose MCP `ToolAnnotations`. Schema-only tools are marked read-only, non-destructive, and idempotent. `execute` is read-only unless `mutation_mode` is `all`, and is always marked open-world because it calls the configured GraphQL endpoint.
--- a/crates/apollo-mcp-server/src/introspection/tools.rs
+++ b/crates/apollo-mcp-server/src/introspection/tools.rs
@@ -5,3 +5,22 @@ pub(crate) mod execute;
 pub(crate) mod introspect;
 pub(crate) mod search;
 pub(crate) mod validate;
+
+use rmcp::model::{Tool, ToolAnnotations};
+
+/// Annotations for built-in tools that only read in-process schema state.
+///
+/// MCP: `readOnlyHint` means the tool does not modify its environment;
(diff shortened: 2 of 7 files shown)


DaleSeo (approved):
Thanks @trippyogi!

PR #818, opened 2026-09-02
feat: support the server/discover RPC

inanna-apollo:
Clients can call `server/discover` before any other request to select a compatible protocol version, including as the opening message on stdio where it acts as a backward-compatibility probe.

rmcp 3.1.4 routes the request and builds the response from the same `get_info()` that backs `initialize`, so the discovery response carries matching capabilities and the `server_info` metadata by construction. Adds tests pinning that, and allows `server/discover` without authentication when `allow_anonymous_mcp_discovery` is enabled, since it precedes `initialize` and clients cannot authenticate before calling it.

The supported-version list stays capped at 2025-11-25. Advertising 2026-07-28 would also commit `initialize` to negotiating it, and that revision replaces the standalone notification stream with `subscriptions/listen`, which is not implemented here — a client would read the advertised `tools.listChanged` capability and never receive those notifications.

### DIFF
--- a/crates/apollo-mcp-server/src/auth.rs
+++ b/crates/apollo-mcp-server/src/auth.rs
@@ -1880,8 +1880,8 @@ emyPxgcYxn/eR44/KJ4EBs+lVDR3veyJm+kXQ99b21/+jh5Xos1AnX5iItreGCc=
 
         #[tokio::test]
         async fn server_discover_without_token_allowed_when_enabled() {
-            // `server/discover` precedes `initialize` in the inline lifecycle,
-            // so a client has no way to authenticate before calling it.
+            // `server/discover` may be a client's first MCP request, so this
+            // option permits it anonymously when anonymous discovery is enabled.
             let app = discovery_router(true);
             let req = Request::builder()
                 .method("POST")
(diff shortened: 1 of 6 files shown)


DaleSeo (approved):
Thanks, @inanna-apollo! Congrats on your first PR to the repo! :)

crates/apollo-mcp-server/src/auth.rs
```diff
@@ -1624,6 +1629,36 @@ emyPxgcYxn/eR44/KJ4EBs+lVDR3veyJm+kXQ99b21/+jh5Xos1AnX5iItreGCc=
             Body::from(r#"{"jsonrpc":"2.0","id":1,"method":"resources/list"}"#)
         }
 
+        fn server_discover_body() -> Body {
+            Body::from(r#"{"jsonrpc":"2.0","id":1,"method":"server/discover"}"#)
+        }
+
+        #[tokio::test]
+        async fn server_discover_without_token_allowed_when_enabled() {
+            // `server/discover` precedes `initialize` in the inline lifecycle,
```
gocamille: I noticed while comparing this with the `2026-07-28` lifecycle that that revision removes the `initialize/initialized` handshake, so `server/discover` does not precede `initialize` there. Authentication is also transport-level in that an HTTP client can attach a bearer token to its first discovery request.
Should we say instead that `server/discover` may be the client’s first MCP request, so this option permits that request anonymously when anonymous discovery is enabled? The same wording appears in `docs/source/auth.mdx` and the changeset.

inanna-apollo: Good catch — you're right on both counts: the 2026-07-28 lifecycle drops `initialize`/`initialized` entirely, and transport-level auth means an HTTP client could attach a bearer token to its very first request anyway. Reworded in auth.rs, auth.mdx, and the changeset to say `server/discover` may be the client's first MCP request, so this option permits that request anonymously — pushed in 5aedb48e9.

PR #820, opened 2026-09-03
refactor: share one supported protocol version list between initialize and rmcp re-negotiation

DaleSeo:
We had the protocol version cap written in two places. `negotiate_protocol_version()`, added in #802, checked that the version was in `KNOWN_VERSIONS` and was less than or equal to `MAX_SUPPORTED_PROTOCOL_VERSION`. `supported_protocol_versions()`, added in #800, built its own list using the same filter. They only stayed in sync because both read `MAX_SUPPORTED_PROTOCOL_VERSION`.

This refactoring moves that list to a module-level `SUPPORTED_PROTOCOL_VERSIONS` next to `MAX_SUPPORTED_PROTOCOL_VERSION`. Both consumers now use the same list, so the rule is defined in one place. Their agreement is now structural instead of just happening because they both read the same constant.


inanna-apollo (approved):
lgtm

PR #822, opened 2026-09-04
fix: express nullability and default values in generated input schemas

### LINKED ISSUE #815: Generated input schemas can't express a nullable variable, so strict clients must send placeholder values
In a generated `inputSchema`, the only thing marking a variable as nullable is that it's missing from `required`. There's never a `{"type": "null"}` on the input side.

So the only way for a model to say "no filter" is to leave the property out. That breaks as soon as a client rewrites the schema for strict function calling — OpenAI-style structured outputs require every key in `properties` to also be in `required`, so the transform removes the one escape hatch the schema had. The model then has to make something up for every optional filter, and it sends `[]` for lists, `""` for dates, and a coin-flip `true`/`false` for booleans.

We hit this with a real client and it made the affected tools unusable.

## Repro

```graphql
query GetThings($statuses: [Status!] = null, $createdAfter: String = null, $isActive: Boolean = null) {
  things(statuses: $statuses, createdAfter: $createdAfter, isActive: $isActive) { id }
}
```

The generated schema has no `required` key at all:

```json
{
  "type": "object",
  "properties": {
    "statuses": { "type": "array", "items": { "$ref": "#/definitions/Status" } },
    "createdAfter": { "type": "string" },
    "isActive": { "type": "boolean" }
  }
}
```

A strict-mode client adds `required: ["statuses", "createdAfter", "isActive"]`, and the call arrives as:

```json
{ "statuses": [], "createdAfter": "", "isActive": false }
```

`createdAfter: ""` fails ISO-8601 parsing and `statuses: []` isn't a valid filter value on our server. The boolean is the one that really hurts: there's no empty value to coerce, so we can't tell "the model was forced to pick one" from "the caller wants unassigned things". We return a confidently wrong result instead of an error.

## The relevant bits

All in `crates/apollo-mcp-server/src/operations/`, on `main`. Same on `v1.7.0`.

`operation.rs:664` — `required` is the only signal:

```rust
if variable.ty.is_non_null() {
    // ...push variable_name into "required"
}
```

`schema_walker/type.rs:36` states the assumption outright:

```rust
// JSON Schema assumes that all properties are nullable unless there is a
// required field, so we treat cases the same here.
```

True for a validator reading the schema as published, but not once a client edits `required`.

Two things suggest the input-property case is just a gap rather than a deliberate stance. Thirteen lines below that comment, list *items* already get an explicit null union ("Arrays, however, do need to specify that fields can be null"). And `schema_walker/output.rs:331` emits `oneOf: [inner, {"type": "null"}]` for nullable output fields. Your own `nullable_named_type` snapshot shows both sides for the same nullable `ID`:

- input: `{"id": {"type": "string"}}`
- output: `{"id": {"oneOf": [{"type": "string"}, {"type": "null"}]}}`

## What we'd suggest

Emit the same union on the way in:

```json
"createdAfter": { "oneOf": [{ "type": "string" }, { "type": "null" }] }
```

That doesn't make the schema fully strict-mode compliant (that also wants `additionalProperties: false`), but it makes the strict transform *sound*: with `null` in the union a client can move everything into `required` and still pass `null`, which GraphQL already treats as the absent-variable case. Lists, scalars and booleans all get fixed at once, and the boolean has no other fix. Lenient clients don't notice — omitting the property stays valid.

If changing the default worries you, a flag to opt in would be enough to unblock us.

## On the existing workaround

Pinning a variable to a fixed value does drop it from the schema entirely (`get_json_schema` skips anything in `variable_overrides`, `operation.rs:632`), which is a nice escape hatch — but it's out of reach for local operation files. `raw_operation.rs` hardcodes `variables: None` in the local-file constructor; only the collection poller populates it. With `operations.source: local` there's no way to get at it without moving to a GraphOS collection. Possibly related to #758.

This also feels like the same family as PR #272 (`oneOf` into `items` for OpenAI), the `Int` → `"integer"` mapping, and the Draft-07 downgrade — client schema compatibility rather than spec purity.

DaleSeo:
Fixes #815.

The generated input schemas showed that a variable was nullable only when it was left out of `required`. Clients that rewrite tool schemas for strict function calling move every property into `required`. That removed the only way for a model to leave a filter unset, so it filled every optional variable with a placeholder like `""`, `[]`, or a guessed boolean.

Now, nullable variables, input object fields, and list items include an explicit `{"type": "null"}` option. This matches the output schema and keeps `null` valid after the rewrite. Omitting the property is still valid for all other clients.

Since strict clients will now send `null` instead of omitting a value, GraphQL default values could otherwise be skipped silently. Defaults declared on variables and input fields are now emitted as JSON Schema `default`. Also, a non-null variable with a default is no longer marked as `required`, which matches how `apollo-compiler` already handles input fields.

Descriptions for list-type variables were moved from `items` to the property itself, so they do not get buried under the new wrapper.

## Testing

Ran the server binary from this branch over stdio against a local schema with the operation from #815 (plus a `$limit: Int! = 10` variable to exercise defaults), captured the `GetThings` tool from `tools/list`, and sent that exact schema to both providers with two prompts: "Show me the active things." and "Show me all things." Each run compares the schema as emitted (lenient mode) with what a strict-mode client sends after its rewrite: every property in `required`, `additionalProperties: false`, `definitions` renamed to `$defs`.

Emitted property, for reference:

```json
"isActive": {
  "description": "Only active (true) or inactive (false) things.",
  "anyOf": [{ "type": "boolean" }, { "type": "null" }],
  "default": null
}
```

### Results

| Schema | OpenAI `gpt-4.1`, `strict: true` | Anthropic `claude-opus-5`, `strict: true` | Lenient mode, both providers |
|---|---|---|---|
| `oneOf` union (first cut of this PR) | 400 `'oneOf' is not permitted` | 400 `Schema type 'oneOf' is not supported` | 200, unset filters omitted |
| `anyOf` union (this PR) | 200 | 200 | 200, unset filters omitted |

Arguments the models sent for "Show me all things." with the `anyOf` schema in strict mode, on both providers:

```json
{ "statuses": null, "createdAfter": null, "isActive": null, "limit": 10 }
```

Arguments for "Show me the active things.":

```json
{ "statuses": null, "createdAfter": null, "isActive": true, "limit": 10 }
```


crates/apollo-mcp-server/src/operations/operation.rs
```diff
@@ -661,7 +662,7 @@ fn get_json_schema(
                 .get_or_insert(&mut Map::default())
                 .insert(variable_name.clone(), nested.into());
 
-            if variable.ty.is_non_null() {
+            if variable.ty.is_non_null() && variable.default_value.is_none() {
```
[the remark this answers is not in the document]
DaleSeo: Keeping the null branch. In manual testing, `default` reached the model and influenced it. With `"default": 10`, both sent 10. Without it, OpenAI made up 20 or 50. If we drop the null branch for variables with defaults, explicit `null` would not be passable for variables where it means something different from omission. I agree the `required` change only helps lenient clients, so I will mention that in the changeset.

crates/apollo-mcp-server/src/operations/schema_walker.rs
```diff
@@ -25,15 +31,64 @@ pub fn type_to_schema(
...
+        default_value,
+    )
+}
+
+/// Modifies a schema to include the default value from the GraphQL definition, if any
+fn with_default(mut schema: Schema, default_value: Option<&GraphQLValue>) -> Schema {
+    if let Some(default) = default_value.and_then(graphql_value_to_json) {
+        schema
+            .ensure_object()
+            .insert("default".to_string(), default);
```
gocamille: I followed the call to `graphql_value_to_json` to understand how we convert defaults and it looks like it converts the literal value without its GraphQL type. For example `$values: [Int!]! = 1` is valid GraphQL and the [list coercion rule](https://spec.graphql.org/September2025/#sec-List.Input-Coercion) turns that default into `[1]`. Here we emit `default: 1` but the generated input schema is looking for an array. If a client uses that default and validates the arguments then validation would fail. Should we use the declared GraphQL type to convert these defaults?

DaleSeo: That's great feedback, @gocamille! I've updated the function to account for the declared GraphQL types.

gocamille (approved):
Thank you @DaleSeo !

PR #824, opened 2026-09-08
fix: report config file errors against the file, not the env

DaleSeo:
A validation error in a YAML config file was reported as coming from `APOLLO_MCP_` environment variables whenever an `APOLLO_MCP_<SECTION>__*` variable was set for the same top-level section. This sent operators looking in their deployment environment instead of the file that actually had the problem.

Figment tags a section built from multiple providers with the provider that wins precedence. Since the environment takes precedence over the file, it was being blamed for errors it did not cause.

Reordering the providers does not fix this. With both join and merge ordering, the provider that wins precedence also owns the tag. The error is now attributed to the source that contains the invalid key, which is usually the file unless the key only comes from the environment.

Errors that really come from an `APOLLO_MCP_` variable still identify the environment. Errors from a config file now include its path instead of only saying "YAML source string".

## Testing

Build the server and create a config file with a missing required field. In this case, `transport.auth.scopes` is missing.

```yaml
# config.yaml
endpoint: https://example.com/graphql
transport:
  type: streamable_http
  auth:
    servers:
      - https://auth.example.com
    resource: https://mcp.example.com/mcp
```

Run it with an unrelated `transport` variable set and confirm that the file is named in the error.

```
APOLLO_MCP_TRANSPORT__PORT=5000 ./target/debug/apollo-mcp-server config.yaml
# Error: missing field `scopes` for key "default.transport" in config file 'config.yaml'
```

Then confirm that a real environment error still identifies the environment. Use a valid config file with `transport.type: streamable_http` and run:

```
APOLLO_MCP_TRANSPORT__BOGUS=1 ./target/debug/apollo-mcp-server config.yaml
# Error: unknown field: found `bogus`, ... for key "TRANSPORT" in `APOLLO_MCP_` environment variable(s)
```

### DIFF
--- a/crates/apollo-mcp-server/src/runtime.rs
+++ b/crates/apollo-mcp-server/src/runtime.rs
@@ -16,12 +16,14 @@ mod schema_source;
 mod schemas;
 pub mod telemetry;
 
+use std::mem::discriminant;
 use std::path::Path;
 
 pub use config::Config;
 use figment::{
-    Figment,
+    Figment, Metadata, Profile, Provider,
     providers::{Env, Format, Yaml},
+    value::{Dict, Map},
 };
 pub use operation_source::{IdOrDefault, OperationSource};
 pub use schema_source::SchemaSource;
@@ -30,19 +32,18 @@ pub use schema_source::SchemaSource;
 const ENV_NESTED_SEPARATOR: &str = "__";
 
 /// Read configuration from environment variables only (when no config file is provided)
-#[allow(clippy::result_large_err)]
+// `figment::Error` is large, but this runs once at startup, so boxing it buys nothing
+#[expect(clippy::result_large_err)]
 pub fn read_config_from_env() -> Result<Config, figment::Error> {
-    Figment::new()
-        .join(apollo_common_env())
-        .join(Env::prefixed("APOLLO_MCP_").split(ENV_NESTED_SEPARATOR))
-        .extract()
+    env_figment().extract()
 }
 
 /// Read in a config from a YAML file, filling in any missing values from the environment.
 ///
 /// Environment variable references using `${env.VAR_NAME}` syntax are expanded
 /// before the YAML is parsed.
-#[allow(clippy::result_large_err)]
+// `figment::Error` is large, but this runs once at startup, so boxing it buys nothing
+#[expect(clippy::result_large_err)]
 pub fn read_config(yaml_path: impl AsRef<Path>) -> Result<Config, figment::Error> {
     // Read and expand environment variables in the config content
     let content = std::fs::read_to_string(yaml_path.as_ref()).map_err(|e| {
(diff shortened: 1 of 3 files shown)


crates/apollo-mcp-server/src/runtime.rs
```diff
@@ -56,11 +57,86 @@ pub fn read_config(yaml_path: impl AsRef<Path>) -> Result<Config, figment::Error
...
+    // the root, so an error about the root itself is always the file's.
+    if key.is_empty() {
+        return false;
+    }
+
+    let Ok(value) = env_figment().find_value(key) else {
+        return false;
+    };
+
+    value.as_dict().is_none() || !Figment::from(file).contains(key)
```
gocamille: I followed how this decides whether an error comes from the file or the environment and I think this may miss the reverse case where there is a valid file with an invalid environment override, something like: 

```yaml
transport:
  type: streamable_http
  address: 127.0.0.1
```

Then `APOLLO_MCP_TRANSPORT__ADDRESS=not-an-ip-address` overriding the valid address. I noticed that Serde reports the error path as just `transport` without `address` but both sources contain taht section so the expression returns `false`. The error then names the file even though the invalid address comes from the environment. Before the change the error would correctly name the environment. Should we cover this case and preserve the environment as a source of the error?

crates/apollo-mcp-server/src/runtime.rs
```diff
@@ -56,11 +57,86 @@ pub fn read_config(yaml_path: impl AsRef<Path>) -> Result<Config, figment::Error
...
+/// both sources filled in belongs to the file unless the file never mentions it.
+fn from_environment(key: &str, file: &ConfigFile<'_>) -> bool {
+    // An empty key is the document root. Environment variables only ever add keys beneath
+    // the root, so an error about the root itself is always the file's.
+    if key.is_empty() {
+        return false;
+    }
+
+    let Ok(value) = env_figment().find_value(key) else {
+        return false;
```
gocamille: I checked here what happens with an invalid item in a list. For example `config.yaml` has `{}` and `APOLLO_MCP_FORWARD_HEADERS='[true]'` supplies a boolean where the server expects a string.
Figment reports the path as `forward_headers.0` where `0` means the first list item, but `find_value()` only looks through dictionaries. It cannot find that list item so this branch returns `false`.
The error then names `config.yaml` even though the file contains no `forward_headers` value. 
Should this lookup also handle list items, or check which source supplies the containing list?

crates/apollo-mcp-server/src/runtime.rs
DaleSeo: Good catch, @gocamille! The environment supplies `Dict({"address": "not-an-ip-address"})` at the coarse path `["transport"]`, so the dictionary check interpreted that as both sources contributing and included the file name. I fixed it and added a regression test.

crates/apollo-mcp-server/src/runtime.rs
DaleSeo: Same root cause as your previous comment. So instead of teaching the lookup to index lists, I removed key-path inference entirely. Now it just checks whether the config file alone reproduces the failure.

gocamille (approved):

PR #826, opened 2026-09-09
docs: describe how GraphQL types map to generated tool schemas

DaleSeo:
#822 and #823 changed how generated tool schemas represent nullability, default values, and `required`. As a quick follow-up, this PR adds a "Generated tool schemas" reference to the Define MCP Tools page, including a worked example with generated `inputSchema` and `outputSchema`.

### DIFF
--- a/docs/source/define-tools.mdx
+++ b/docs/source/define-tools.mdx
@@ -497,9 +497,9 @@ An input property's `description` comes from the first available source:
 
 ### Output schema
 
-With `overrides.enable_output_schema: true`, the tool also carries an `outputSchema` describing the GraphQL response envelope: `data`, an optional `errors` array in the standard GraphQL error shape, and `extensions`. The `data` subtree follows the operation's selection set. For the example above:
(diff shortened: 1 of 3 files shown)


docs/source/define-tools.mdx
```diff
@@ -304,6 +304,273 @@ overrides:
...
+The generated schema always includes all three envelope members. `errors` is an array of the standard GraphQL error shape (`message`, `locations`, `path`, `extensions`), and `extensions` is an unconstrained object.
+
+The output side follows the same conventions as the input side, with a few differences:
+
+- Nullable fields and lists are wrapped in `anyOf` with `null`. Object fields are `required` when their GraphQL type is non-nullable.
+- `ID` is `{"oneOf": [{"type": "string"}, {"type": "integer"}]}`, since servers may return either.
+- A union field with inline fragments becomes an `anyOf` of one object schema per fragment. For an interface field, fields from inline fragments are merged into a single object schema. `__typename`, when selected, is a plain string and isn't a discriminator.
+- Fields marked `@private` are omitted from the schema. See the [`@private` directive](/apollo-mcp-server/mcp-apps-reference#private-directive) for details.
+
+MCP clients that validate `structuredContent` against `outputSchema` reject responses that don't match, so the schema is designed to accept every response the operation can legitimately produce.
```
gocamille: Can we qualify the claim that every valid response is accepted? For example, `name @include(if: false)` correctly omits `name` but the generated schema still requires it when its type is `String!`. A validating MCP client can then reject a successful response.

gocamille (approved):

PR #829, opened 2026-09-09
Add TTL and cacheScope to list responses

inanna-apollo:
Prepares SEP-2549 cache hints for `tools/list`, `resources/list`, `resources/read`, and `prompts/list` responses.

Adds top-level `caching.ttl_ms` (default: 300,000 ms), configurable through `APOLLO_MCP_CACHING__TTL_MS`. Hints use `private` scope because response visibility can depend on authentication scope. Remote app resource reads omit hints because their content can change independently of server configuration.

Cache hints remain inactive. The helper requires a peer version of `2026-07-28` or newer that also falls within the server's supported range; the server deliberately caps support at `2025-11-25`. Activation requires a separate protocol-support change.

Validation covers configuration parsing, environment precedence, schema defaults, hint construction through rmcp builders, both protocol bounds, and local/remote resource selection. HTTP regression tests cover unsupported-version requests with and without metadata in both transport configurations. An initialized supported client still receives tools without hints. Positive end-to-end hint availability remains future activation work.

Workspace tests with all features, Clippy across all targets/features, and formatting checks passed.


crates/apollo-mcp-server/src/runtime/overrides.rs
```diff
@@ -45,6 +46,11 @@ pub struct Overrides {
     /// with `WWW-Authenticate: Bearer error="insufficient_scope"`.
     #[serde(default)]
     pub required_scopes: HashMap<String, OperationRequiredScopes>,
+
+    /// Caching behavior for MCP list/read response cache hints (SEP-2549: `ttlMs`/`cacheScope`
+    /// on `tools/list`, `resources/list`, `resources/read`, and `prompts/list` responses).
+    #[serde(default)]
+    pub caching: Caching,
```
DaleSeo: Every sibling key under `overrides` replaces something derived from the operation source, but this new key does not override anything under it. Have you considered putting it at the top-level instead? Moving it after release would be a breaking config change, so I'd like us to think it through more before shipping it.

crates/apollo-mcp-server/src/caching.rs
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    /// Applies the `ttlMs`/`cacheScope` cache hints to `result` when the negotiated protocol
+    /// version supports them (SEP-2549, MCP `2026-07-28`+); older peers are left untouched.
+    pub(crate) fn apply_to<T: CacheHints>(
+        &self,
+        result: &mut T,
+        protocol_version: Option<&ProtocolVersion>,
+    ) {
+        if protocol_version.is_some_and(|v| *v >= ProtocolVersion::V_2026_07_28) {
```
DaleSeo: I think the gate should also check the server's own `MAX_SUPPORTED_PROTOCOL_VERSION`, not just the peer's claim. Otherwise, a client that sends `MCP-Protocol-Version: 2026-07-28` on `tools/list` could receive `ttlMs` and `cacheScope` from a server that only supports up to `2025-11-25`.

```suggestion
        let negotiated = protocol_version.filter(|v| **v <= MAX_SUPPORTED_PROTOCOL_VERSION);
        if negotiated.is_some_and(|v| *v >= ProtocolVersion::V_2026_07_28) {
```

crates/apollo-mcp-server/src/server/states/running.rs
```diff
@@ -540,12 +574,14 @@ impl Running {
...
         let app_param = extract_app_param(&extensions);
-        let app_target = AppTarget::try_from((extensions, client_capabilities))?;
+        let app_target = AppTarget::try_from((extensions, peer.client_capabilities))?;
 
         if let Some(app_name) = app_param {
             let resource =
                 get_app_resource(&self.apps, request, request_uri, &app_target, &app_name).await?;
-            Ok(ReadResourceResult::new(vec![resource]))
+            let mut result = ReadResourceResult::new(vec![resource]);
+            self.caching.apply_to(&mut result, peer.protocol_version);
```
DaleSeo: `get_app_resource` fetches `AppResourceSource::Remote` content over HTTP every time it is read. Also, `get_info` only advertises a bare `ResourcesCapability::default()` with no `subscribe` or `list_changed`.

```suggestion
            if !app_resource_is_remote {
                self.caching.apply_to(&mut result, peer.protocol_version);
            }
```

crates/apollo-mcp-server/src/caching.rs
```diff
@@ -0,0 +1,80 @@
...
+use schemars::JsonSchema;
+use serde::Deserialize;
+
+/// Caching behavior for MCP list/read responses.
+#[derive(Debug, Clone, Copy, Deserialize, JsonSchema)]
+#[serde(default, deny_unknown_fields)]
+pub struct Caching {
+    /// Time, in milliseconds, that `tools/list`, `resources/list`, `resources/read`, and
+    /// `prompts/list` responses may be treated as fresh by clients.
+    #[serde(default = "Caching::default_ttl_ms")]
```
DaleSeo: Isn't this redundant with the manual `Default` impl below?

crates/apollo-mcp-server/src/caching.rs
```diff
@@ -0,0 +1,80 @@
...
+        self.cache_scope = Some(scope);
+    }
+}
+
+impl CacheHints for rmcp::model::ReadResourceResult {
+    fn set_cache_hints(&mut self, ttl_ms: u64, scope: CacheScope) {
+        self.ttl_ms = Some(ttl_ms);
+        self.cache_scope = Some(scope);
+    }
+}
```
DaleSeo: All four result types already expose `with_ttl_ms` and `with_cache_scope` in rmcp, and the four impl bodies here are identical. Would it be clearer to delegate to upstream builders generated by a single macro? This is not a blocking comment.

inanna-apollo: I replaced the four repeated implementations with one small macro delegating to rmcp's `with_ttl_ms` and `with_cache_scope` builders. The shared helper now consumes and returns the result, avoiding clones or temporary replacement values. Tests cover all four result types and preservation of existing fields.

crates/apollo-mcp-server/src/caching.rs
inanna-apollo: Removed the field-level Serde default. Deserialization uses the struct-level default and manual `Default` implementation. A schema-only default annotation preserves the generated JSON Schema default; tests cover absent/empty sections and the schema value.

crates/apollo-mcp-server/src/server/states/running.rs
inanna-apollo: Remote reads now omit both hints. The resolver reports the actual selected source's local/remote origin, so target-specific resources follow the selected target rather than the overall app. Tests verify content retrieval, source selection, and eligibility independently of the currently inactive protocol gate. The docs explicitly describe the remote-read exception without claiming that omitted hints prohibit all client caching.

crates/apollo-mcp-server/src/caching.rs
inanna-apollo: Added an explicit server-cap check in the shared caching helper: the peer version must be at least `2026-07-28` and no greater than `MAX_SUPPORTED_PROTOCOL_VERSION`. Handler tests now verify that even a synthetic future peer cannot enable hints under the current cap. HTTP regression tests also verify rejection of future-version requests both without metadata and with matching metadata, with legacy sessions enabled and disabled. This no longer relies solely on rmcp's rejection path.

crates/apollo-mcp-server/src/runtime/overrides.rs
inanna-apollo: Moved this to top-level `caching.ttl_ms`, including runtime wiring, documentation, changeset, schema coverage, and `APOLLO_MCP_CACHING__TTL_MS` precedence coverage.

crates/apollo-mcp-server/src/caching.rs
```diff
@@ -0,0 +1,148 @@
...
+
+/// Implemented by MCP list/read result types that carry SEP-2549 cache hints, so
+/// [`Caching::apply_to`] can set them without duplicating the gating logic per call site.
+pub(crate) trait CacheHints {
+    fn with_cache_hints(self, ttl_ms: u64, scope: CacheScope) -> Self;
+}
+
+fn supports_cache_hints(peer: Option<&ProtocolVersion>, server_max: &ProtocolVersion) -> bool {
+    peer.is_some_and(|version| *version >= ProtocolVersion::V_2026_07_28 && version <= server_max)
```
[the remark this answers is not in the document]
inanna-apollo: Keeping the strict upper bound intentionally. In rmcp 3.2, initialized peers have their peer_info protocol version updated to the negotiated version; the original requested version is not what these handlers subsequently receive. Inline HTTP requests declaring an unsupported version are rejected before handler dispatch. A min(requested, server_max) here would instead treat an unsupported claim as if negotiation had succeeded, weakening the explicit server-cap guard requested in the earlier review.

Added does_not_treat_an_unsupported_claim_as_a_negotiated_version in 0fea999e7: even with a server cap of 2026-07-28, a peer claiming 2030-01-01 receives no hints. Existing transport negotiation tests and the unsupported-inline-request regression remain in place.

crates/apollo-mcp-server/src/server/states/running.rs
```diff
@@ -2776,6 +2969,71 @@ mod integration_tests {
...
+            );
+            assert!(result.get("ttlMs").is_none());
+            assert!(result.get("cacheScope").is_none());
+        }
+
+        #[rstest::rstest]
+        #[case::missing_metadata(false, -32602)]
+        #[case::unsupported_version(true, -32022)]
+        #[tokio::test]
+        async fn rejects_future_protocol_before_returning_cache_hints(
```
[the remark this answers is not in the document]
inanna-apollo: Addressed in 0fea999e7. Split the cases into rejects_stateless_tools_request_without_required_metadata and rejects_stateless_tools_request_with_unsupported_protocol, each covering legacy-session mode enabled and disabled. Assertions now use ErrorCode::INVALID_PARAMS and ErrorCode::UNSUPPORTED_PROTOCOL_VERSION and include the response body in failure diagnostics. The tests share only request construction/dispatch; each names and asserts its own behavior.

crates/apollo-mcp-server/src/caching.rs
```diff
@@ -0,0 +1,148 @@
...
+
+    /// Applies the `ttlMs`/`cacheScope` cache hints to `result` when the negotiated protocol
+    /// version supports them (SEP-2549, MCP `2026-07-28`+) and is within this server's
+    /// supported range. A peer's claim alone cannot enable an unsupported revision.
+    pub(crate) fn apply_to<T: CacheHints>(
+        &self,
+        result: T,
+        protocol_version: Option<&ProtocolVersion>,
+    ) -> T {
+        if supports_cache_hints(protocol_version, &MAX_SUPPORTED_PROTOCOL_VERSION) {
```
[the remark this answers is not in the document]
inanna-apollo: Addressed in 0fea999e7. Caching::apply_to now takes an explicit server_max, removing the caching -> server::states dependency. Production entry points remain fixed to MAX_SUPPORTED_PROTOCOL_VERSION, while the shared handler implementations accept the cap explicitly.

New parameterized tests exercise default/custom/zero TTL and private scope through all four handler implementations with a supported cap and nonempty results. Remote-resource reads are also exercised with both the current cap and a cap that permits hints, verifying successful content retrieval without hints. The redundant always-None expected-value parameters were removed from the existing negative tests.

All 1,037 workspace tests, Clippy across all targets/features, formatting, and diff checks passed. This is direct handler-implementation coverage; positive end-to-end availability still awaits protocol activation.

crates/apollo-mcp-server/src/server/states/running.rs
```diff
@@ -275,8 +305,19 @@ impl Running {
...
         extensions: Extensions,
-        client_capabilities: Option<&ClientCapabilities>,
-        protocol_version: Option<&ProtocolVersion>,
+        peer: PeerContext<'_>,
+    ) -> Result<ListToolsResult, McpError> {
+        self.list_tools_with_protocol_max(extensions, peer, &MAX_SUPPORTED_PROTOCOL_VERSION)
+            .await
+    }
+
+    // Keep the production cap fixed at the entry point while allowing the shared
```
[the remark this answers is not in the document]
inanna-apollo: Keeping the explicit wrappers after considering this tradeoff. The server support limit is server policy rather than peer state; putting an optional override into PeerContext would mix those responsibilities and make the fixed production boundary less explicit. The current entry points always supply MAX_SUPPORTED_PROTOCOL_VERSION, while tests exercise the same implementations with an explicit supported cap. This also preserves the existing call sites, and caching.rs remains independent of the server state machine. The extra forwarding methods are a deliberate, bounded cost for that separation, so we are leaving this nonblocking design suggestion unapplied.

crates/apollo-mcp-server/src/server/states/running.rs
```diff
@@ -1243,14 +1417,35 @@ mod tests {
...
             .resources;
 
             assert_eq!(resources.len(), 1);
             assert_eq!(resources[0].uri, RESOURCE_URI);
         }
 
+        #[rstest]
+        #[case::above_server_cap(ProtocolVersion::V_2026_07_28, 60_000)]
+        #[case::legacy(ProtocolVersion::V_2025_06_18, 60_000)]
+        fn resource_list_cache_hints_gated_by_protocol_version(
```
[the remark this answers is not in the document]
inanna-apollo: Addressed in f5e29b5e2. All four negative tests now parameterize only protocol_version and set a non-default 60,000 ms TTL directly in the fixture. Default/custom/zero TTL coverage remains in the separate positive-path tests. Workspace tests, Clippy, formatting, and diff checks passed.

crates/apollo-mcp-server/src/server/states/running.rs
DaleSeo: Nice! Thanks for handling the case I hadn't considered. :)

DaleSeo (approved):
Thanks for addressing my feedback!

PR #830, opened 2026-09-09
Refactor/service owned tool notifications

inanna-apollo:
# refactor: make legacy tool notifications service-owned

Catalog updates currently fan out through a shared registry of MCP peers, allowing slow clients to delay other clients and requiring manual cleanup. Replace that registry with a coalescing catalog-change signal and a notification task owned by each initialized legacy rmcp service. Catalog updates finish independently of notification delivery, and teardown releases delivery resources without another update.

A distinct `McpService` implements the protocol handler; application state cannot accidentally be served without lifecycle ownership. A receiver is registered before returning the initialize response, preserving catalog invalidations while rmcp schedules the initialized callback. Delivery starts after that callback and is shared across service clones.

Tests exercise delayed initialization, repeated callbacks, reconnects, session deletion, final-owner cleanup, and both shutdown sources through production HTTP construction and serving functions. The forwarding property model covers idle and pending changes, send success/failure, and cancellation. Existing sessionless requests remain covered separately from legacy notification tests.

This is preparatory work for the stateless-protocol ticket. It does not implement `subscriptions/listen`, enable MCP `2026-07-28`, or change the `stateful_mode` default. Those remain companion work; this PR should not close the overall ticket.

Validation: workspace tests and all-feature instrumented workspace tests pass; all-target Clippy with warnings denied, formatting, and diff checks pass. Changed production executable-line coverage is 101/104 (97.1%); including test code, it is 535/553 (96.7%). The production calculation excludes cfg(test) items. The missed-update regression was verified by temporarily restoring late registration: the test failed waiting for the missing notification. Documentation builds with the existing private-link warning in `host_validation.rs`.

### DIFF
--- a/crates/apollo-mcp-server/Cargo.toml
+++ b/crates/apollo-mcp-server/Cargo.toml
@@ -77,8 +77,8 @@ chrono = { version = "0.4.41", default-features = false, features = ["now"] }
 figment = { version = "0.10.19", features = ["test"] }
 insta.workspace = true
 mockito = "1.7.0"
-proptest = "1"
 opentelemetry_sdk = { version = "0.32.1", features = ["testing"] }
+proptest = "1"
 rstest.workspace = true
 secrecy.workspace = true
 serde_yaml = "0.9.34"
--- a/crates/apollo-mcp-server/Cargo.toml
+++ b/crates/apollo-mcp-server/Cargo.toml
@@ -58,7 +58,7 @@ serde_yaml = "0.9.34"
 shellexpand = { version = "3.1", default-features = false, features = ["base-0"] }
 thiserror.workspace = true
 tokio.workspace = true
-tokio-util = "0.7.15"
+tokio-util = { version = "0.7.15", features = ["rt"] }
 tonic = "0.14"
 tower-http = { version = "0.6.6", features = ["cors", "trace"] }
 tracing-appender = "0.2.3"
@@ -77,11 +77,12 @@ chrono = { version = "0.4.41", default-features = false, features = ["now"] }
 figment = { version = "0.10.19", features = ["test"] }
 insta.workspace = true
 mockito = "1.7.0"
+proptest = "1"
 opentelemetry_sdk = { version = "0.32.1", features = ["testing"] }
 rstest.workspace = true
 secrecy.workspace = true
 serde_yaml = "0.9.34"
-tokio.workspace = true
+tokio = { workspace = true, features = ["test-util"] }
 tower = "0.5.2"
 tracing-test = "0.2.5"
 
--- a/crates/apollo-mcp-server/Cargo.toml
+++ b/crates/apollo-mcp-server/Cargo.toml
@@ -82,6 +82,7 @@ proptest = "1"
 rstest.workspace = true
 secrecy.workspace = true
 serde_yaml = "0.9.34"
+sse-stream = "0.2.5"
 tokio = { workspace = true, features = ["test-util"] }
 tower = "0.5.2"
 tracing-test = "0.2.5"
--- a/crates/apollo-mcp-server/src/server/states/running.rs
+++ b/crates/apollo-mcp-server/src/server/states/running.rs
@@ -4027,16 +4027,6 @@ mod integration_tests {
                 .unwrap()
         }
 
-        async fn wait_for_receivers(running: &Running, expected: usize) {
-            tokio::time::timeout(std::time::Duration::from_secs(2), async {
-                while running.tool_list_changes.receiver_count() != expected {
-                    tokio::task::yield_now().await;
-                }
-            })
-            .await
-            .expect("initialization must register a change receiver");
-        }
-
         async fn initialize_legacy_session(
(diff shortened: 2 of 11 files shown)


crates/apollo-mcp-server/src/server/states/tool_list_changes.rs
```diff
@@ -0,0 +1,396 @@
...
+                return;
+            }
+            Ok(Err(error)) => {
+                error!(?error, "Failed to notify client of tool list change");
+            }
+            Err(_) => {
+                error!(
+                    "Timed out notifying client of tool list change after 5s - stopping legacy delivery"
+                );
+                return;
```
DaleSeo, line 109: Now that a slow client can't block other sessions, do we still want to give up on this one permanently?

crates/apollo-mcp-server/src/server/states/running.rs
```diff
@@ -3745,82 +3648,544 @@ mod integration_tests {
...
+                .clone()
+                .oneshot(session_request(
+                    &session,
+                    http::Method::POST,
+                    json!({"jsonrpc": "2.0", "method": "notifications/initialized"}),
+                ))
+                .await
+                .unwrap();
+            assert_eq!(response.status(), StatusCode::ACCEPTED);
+            wait_for_receivers(running, previous + 1).await;
```
DaleSeo: What is this meant to wait for?

crates/apollo-mcp-server/src/server/states/tool_list_changes.rs
```diff
@@ -0,0 +1,396 @@
...
+            Ok(Err(ServiceError::TransportSend(_) | ServiceError::TransportClosed)) => {
+                error!("Failed to notify client of tool list change - stopping legacy delivery");
+                return;
+            }
+            Ok(Err(error)) => {
+                error!(?error, "Failed to notify client of tool list change");
+            }
+            Err(_) => {
+                error!(
+                    "Timed out notifying client of tool list change after 5s - stopping legacy delivery"
```
DaleSeo: `5s` drifts if NOTIFY_TIMEOUT changes.

crates/apollo-mcp-server/Cargo.toml
```diff
@@ -77,11 +77,12 @@ chrono = { version = "0.4.41", default-features = false, features = ["now"] }
 figment = { version = "0.10.19", features = ["test"] }
 insta.workspace = true
 mockito = "1.7.0"
+proptest = "1"
```
DaleSeo: nit: putting this after `opentelemetry_sdk` would keep the list sorted.

DaleSeo (approved):
Thanks for the refactor! I only have questions and nitpicks.

crates/apollo-mcp-server/src/server/states/tool_list_changes.rs
inanna-apollo: Agreed—removed the timeout. A slow send can now recover without losing future notifications, and the test checks that it doesn't hold up other clients.

crates/apollo-mcp-server/src/server/states/running.rs
inanna-apollo: Good catch, this became redundant when registration moved into initialize. The assertion above already checks it. Removed this call and the one in the stdio test.

crates/apollo-mcp-server/src/server/states/tool_list_changes.rs
inanna-apollo: Removed along with the timeout, per the discussion above.

crates/apollo-mcp-server/Cargo.toml
inanna-apollo: Fixed, thanks!

crates/apollo-mcp-server/src/server/states/tool_list_changes.rs
```diff
@@ -0,0 +1,396 @@
...
+        }
+
+        let result = tokio::select! {
+            biased;
+            _ = shutdown.cancelled() => return,
+            result = tokio::time::timeout(NOTIFY_TIMEOUT, notify()) => result,
+        };
+        match result {
+            Ok(Ok(())) => {}
+            Ok(Err(ServiceError::TransportSend(_) | ServiceError::TransportClosed)) => {
```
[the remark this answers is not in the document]
inanna-apollo: Checked rmcp 3.2: the local HTTP session's common notification channel caches the event and suppresses the closed SSE receiver's send error, so that reconnect gap doesn't itself produce TransportSend. The updated test publishes while both GET streams are closed, resumes with Last-Event-ID, skips rmcp's inclusive replay of the cursor event, and requires a distinct notification and the updated catalog. It also checks delivery after reconnection.

We're keeping the pre-PR stop-on-TransportSend policy here rather than introducing retries. The code now notes that this is application policy, not a guarantee that TransportSend always means permanent closure. Explicit TransportClosed logs at debug; TransportSend still logs its underlying error, which can include transport-specific closure errors. The separate five-second cutoff is removed, so a slow send alone no longer disables forwarding.

PR #831, opened 2026-09-10
refactor: delegate protocol version negotiation to rmcp

DaleSeo:
rmcp 3.3.0 added `ServerHandler::negotiate_initialize` and `ProtocolVersion::known_up_to` in https://github.com/modelcontextprotocol/rust-sdk/pull/1247. These provide the SDK equivalents of the negotiation this server was handling manually: choosing the protocol version for the `initialize` response and capping it at the newest version the server actually supports.

#820 made that cap apply across all transports by narrowing `supported_protocol_versions`. This refactor removes the remaining duplication by deleting the local `negotiate_protocol_version` and the `SUPPORTED_PROTOCOL_VERSIONS` `LazyLock`. That keeps the rule upstream, so it cannot drift when rmcp changes.

### DIFF
--- a/.changeset/delegate_protocol_version_negotiation_to_rmcp.md
+++ b/.changeset/delegate_protocol_version_negotiation_to_rmcp.md
@@ -1,0 +1,9 @@
+---
+default: patch
+---
+
+# Delegate protocol version negotiation to rmcp
+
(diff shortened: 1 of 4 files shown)


inanna-apollo (approved):
lgtm

PR #837, opened 2026-09-14
fix: adapt server icons to the service refactor

gocamille:
The merge of #806 fails to compile after the service refactor. `McpService` accesses a field on the wrong owner. Both icon tests call `Running::get_info()`, which `Running` no longer implements.

Read icons from `self.application.server_info` and call `running.for_service().get_info()` in both existing icon tests. This preserves default icon omission and forwarding of configured icons, including optional fields.

Validation against main at `e624c8e38929208c95975793f436a2d7b68886d8` plus this fix:

- `cargo test --locked`: 1,072 tests passed, including both icon tests.
- `cargo clippy --all-targets --all-features --locked -- --deny warnings`: passed.
- `cargo fmt --check`: passed.
- `cargo doc --no-deps --locked`: passed with the existing private-link warning in `host_validation.rs`.
- `cargo llvm-cov --all-features --workspace --locked --codecov`: passed. Changed executable lines: 3/3 covered (100%).
- Fresh Docker release build: passed. Existing RTF smoke plan against `apollo-mcp-server:fix806-e624c8e`: 28/28 assertions passed.


.changeset/fix_server_icons_service_refactor.md
DaleSeo, line 1: How about we skip the changeset file for this so the release notes focus on what users can actually see?

DaleSeo (approved):

PR #840, opened 2026-09-17
docs: Update Deployment section to include Apollo GraphOS Operator

jonathanrainer:
Update the docs to include the fact that we can now deploy the MCP Server via the Operator

This links out to Operator docs that won't be published until 1.5.0 goes live, but I wanted to raise this here to get Docs team feedback on it before then so we're not caught out.

I've tried to position the Operator as the best way to deploy onto Kubernetes, while the AIO container is the best way to deploy to anything else, which I think is basically the position


docs/source/deploy.mdx
```diff
@@ -3,9 +3,23 @@ title: Deploy the MCP Server
...
 ---
 
-To deploy Apollo MCP Server in your production environment, use the recommended [Apollo Runtime Container](#apollo-runtime-container-recommended). You can also use a [standalone Apollo MCP Server container](#standalone-apollo-mcp-server-container) if needed.
+To deploy Apollo MCP Server in your production environment, use the [Apollo GraphOS Operator](#deploy-on-kubernetes-with-the-apollo-graphos-operator-recommended-for-kubernetes) if you're running on Kubernetes. Otherwise, use the [Apollo Runtime Container](#apollo-runtime-container-recommended-for-non-kubernetes-deployments). A [standalone Apollo MCP Server container](#standalone-apollo-mcp-server-container) is also available if you need to run it separately from your GraphQL API.
 
-## Apollo Runtime Container (Recommended)
+## Deploy on Kubernetes with the Apollo GraphOS Operator (Recommended for Kubernetes)
+
+If you run Apollo Router on Kubernetes using the [Apollo GraphOS Operator](/apollo-operator/), use it to deploy Apollo MCP Server too. The operator manages the Deployment, Service, ConfigMap, and ephemeral graph API key Secret for you, and can wire an MCP Server directly to an existing Router it also manages.
```
DaleSeo, line 10: While the Runtime Container and standalone container paths carry no plan gate. Since this section now sits above both of them as the recommendation, should the plan requirement be stated here rather than discovered after the reader follows the link?

docs/source/deploy.mdx
```diff
@@ -80,13 +94,13 @@ docker run \
...
-| Kubernetes/orchestrated environment | Apollo Runtime Container     | Fewer moving parts, simpler networking                                                                    |
-| Adding MCP to existing GraphQL API  | Standalone Apollo MCP Server | Connect to your existing GraphQL endpoint                                                                 |
-| Local development                   | `rover dev`                  | [Run `rover dev`](/apollo-mcp-server/run#with-the-rover-cli) to develop locally with both GraphQL and MCP |
+| Scenario                                      | Recommended Option           | Why                                                                                                       |
+|-----------------------------------------------|------------------------------|-----------------------------------------------------------------------------------------------------------|
+| Kubernetes environment                        | Apollo GraphOS Operator      | Manages the Deployment/Service/ConfigMap/Secret lifecycle for you                                         |
```
DaleSeo: Since it can't serve a graph outside GraphOS.

```suggestion
| Kubernetes + GraphOS-managed graph                        | Apollo GraphOS Operator      | Manages the Deployment/Service/ConfigMap/Secret lifecycle for you                                         |
```

docs/source/deploy.mdx
DaleSeo, line 129: @jonathanrainer Do you have anything to add here for operator-managed deployments?

docs/source/deploy.mdx
```diff
@@ -3,9 +3,23 @@ title: Deploy the MCP Server
 subtitle: Deployment using Docker containers, when to choose which option, and production considerations
```
DaleSeo: Can we also update the subtitle to include Apollo GraphOS Operator?

jonathanrainer: Done :)

docs/source/deploy.mdx
jonathanrainer: I've put them in a note below, happy to move it around if you'd prefer?

docs/source/deploy.mdx
jonathanrainer: Done in latest changes

docs/source/deploy.mdx
jonathanrainer: Have added a bit about using the `scale` subresource which means you can use HorizontalPodAutoscalers (or even the KEDA stuff where you can scale based on pretty much anything you can reduce to a metric). Let me know if you think it's enough

docs/source/deploy.mdx
DaleSeo: The operator's own overview page uses the `<PlanRequired>` component. Would it be worth matching?

DaleSeo (approved):
Thanks for addressing my feedback, @jonathanrainer! Looks good!

